PR #5, opened 2021-02-02
6345 6344 Pylint updates, updates from data model changes. Add Docker…

doug-lovett:
… Oracle DB info.


ppr-api/src/ppr_api/models/financing_statement.py
```diff
@@ -52,42 +58,46 @@ class FinancingTypes(Enum):
...
-    lien_amount = db.Column('lien_amount', db.Numeric, nullable=True)
-    surrender_dt = db.Column('surrender_dt', db.Date, nullable=True)
+#    financing_id = db.Column('financing_id', db.Integer, primary_key=True, server_default=db.FetchedValue())
+    financing_id = db.Column('financing_id', db.Integer,
+                             db.Sequence('financing_id_seq'), primary_key=True)
+    state_type_cd = db.Column('state_type_cd', db.String(3), nullable=False)
+                              #, db.ForeignKey('state_type.state_type_cd'))
+    registration_num = db.Column('registration_number', db.String(10), nullable=False)
+    life = db.Column('life', db.Integer, nullable=True)
+    expire_date = db.Column('expire_date', db.Date, nullable=True)
```
thorwolpert, line 65: are these just date, or date times?

ppr-api/src/ppr_api/models/registration.py
```diff
@@ -12,37 +12,44 @@
...
 # limitations under the License.
 """This module holds common statement registration data."""
+
+# pylint: disable=too-many-statements, too-many-branches
+
 #import pycountry
 
 from enum import Enum
 from http import HTTPStatus
+from datetime import date
```
thorwolpert, line 22: in the other apps we're mostly importing our util/datetime.py which wraps datetime and adds in non-naive now()

ppr-api/src/ppr_api/models/search_result.py
```diff
@@ -16,44 +16,36 @@
...
-from .db import db
-
-from ppr_api.utils.datetime import format_ts, now_ts, ts_from_iso_format
-from ppr_api.exceptions import BusinessException
-
-from .search import Search  # noqa: F401 pylint: disable=unused-import; needed by the SQLAlchemy relationship
-
-# temporary mock responses
-import copy
 from registry_schemas.example_data.ppr import FINANCING_STATEMENT_HISTORY
```
thorwolpert, line 25: why is example data here?

thorwolpert: got it, you're mocking responses.

ppr-api/src/ppr_api/resources/party_codes.py
```diff
@@ -13,22 +13,23 @@
 # limitations under the License.
 """API endpoints for maintaining preset client registering and secured parties."""
 
+# pylint: disable=too-many-return-statements
+
 from http import HTTPStatus
 
-from flask import abort, g, request
+from flask import request
 from flask_restplus import Namespace, Resource, cors
```
thorwolpert: i must have missed this earlier, should be restx, or just plain flask

thorwolpert (commented):
just a few comments to resolve
larger discussion on Date vs DateTime

ppr-api/src/ppr_api/models/financing_statement.py
doug-lovett: Bob created the table definitions as all Dates.  Where we care about the timestamp the time is saved correctly by saving as SQLAlchemy datetime

ppr-api/src/ppr_api/models/registration.py
doug-lovett: It's only used for date.fromisoformat...is this a problem?

ppr-api/src/ppr_api/resources/party_codes.py
doug-lovett: Okay, I can make the change later - I used the dependencies from lear.

kialj876 (approved):
There's a lot here so I just skimmed through most of it. Looking good to me, nice work

ppr-api/src/ppr_api/models/registration.py
doug-lovett: To clarify, date.fromisoformat is used when creating a financing statement and creating a renewal statement. The code  can be corrected when developing those transactions.

ppr-api/src/ppr_api/models/financing_statement.py
doug-lovett: To clarify, expire_date as a Date is correctly saving date and time information in the Oracle database. Both Date and DateTime store time information with Oracle Date table column types. The spec defines expiry date as a Date which the model reflects. When developing the financing and renewal statements the model can be revised.

thorwolpert: we'll migrate these to DateTime next sprint, and as the DB is not currently being created from the model, this should not be an issue.

PR #6, opened 2021-02-04
UI framework

kialj876:

Ticket: https://github.com/bcgov/entity/issues/6300

- UI framework
  - setup linting, etc., based off our most current vue projects
  - error framework based off our most current vue projects
  - redirects to auth if not logged in
  - loads dashboard (no styling, etc., but loads search bar)
  - basic tests running/passing error free
  - kept some code in that will be used/tested/tweaked more once we hook up payment and LD and ppr backend
  - kept some of the basic repo stuff in (license, how to contribute, etc., in case this moves to another repo)
  - kept nginx / dockerfiles in , but they may need to be tweaked (didn't spend any time looking into these)


ppr-ui/.eslintrc.js
```diff
@@ -0,0 +1,31 @@
...
+  },
+  extends: [
+    'plugin:vue/essential',
+    '@vue/standard',
+    '@vue/typescript'
+  ],
+  rules: {
+    'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+    'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
+    'max-len': ['warn', { code: 120, ignoreRegExpLiterals: true }]
```
thorwolpert, line 14: Can we add
'no-unused-vars': process.env.NODE_ENV === 'production' ? 'error' : 'warn',

ppr-ui/.env.production
```diff
@@ -0,0 +1 @@
+VUE_APP_PATH = ppr
```
thorwolpert, line 1: do we need this?

ppr-ui/nginx.conf
```diff
@@ -0,0 +1,38 @@
...
+ server {
+   listen 8080;
+   server_name _;
+
+   index index.html;
+   error_log /dev/stdout info;
+   access_log /dev/stdout;
+   root /app;
+
+   location /ppr {
```
thorwolpert, line 27: only being set because of the VUE_APP_PATH

ppr-ui/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "eslint-plugin-vue": "^6.2.2",
+    "sass": "^1.26.5",
+    "sass-loader": "^8.0.2",
+    "typescript": "~3.9.3",
+    "vue-class-component": "^7.2.3",
+    "vue-cli-plugin-vuetify": "^2.1.0",
+    "vue-template-compiler": "^2.6.11",
+    "vuex-module-decorators": "^0.16.1"
+  },
+  "eslintConfig": {
```
thorwolpert, line 58: as there is a separate .eslintrc.js, shouldn't everything be moved there?

thorwolpert (commented):
nice, and good leverage of existing apps.
would it be worth removing the couple of mixins now? if os, we could keep this as a proj-stub!
we might want to poke Jeremy on the main panels for grid layout, as the existing systems are not responsive at all, and I think this one will need to be.

PR #7, opened 2021-02-09
6283 UI login

kialj876:
ticket: https://github.com/bcgov/entity/issues/6283

- would like to add some better testing around login / feature flags, but I didn't want to spend too much time trying to figure out how in this ticket (I tried a few things to test the redirect, etc., but it didn't work how I expected in jest)
  - our other apps don't have tests around this so I am leaving it as is (maybe we can have a ticket to figure out login/ff testing for all of them)


doug-lovett (approved):
I scanned the changes, as Thor has a much better handle on the UI framework I am approving with his review the one that really counts.

thorwolpert (approved):
thanks for enabling the ldconfig !

PR #8, opened 2021-02-10
6157 develop UI ppr api client call for search.

doug-lovett:


kialj876 (commented):
Very nice and clear, thank you for all the comments. The lint gives a warning for the token on your test (max-len), can you just add an eslint-disable for that line? 
I'm also getting this when I run your test: `Error: Cross origin http://localhost forbidden` causing it to fail. I'm assuming I just need the correct api key set, can you message me that?

kialj876:
> Very nice and clear, thank you for all the comments. The lint gives a warning for the token on your test (max-len), can you just add an eslint-disable for that line?
> I'm also getting this when I run your test: Error: Cross origin http://localhost forbidden causing it to fail. I'm assuming I just need the api key set, can you message me that?

other than that this looks great to me

PR #11, opened 2021-02-19
serial number search

kialj876:
ticket #: bcgov/entity#6153
- see ticket for details


doug-lovett (approved):
Scanned the PR, my only comment is: should the API key be stored somewhere more secure? Is ppr-ui/public/config/configuration.json the right place long-term?

kialj876:
> Scanned the PR, my only comment is: should the API key be stored somewhere more secure? Is ppr-ui/public/config/configuration.json the right place long-term?

This public config file is for local use only and will be getting overwritten by the config we set in openshift (only viewable/editable to us). For now I put the key in here since it had already been added elsewhere, but in the future we would not commit any secret values that we define in the config

ppr-ui/package.json
```diff
@@ -12,6 +12,7 @@
   "dependencies": {
     "@sentry/browser": "^6.0.4",
     "@sentry/integrations": "^6.0.4",
+    "@vue/composition-api": "^1.0.0-rc.1",
```
thorwolpert, line 15: woo hoo!!

ppr-ui/src/components/search/Search.vue
```diff
@@ -1,44 +1,148 @@
-<template>
+<template `>
```
thorwolpert, line 1: ` <- is grumpy in my IDE

ppr-ui/src/enums/searchTypes.ts
```diff
@@ -1,8 +1,16 @@
...
+export enum APISearchTypes {
   AIRCRAFT = 'AIRCRAFT_DOT',
   BUSINESS_DEBTOR = 'BUSINESS_DEBTOR',
-  INDIVIDCUAL_DEBTOR = 'INDIVIDUAL_DEBTOR',
+  INDIVIDUAL_DEBTOR = 'INDIVIDUAL_DEBTOR',
   MHR_NUMBER = 'MHR_NUMBER',
   REGISTRATION_NUMBER = 'REGISTRATION_NUMBER',
   SERIAL_NUMBER = 'SERIAL_NUMBER'
 }
+export enum UISearchTypes {
```
thorwolpert, line 9: is this something that we'd leverage I18N on?

thorwolpert (approved):

ppr-ui/src/enums/searchTypes.ts
kialj876: It could be later I guess. It was more just a way to define what the UI is displaying each category as vs what the api is expecting for now

ppr-ui/src/components/search/Search.vue
kialj876: ah yes I will have that removed in my next PR

PR #14, opened 2021-02-25
added results data table

kialj876:

ticket #: bcgov/entity#6154


ppr-ui/src/components/results/Results.vue
```diff
@@ -0,0 +1,180 @@
...
+        <b>selected: {{ selectedLength }}</b>
+      </v-col>
+      <v-col v-else>
+        {{ totalResultsLength }} registrations found
+      </v-col>
+    </v-row>
+    <v-row v-if="totalResultsLength !== 0" no-gutters>
+      <v-container fluid no-gutters>
+        <v-row no-gutters>
+          <v-col class="auto">
```
cameron-eyds: Is this meant to be `cols=auto`?

ppr-ui/src/components/results/Results.vue
```diff
@@ -0,0 +1,180 @@
...
+
+<script lang="ts">
+import { computed, defineComponent, reactive, toRefs, watch } from '@vue/composition-api'
+import { useGetters } from 'vuex-composition-helpers'
+
+import { tableHeaders } from '@/resources'
+import { SearchResponseIF, SearchResultIF, TableHeadersIF } from '@/interfaces' // eslint-disable-line no-unused-vars
+import { MatchTypes } from '@/enums'
+
+export default defineComponent({
```
cameron-eyds, line 57: For awhile now we have been writing class based components opposed to using object notation. 
This approach certainly isn't wrong and i think it's more or less a preference.
I think we opted for class based components due to the way it worked well with TypeScript. 

Here is a really simple example that utilizes a lot of Vue Property Decorators, it just makes for very easy to read components.

https://github.com/bcgov/bcrs-shared-components/blob/master/src/components/DetailComment/DetailComment.vue

Not saying you should change anything, just making aware of the approach we are currently taking in our other UI's is a little different, utilizing `vue-property-decorator`.

ppr-ui/src/components/results/Results.vue
```diff
@@ -0,0 +1,180 @@
...
+  width: 100%;
+  height: 8rem;
+}
+// .result-header {
+//   font-size: 0.825rem !important;
+//   color: $gray7 !important;
+// }
+// .results-table {
+//   height: 500px;
+// }
```
cameron-eyds: For future styling?

ppr-ui/src/components/results/Results.vue
```diff
@@ -0,0 +1,180 @@
...
+.v-select-list {
```
cameron-eyds: Potentially some unused classes/id's that could removed if they are not used within the template.

ppr-ui/src/views/dashboard.vue
```diff
@@ -1,42 +1,82 @@
...
-      <v-col cols="12" class="search-info pt-2">
-        <span>
-          Select a search category and then enter a value to search.
-          <b>Note:</b>
-          Each search incurs a fee (including searches that return no results).
-        </span>
+    <v-row>
+      <v-col>
+        <v-row no-gutters>
+          <v-col cols="12" class="search-title">
```
cameron-eyds, line 6: Right on! 
The biggest reason name request wasn't responsive was due to the col definitions, or lack of.

ppr-ui/src/views/dashboard.vue
```diff
@@ -1,42 +1,82 @@
...
+              Select the registrations you want to download in a printable PDF.
+              Your search results, selected registrations, and PDF are automatically saved to My Searches.
+            </span>
+            <span v-else>
+              Your search results and PDF are automatically saved to My Searches.
+            </span>
+          </v-col>
+        </v-row>
+        <v-row v-if="getSearchResults">
+          <result :data="getSearchResults">
```
cameron-eyds: You can self close components if you prefer. ie `<result :data="getSearchResults" />`

ppr-ui/src/views/dashboard.vue
```diff
@@ -1,42 +1,82 @@
...
 import { SearchResponseIF } from '@/interfaces' // eslint-disable-line no-unused-vars
 
 @Component({
   components: {
-    Search
+    Search,
+    Result
   }
 })
 export default class Dashboard extends Vue {
```
cameron-eyds, line 72: Oh, this component is class based and follows our other architecture! 
Ok, scratch that earlier comment you are definitely aware :) 
So why the other structure in the results component?

ppr-ui/src/views/dashboard.vue
```diff
@@ -51,6 +91,42 @@ export default class Dashboard extends Vue {
...
+      let min = `0${searchDate.getMinutes()}`
+      let sec = `0${searchDate.getSeconds()}`
+      if (hour.length > 2) hour = hour.slice(1)
+      if (min.length > 2) min = min.slice(1)
+      if (sec.length > 2) sec = sec.slice(1)
+      const datetime = `${hour}:${min}:${sec}`
+      return 'saved ' + moment(searchDate).format('MMMM D, Y') + ` ${datetime} ${timezone}`
+    }
+    return ''
+  }
```
cameron-eyds: If this is something you might re-use i'd probably make it common within a mixin (or perhaps composition api?! not sure how works yet), and just pass the `getSearchResults` to it.

ppr-ui/src/views/dashboard.vue
```diff
@@ -51,6 +91,42 @@ export default class Dashboard extends Vue {
...
+
+  private get searchValue (): string {
+    const searchResult = this.getSearchResults
+    if (searchResult) {
+      return searchResult.searchQuery.criteria.value
```
cameron-eyds: Might not hurt to optional chain here, to protect yourself in the event `searchResults` is present but `searchQuery` is not. 
ie `return searchResult.searchQuery?.criteria?.value`

ppr-ui/src/views/dashboard.vue
```diff
@@ -109,14 +180,21 @@ export default class Dashboard extends Vue {
 }
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
```
cameron-eyds, line 170: I think scoped is ok here. 
Are you trying to modify some styling the is nested in a Vuetify component?

If `.v-simple-checkbox .theme--light.v-icon` was the issue, you can always keep the components styling scoped and use `::v-deep .v-simple-checkbox .theme--light.v-icon` to modify the specific style.

cameron-eyds (commented):
A lot of good things here! Just had a few questions/comments

doug-lovett (approved):
Approving again as a quick scan, given Cameron's thorough review.

ppr-ui/src/views/dashboard.vue
kialj876: ah okay sounds good I will update that tomorrow morning

ppr-ui/src/components/results/Results.vue
kialj876: yep! Will update

ppr-ui/src/components/results/Results.vue
kialj876: answered below

ppr-ui/src/components/results/Results.vue
kialj876: oops will take this out!

ppr-ui/src/components/results/Results.vue
kialj876: yes definitely

ppr-ui/src/views/dashboard.vue
kialj876: ah okay sounds good will update as it looks cleaner that way

ppr-ui/src/views/dashboard.vue
kialj876: yeah I know :P this was one of the things I was thinking we need to discuss with more devs. The type inference works really well using the composition api, but when you combine it with the class based component stuff it gets a bit buggy and loses some of the typing (at least when I was implementing it - there are probably ways to get around what my issues were). I was also thinking that while trying the composition api (at least in the beginning) we should be doing it in the way the vue devs were expecting it to be used so we don't run into weird problems that are 'mostly' fixed by imported libraries. With that thought and since we're only using it in 2 small components right now I thought I would go ahead this way as a proof of concept, but if needed we can alter these back into the class component style pretty quickly. Definitely worthy of a discussion later I think though

ppr-ui/src/views/dashboard.vue
kialj876: Yes good point I will update this too

ppr-ui/src/views/dashboard.vue
kialj876: Ah good point! Will update

kialj876:
> A lot of good things here! Just had a few questions/comments

Thanks Cam!

ppr-ui/src/views/dashboard.vue
kialj876: Actually having it scoped prevents my custom classes in the data-table from rendering. I can see them being applied in the html, but they don't register any of the css. The weird thing is they don't get overwritten, they just don't show that they have any css associated with the class. I tried using the ::v-deep but it doesn't solve it so I am leaving it unscoped for now. I removed the unneeded classes from there so all the defined classes are unique to this component now

cameron-eyds (approved):

ppr-ui/src/views/dashboard.vue
thorwolpert: definitely something to discuss and check new best practices of those moving to Vue3, where I do see lots of discussion on class based components.

ppr-ui/src/views/dashboard.vue
thorwolpert: the latest best practices doc suggests using **module** instead of **scoped** to fully avoid css collisions.

thorwolpert (commented):
just a comment on scope

PR #19, opened 2021-03-09
autocomplete component added, tests, etc.

kialj876:

ticket: bcgov/entity/issues/6439
- see ticket for details


doug-lovett (approved):

ppr-ui/src/components/search/AutoComplete.vue
```diff
@@ -0,0 +1,100 @@
...
+    })
+
+    return {
+      ...toRefs(localState)
+    }
+  }
+})
+</script>
+
+<style lang="scss" scoped>
```
thorwolpert, line 85: maybe we stick to module on everything instead of scoped.
re: best-practices - vue core developers

ppr-ui/src/utils/axios-von.ts
```diff
@@ -0,0 +1,14 @@
+import Axios from 'axios'
```
thorwolpert, line 1: is this so setup doesn't collide with axios to the rest of the services?

ppr-ui/src/utils/config-helper.ts
```diff
@@ -65,6 +65,10 @@ export async function fetchConfig (): Promise<any> {
   sessionStorage.setItem('KEYCLOAK_CONFIG_PATH', keycloakConfigPath)
   console.info('Set Keycloak Config Path to: ' + keycloakConfigPath)
 
+  const vonApiUrl: string = response.data.VON_API_URL
+  sessionStorage.setItem('VON_API_URL', vonApiUrl)
+  console.log('Set VON API URL to: ' + vonApiUrl)
```
thorwolpert, line 70: doesn't this bark / warn for prod builds?

thorwolpert (commented):
just a couple of comments
i like the bigger names in the tests, less likely to trip over the mocks in the future +1

ppr-ui/src/utils/axios-von.ts
kialj876: yeah we could have all of these in one file, but we'd still need to define/export the different ones since they need slightly different things (or I could put some work into making these more generic, but that could give some grief down the line depending on what future calls we have, although even then it would probably be minor changes needed). Not sure what the best practice here is so I followed what we already have going (different file for each service)

ppr-ui/src/utils/config-helper.ts
kialj876: I don't get that when I run it or run the tests/lint is this happening for you? Or are you expecting it to complain for some reason?

ppr-ui/src/components/search/AutoComplete.vue
kialj876: Okay I will update all the components to use module in my next PR

ppr-ui/src/utils/axios-von.ts
kialj876: I may have misunderstood your question haha lmk

PR #53, opened 2021-04-23
optional drop of functions

jordiwes:
*Issue #:* /bcgov/entity###

*Description of changes:*


thorwolpert (changes requested):
Why are DROPs being added / changed?
if you are worried about seeing drop errors in the script.
Do CREATE OR REPLACE instead.

jordiwes:
Drops were already there... so I made them conditional. I change change the script to CREATE OR REPLACE if Oracle supports it

PR #57, opened 2021-04-29
PPR API  API DB Healthcheck

doug-lovett:

*Issue #:* /bcgov/entity###

*Description of changes:*
Add /ops/healthz db connection health check.
Disable schema validation for drafts to allow incomplete saves.
Turn off debug db connection logging.
Include CROWNLIEN registration class in financing statement queries.

### DIFF
--- a/oracle-db/scripts/create_views_ddl.sql
+++ b/oracle-db/scripts/create_views_ddl.sql
@@ -9,7 +9,7 @@ SELECT r.registration_type_cd,
   FROM registration r, financing_statement fs, registration r2
  WHERE r2.financing_id = r.financing_id
    AND r.financing_id = fs.financing_id
-   AND r.registration_type_cl IN ('PPSALIEN', 'MISCLIEN')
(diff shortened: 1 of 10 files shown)


thorwolpert (changes requested):
This has a lot of stuff that is not related to the OPs end-point.
The ope end-point is missing the mount here as well
https://github.com/bcgov/ppr/blob/main/ppr-api/src/ppr_api/resources/__init__.py#L69

doug-lovett:
The OPS mount config is on line 58 and has a different base path.
And you are right the PR has more than just the heartbeat change. It is cumulative because of the lag on reviewing PR's. Case in point: I just added 7340 changes which I completed after opening this PR. And I already have UI changes for another PR  ready to go.

thorwolpert:
so can we split this into the ops end point, and what ever the other stuff is?

doug-lovett:
Close this PR and open 2 new ones?

PR #63, opened 2021-05-03
7259 add collateral

jordiwes:
*Issue #:* /bcgov/entity#7259

*Description of changes:*

This is the first draft of the add collateral components. The following is missing and will be addressed in the next PR:
- validation of fields
- tests
- some UI tweaks to match ux pin


thorwolpert (changes requested):
This needs to change in the package file 
"vue-test-utils-helpers": "https://github.com/bcgov/vue-test-utils-helpers.git",

jordiwes:
I did update the vue test utils package

doug-lovett (approved):

jordiwes:
I have added some tests

ppr-ui/src/components/collateral/EditCollateral.vue
```diff
@@ -0,0 +1,215 @@
...
+    activeIndex: {
+      type: Number,
+      default: -1
+    },
+    invalidSection: {
+      type: Boolean,
+      default: false
+    }
+  },
+  emits: ['addEditVehicle', 'resetEvent'],
```
kialj876, line 148: is this used at all or is it for making it easier for us to see what emit options there are? I'm wondering if it should get added to all the composition components I made previously

kialj876 (approved):
Looks good to me, I just had one question for my own interest

ppr-ui/src/components/collateral/EditCollateral.vue
jordiwes: I think this is necessary for an emit in the composition api? based on my research. So I think it's used

PR #66, opened 2021-05-05
PPR UI step 1  review and confirm

doug-lovett:

*Issue #:* /bcgov/entity#7266

*Description of changes:*
Add step 1 read only summary to step 4 review and confirm.
Clean up console errors.
Partial UI design updates for the stepper.

### DIFF
--- a/ppr-ui/src/components/registration/RegistrationLengthTrust.vue
+++ b/ppr-ui/src/components/registration/RegistrationLengthTrust.vue
@@ -1,5 +1,40 @@
 <template>
-  <v-container fluid no-gutters class="white pa-6">
+  <v-container fluid no-gutters class="white pa-0"  v-if="summaryView">
+    <v-card flat id="length-trust-summary">
+      <v-row no-gutters :class="[$style['length-trust-summary-header'], pa-2]">
+        <v-col cols="auto">
+          <v-icon color="#38598A">mdi-calendar-clock</v-icon>
+          <label class="length-trust-title pl-3"><strong>Registration Length and Trust Indenture</strong></label>
+        </v-col>
+      </v-row>
+      <v-row no-gutters v-if="showErrorSummary" :class="[$style['length-trust-invalid-message'], 'pa-6']">
+        <v-col cols="auto">
+          <v-icon color="#D3272C">mdi-information-outline</v-icon>
+          This step is unfinished
+          <router-link id="router-link-length-trust" class="length-trust-invalid-link" :to="{ path: '/length-trust' }">
+            Return to this step to complete it.
+          </router-link>
+        </v-col>
+      </v-row>
+      <v-row no-gutters class="length-trust-content pa-4">
+        <v-col cols="4" class="length-trust-title">
+          <strong>Registration Length</strong>
+        </v-col>
+        <v-col cols="auto">
+          {{lengthSummary}}
+        </v-col>
+      </v-row>
+      <v-row no-gutters class="length-trust-content pl-4 pb-2">
+        <v-col cols="4" class="length-trust-title">
+          <strong>Trust Indenture</strong>
+        </v-col>
+        <v-col cols="auto">
+          {{trustIndentureSummary}}
+        </v-col>
+      </v-row>
+    </v-card>
+  </v-container>
+  <v-container fluid no-gutters class="white pa-6"  v-else>
     <v-row no-gutters>
       <v-col cols="2" :class="$style[length-trust-label]">
         Registration Length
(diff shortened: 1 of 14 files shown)


ppr-ui/src/components/registration/RegistrationLengthTrust.vue
```diff
@@ -1,5 +1,40 @@
 <template>
-  <v-container fluid no-gutters class="white pa-6">
+  <v-container fluid no-gutters class="white pa-0"  v-if="summaryView">
+    <v-card flat id="length-trust-summary">
+      <v-row no-gutters :class="[$style['length-trust-summary-header'], pa-2]">
+        <v-col cols="auto">
+          <v-icon color="#38598A">mdi-calendar-clock</v-icon>
+          <label class="length-trust-title pl-3"><strong>Registration Length and Trust Indenture</strong></label>
```
kialj876: is the `length-trust-title` css actually getting applied here? If so I guess it's because its nested under `length-trust-summary-header` or did you find another way to avoid the `$style` notation?

ppr-ui/src/components/registration/RegistrationLengthTrust.vue
```diff
@@ -176,4 +234,55 @@ export default defineComponent({
 .length-trust-label {
   font-size: 0.875rem;
 }
+
+.length-trust-summary-header {
+  display: flex;
+  background-color: $gray3;
+  padding: 1.25rem;
```
kialj876: I don't know what the best practice is, but I generally find it easier/simpler to adjust things on the UI when most of the padding/margins are defined in the html with the `p/m..-..` classes. I see you've defined padding in here and with `pa-2` in the html haha. Can you consolidate them so the padding is defined in one place for this class and any other ones that do this? If there isn't a pa-x value that works then just having the value here is good I think

kialj876 (commented):
overall looks good, but I just had a couple questions

ppr-ui/src/components/registration/RegistrationLengthTrust.vue
```diff
@@ -144,7 +202,7 @@ export default defineComponent({
       lengthTrust.trustIndenture = val
       setLengthTrust(lengthTrust)
     })
-    watch(() => localState.lifeInfinite, (val: boolean) => {
+    watch(() => localState.lifeInfinite, (val: string) => {
```
jordiwes, line 205: why change to a string here?

ppr-ui/tests/unit/ReviewConfirm.spec.ts
```diff
@@ -106,3 +108,87 @@ describe('ReviewConfirm new registration component', () => {
...
+    await Vue.nextTick()
+    wrapper.find(nextBtn).trigger('click')
+    await Vue.nextTick()
+    await Vue.nextTick()
+    expect(wrapper.findComponent(ReviewConfirm).exists()).toBe(true)
+    expect(wrapper.findComponent(Stepper).exists()).toBe(true)
+    expect(wrapper.findComponent(Tombstone).exists()).toBe(true)
+    expect(wrapper.findComponent(RegistrationFee).exists()).toBe(true)
+    expect(wrapper.findComponent(RegistrationLengthTrust).exists()).toBe(true)
+    // expect(wrapper.findComponent(RegistrationLengthTrust).isSummary).toBe(true)
```
jordiwes, line 162: should these be removed or commented back in?

ppr-ui/src/components/registration/RegistrationLengthTrust.vue
doug-lovett: Updated, removed length-trust-title as it's only padding left. I started with a create confirm template and this came with it.

ppr-ui/src/components/registration/RegistrationLengthTrust.vue
doug-lovett: Updated, moved from css to html.

ppr-ui/src/components/registration/RegistrationLengthTrust.vue
doug-lovett: The radio buttons are actually a string. As a string the initial state of both deselected can be represented as ''.

ppr-ui/tests/unit/ReviewConfirm.spec.ts
doug-lovett: I left them in to circle back later. The assertions are not yet working and I didn't want to hold up the PR.

kialj876 (approved):

PR #71, opened 2021-05-12
PPR api client party code changes

doug-lovett:

*Issue #:* /bcgov/entity#NA

*Description of changes:*
1. API updates from Bob's latest db model changes for client party codes.
2. Add endpoint to support UI party code lookup by getting party codes by head office name or code. 


jordiwes (approved):
LGTM, but Thor should review :)

PR #75, opened 2021-05-19
validation fix for 7 digit registration  numbers

jordiwes:
*Issue #:* /bcgov/entity#7530

*Description of changes:*
validation fix for registration number


thorwolpert (changes requested):
The test need to be updated, as they fail now.

jordiwes:
I even said that in the meeting and then I forgot. They are there now :)

thorwolpert (approved):

PR #77, opened 2021-05-21
7346 add collateral updates

jordiwes:
*Issue #:* /bcgov/entity#7346

*Description of changes:*
- collateral validations part 1
- collateral ux part 1
- entire app ux part 2

*To do:*
- clean up validations
- clean up ux
- add tests


kialj876 (changes requested):
hey Steph, I'm just checking out your PR -- is it ready for review? I'm getting some errors when I run it locally (just type errors I think?). Also, I like the idea of having tombstone in App, but it is broken now. Is it needed in this PR? It also might be worth keeping the base tombstone component and having another one in app.vue that uses it and handles the breadcrumb logic (that way we could potentially re-use the base tombstone component in other apps later / make it a shared component)

I think you mentioned you were still working on new tests, but existing ones currently fail too. Also CI fails on the npm install currently

jordiwes:
Yes, I do not understand why this fails:
npm ERR! /usr/bin/git ls-remote -h -t ssh://git@github.com/bcgov/vue-test-utils-helpers.git
npm ERR! 
npm ERR! Warning: Permanently added the RSA host key for IP address '140.82.114.3' to the list of known hosts.


But I have a bunch of summary stuff to add, so I will delete this PR. I had to add ts ignores for those lines because it wasn't picking up the toRefs from the composable... anyhow, stay tuned!

PR #78, opened 2021-05-28
7436 - collateral validation, ux changes, and overall ux changes

jordiwes:
*Issue #:* /bcgov/entity#7436 and /bcgov/entity#7358
*Description of changes:*

- collateral validation
- review & summary for collateral
- many ux updates
- refactoring for composition api
- tests


ppr-ui/src/components/collateral/EditCollateral.vue
```diff
@@ -147,61 +164,49 @@ export default defineComponent({
...
-    })
-    const getYears = computed(function () {
-      const year = new Date().getFullYear()
-      return Array.from(
-        { length: year - 1900 },
-        (value, index) => 1901 + index
-      ).reverse()
-    })
+    const {
+      // @ts-ignore - returned by toRef
```
kialj876, line 170: Do you know if you can return the whole object instead of the toRef, and then dereference it in the return for setup or not? Might be a way to get around the lint unless that breaks the reactivity

kialj876 (commented):
Hey Steph, this is looking very good. The npm install in the CI is failing though -- is this a known issue or does it need a fix before this can merge? One other question, but I think this is good as long as the CI/CD will build it and the tests/lint are passing

jordiwes:
@kialj876  I do not know how to fix the CI/CD :( Hoping someone does!

ppr-ui/src/components/collateral/EditCollateral.vue
jordiwes: Yes, I need the reactivity........ trust me I tried it 10 other ways

doug-lovett (approved):
Running locally UX changes look great.
Here are a few serial collateral comments:
Validation:
Add vehicle MH type, MHR number entered but no serial number, the serial number error message is "Enter the Manufactured Home Registration Number". MHR field validation is showing up under the serial number.
Step 4 Review and Confirm
MH number column is truncated. Vehicle Type descriptions are truncated.

jordiwes:
@doug-lovett , I did update that mh number validation to show up on the right field. I also squished that table as best as possible to fit all of the columns

doug-lovett (approved):
Latest commit with the validation and review and confirm changes look good. One additional comment: my understanding for the MH type is that one of MHR number or serial number is required? The app requires an MH number when a serial number is present.

kialj876 (approved):

PR #81, opened 2021-06-02
left over fixes

jordiwes:
*Issue #:* /bcgov/entity#7358

*Description of changes:*
- small tweaks for UI

### DIFF
--- a/ppr-ui/src/App.vue
+++ b/ppr-ui/src/App.vue
@@ -134,7 +134,7 @@ export default class App extends Mixins(AuthMixin) {
   }
 
   private get dashboardURL (): string {
-    return window.location.origin + 'dashboard'
+    return window.location.origin + '/dashboard'
   }
 
   /** The login URL. */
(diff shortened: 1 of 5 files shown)


doug-lovett (approved):

ppr-ui/src/App.vue
```diff
@@ -134,7 +134,7 @@ export default class App extends Mixins(AuthMixin) {
   }
 
   private get dashboardURL (): string {
-    return window.location.origin + '/dashboard'
+    return window.location.origin + 'dashboard'
```
kialj876: I think localtion.origin will be '.../ppr' so won't this cause an error without the slash?

kialj876 (commented):
Looks good, but just need to verify the path is still good -- I don't think this is something that would show locally

ppr-ui/src/App.vue
jordiwes: changed the wrong slash

kialj876 (approved):

PR #90, opened 2021-06-04
API registration payment, search selection schema validation udpate.

doug-lovett:

*Issue #:* /bcgov/entity#7639

*Description of changes:*
- Bug 7639 Update schema validation on save search selection to include legacy registration types.
- Add payment for financing, renewal, change, and amendment statement registrations. 
- Payment pay-api request errors return 402 http status code
- Create non-financing registrations return 201 instead of 200 to match the spec.


ppr-api/src/ppr_api/resources/financing_statements.py
```diff
@@ -261,17 +292,30 @@ def post(registration_num):
...
-            return statement.json, HTTPStatus.OK  # CREATED
+            # Set up registration and pay.
+            registration = setup_registration_pay(request_json,
+                                                  model_utils.REG_CLASS_CHANGE,
+                                                  statement,
+                                                  registration_num,
+                                                  account_id)
+            invoice_id = str(registration.pay_invoice_id)
+
+            # Try to save the change statement: failure throws an exception.
```
jordiwes: Can/should the registration save be abstracted to a function?

doug-lovett: Great suggestion, I have made this change in a second commit.

PR #93, opened 2021-06-11
7690 address component

kialj876:
*Issue #:* /bcgov/entity#7690

*Description of changes:*
- composition api composable of baseAddress


kialj876:
fyi: I left the styling as scoped because we are planning to make this a shared component. Until we can figure out how to adjust the 'v-deep' styling inside the component using module I think it will need to stay as scoped. Either that or once we have design changes for it we can possibly remove the v-deep styling (without new specs I didn't want to adjust any of the current styling for the component)

thorwolpert (approved):

ppr-ui/src/composables/address/BaseAddress.vue
```diff
@@ -0,0 +1,255 @@
...
+
+    <!-- Edit fields -->
+    <v-expand-transition>
+      <v-form v-if="editing" ref="addressForm" name="address-form" lazy-validation>
+        <div class="form__row">
+          <!-- NB1: AddressComplete needs to be enabled each time user clicks in this search field.
+               NB2: Only process first keypress -- assumes if user moves between instances of this
+                   component then they are using the mouse (and thus, clicking). -->
+          <v-text-field autocomplete="chrome-off"
+                        :name="Math.random()"
```
jordiwes, line 28: why random :)

ppr-ui/src/composables/address/factories/validation-factory.ts
```diff
@@ -0,0 +1,121 @@
+/* eslint-disable */
+import { computed, ComputedRef, ref, Ref, toRefs, reactive } from '@vue/composition-api'
+import { maxLength, required } from 'vuelidate/lib/validators'
+
+import { AddressIF, SchemaIF } from '@/composables/address/interfaces'
+
+const useVuelidate = require("@vuelidate/core").default;
+
+// export const maxLength = (param: number) => helpers.withParams(
```
jordiwes: remove?

jordiwes (commented):
I prefer scoped styling!!
There are a few build errors:
operty 'addressCountry' does not exist on type 'Ref<AddressIF>'.

kialj876:
> I prefer scoped styling!!
> There are a few build errors:
> operty 'addressCountry' does not exist on type 'Ref'.

awe darn those build errors are from the auto linting changes. I made some last minute changes too, I'll go through those.

-- ew I definitely broke the labels at some point and didn't notice

ppr-ui/src/composables/address/BaseAddress.vue
kialj876: that's what they put in the other common one -- I think it's so that it changes every time to prevent the chrome autocomplete?

ppr-ui/src/composables/address/factories/validation-factory.ts
kialj876: I was going to leave that just in case vuelidates next update makes this work (this is what they show in their documentation to allow me access to the $params of a custom validator). As of now it doesn't work though so I will remove it haha

jordiwes (approved):

PR #96, opened 2021-06-15
7138 add secured party

jordiwes:
*Issue #:* /bcgov/entity#7138

*Description of changes:*
- add business or individual secured party
- add address component to secured party and debtor
- mocked in adding registering party as secured party
- still needs validations defined and added
- tests

### DIFF
--- a/ppr-ui/src/components/parties/EditDebtor.vue
+++ b/ppr-ui/src/components/parties/EditDebtor.vue
@@ -132,14 +132,13 @@
                 <label class="general-label">Address</label>
               </v-col>
             </v-row>
-            <!-- <base-address ref="regMailingAddress"
+            <base-address ref="regMailingAddress"
                     id="address-debtor"
-                    :address="currentDebtor.address"
+                    v-model="currentDebtor.address"
                     :editing="true"
                     :schema="addressSchema"
-                    @update:address="updateAddress($event)"
                     @valid="updateValidity($event)"
-                  /> -->
+                  />
 
             <v-row>
               <v-col>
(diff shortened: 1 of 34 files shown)


ppr-ui/src/components/parties/composables/useDebtor.ts
```diff
@@ -35,6 +39,7 @@ export const useDebtor = (props, context) => {
   })
 
   const getDebtor = () => {
+    console.log(props.activeIndex)
```
kialj876: left in in a few log statements in here, pls remove

ppr-ui/src/components/parties/PartySearch.vue
```diff
@@ -0,0 +1,68 @@
...
+import {
+  defineComponent,
+  reactive,
+  toRefs
+} from '@vue/composition-api'
+
+export default defineComponent({
+  emits: ['showSecuredPartyAdd'],
+  setup (props, context) {
+    const localState = reactive({})
```
kialj876, line 47: can localState be removed or are you expecting to add to this later?

ppr-ui/src/components/parties/EditDebtor.vue
```diff
@@ -132,14 +132,13 @@
                 <label class="general-label">Address</label>
               </v-col>
             </v-row>
-            <!-- <base-address ref="regMailingAddress"
+            <base-address ref="regMailingAddress"
                     id="address-debtor"
-                    :address="currentDebtor.address"
+                    v-model="currentDebtor.address"
                     :editing="true"
                     :schema="addressSchema"
```
kialj876, line 139: I talked to kaine about the validation and he said it's okay if we come back to it later. I think passing in the schema will give validation that is broken so it's better not to pass it in so that the validation is ignored by the address component

kialj876 (commented):
Hey Steph, this looks very good I just had a few small comments

ppr-ui/src/components/parties/EditDebtor.vue
jordiwes: If I don't pass it in, will it get updated in the state model?

ppr-ui/src/components/parties/PartySearch.vue
jordiwes: This is just the structure, the actual search is another ticket, so probably leave it in :)

ppr-ui/src/components/parties/EditDebtor.vue
kialj876: it should default to null and skip all the schema things -- that's what we want for now right?

kialj876 (approved):

PR #120, opened 2021-07-12
small adjustments for step 2- step 4

jordiwes:
*Issue #:* /bcgov/entity#7268

*Description of changes:*
- adjustments made based on Tracey's feedback

### DIFF
--- a/ppr-ui/src/composables/useParty.ts
+++ b/ppr-ui/src/composables/useParty.ts
@@ -30,6 +30,9 @@ export const useParty = () => {
       party.address.region +
       ' ' +
       party.address.postalCode
+    if (party.address.country) {
+      address = address + '<br>' + party.address.country
+    }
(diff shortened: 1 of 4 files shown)


ppr-ui/src/composables/useParty.ts
```diff
@@ -30,6 +30,9 @@ export const useParty = () => {
       party.address.region +
       ' ' +
       party.address.postalCode
+    if (party.address.country) {
+      address = address + '<br>' + party.address.country
```
kialj876, line 34: does this need a `</br>` at the end?

jordiwes: Don't think so... it already has cell padding

kialj876 (approved):

thorwolpert (approved):

PR #124, opened 2021-07-15
API changes to improve search performance.

doug-lovett:

*Issue #:* /bcgov/entity#8037

*Description of changes:*
- Add parties.email_address as a PPR UI requirement.
- Add search_requests.user_id as an audit requirement.
- Update business debtor search and individual debtor search to improve performance: switch to trigram algorithm https://www.postgresql.org/docs/11/pgtrgm.html
- Search add index parties.middle_initial, serial_collateral.mhr_number
- Add indexes on all foreign keys (explicitly required) to improve performance.
- Defect 8192 update search by mhr number to not match on search_vin 


kialj876 (approved):

kialj876:
Looks good to me! I approved, but the linting should be fixed for 3.8 -- not sure if we're being strict for 3.9 yet or not

PR #135, opened 2021-07-26
API more changes to support UI my registrations table

doug-lovett:

*Issue #:* /bcgov/entity#7852

*Description of changes:*
- Add registering name column. Modify data model add drafts.user_id.
- Add get registrations list "collapse" query parameter. Collapse if true returns a hierarchical parent/child representation of financing statements and registrations.
- Add get individual financing statement "current" query parameter. Current if true returns the current view of the financing statement.
- Report add legacy "RC" registry correction change type to report change type descriptions.
- Improve performance of report data generation: do not use pycountry to get CA, US descriptions.
- Improve performance of debtor individual search: always use indexed columns.     


ppr-api/src/ppr_api/models/draft.py
```diff
@@ -58,6 +58,7 @@ class DraftTypes(Enum):
     draft = db.Column('draft', db.JSON, nullable=False)
     registration_number = db.Column('registration_number', db.String(10), nullable=True)
     update_ts = db.Column('update_ts', db.DateTime, nullable=True)
+    user_id = db.Column('user_id', db.String(1000), nullable=True)
```
thorwolpert, line 61: should this be a foreign key to make it easier to link?

ppr-api/src/ppr_api/resources/drafts.py
```diff
@@ -17,7 +17,7 @@
 
 from http import HTTPStatus
 
-from flask import request, jsonify
+from flask import request, jsonify, g
```
thorwolpert: doesn't the linter complain about alphabetical order here?

thorwolpert (commented):
couple of small things

ppr-api/src/ppr_api/resources/drafts.py
doug-lovett: No complaints but sorted anyway as a second commit.

ppr-api/src/ppr_api/models/draft.py
doug-lovett: I considered it, but the direct api consumers may submit requests with no users record, throwing a foreign constraint violation. With a constraint, we need to check/insert every time on high volume get drafts/get financing statements/get registrations requests. So I left the constraint out and use subqueries which allow for no users record.

PR #143, opened 2021-08-05
8422 API reg number search non-financing reg type

doug-lovett:

*Issue #:* /bcgov/entity#8422

*Description of changes:*
- When searching by registration number on a non-financing registration, return the financing statement (base registration) registration type, not the discharge, renewal...registration type.
- 8422 added: include debtor birth date in summary search results for an individual debtor search.


kialj876 (approved):

kialj876:
Looks good, but should update the file that the linter is failing - even if you just get it to ignore that one since it's just over by 1

doug-lovett:
@kialj876  Thanks for catching the linter complaint - resolved in last commit.

PR #147, opened 2021-08-06
API final search results report change from UX assurance.

doug-lovett:

*Issue #:* /bcgov/entity#7885

*Description of changes:*
- 7885 Financing Statement remove Registration Act from reports.
- 8422 Search by registration number workaround for legacy migration records where registrations.base_reg_number is null for non financing statement registrations.
- 8483 Fix to ignore case in search criteria when searching by registration number, serial number, aircraft DOT.

### DIFF
--- a/ppr-api/src/ppr_api/models/search_request.py
+++ b/ppr-api/src/ppr_api/models/search_request.py
@@ -114,7 +114,7 @@ class SearchRequest(db.Model):  # pylint: disable=too-many-instance-attributes
     def search_by_registration_number(self):
         """Execute a search by registration number query."""
         reg_num = self.request_json['criteria']['value']
-        result = db.session.execute(search_utils.REG_NUM_QUERY, {'query_value': reg_num})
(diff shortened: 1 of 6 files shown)


ppr-api/src/ppr_api/models/search_request.py
```diff
@@ -114,7 +114,7 @@ def update_search_selection(self, search_json):
     def search_by_registration_number(self):
         """Execute a search by registration number query."""
         reg_num = self.request_json['criteria']['value']
-        result = db.session.execute(search_utils.REG_NUM_QUERY, {'query_value': reg_num})
+        result = db.session.execute(search_utils.REG_NUM_QUERY, {'query_value': reg_num.strip().upper()})
```
kialj876, line 117: might want to safe guard this in case reg_num is null (unless validation protects this)

kialj876 (approved):
Looks good, just one thing to consider

ppr-api/src/ppr_api/models/search_request.py
doug-lovett: You are right, search criteria null/empty string fails validation so it is already covered.

PR #148, opened 2021-08-09
various registration types changes for step 1

jordiwes:
There are also some 2 & 3 changes as well as fee changes.
This is intended to set up some frameworks for the steps to parse out all of the changes for each step.

*Issue #:* /bcgov/entity#7860

*Description of changes:*

### DIFF
--- a/ppr-ui/src/views/LengthTrust.vue
+++ b/ppr-ui/src/views/LengthTrust.vue
@@ -16,7 +16,7 @@
             </v-row>
             <stepper class="mt-4" />
             <v-row no-gutters class="pt-10">
-              <v-col cols="auto" class="sub-header" v-if="registrationType !== registrationTypes['REPAIRERS_LIEN']">
+              <v-col cols="auto" class="sub-header" v-if="registrationType !== registrationTypeRL">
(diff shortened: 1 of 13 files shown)


ppr-ui/src/views/LengthTrust.vue
```diff
@@ -16,23 +16,16 @@
             </v-row>
             <stepper class="mt-4" />
             <v-row no-gutters class="pt-10">
-              <v-col cols="auto" class="sub-header" v-if="registrationType !== registrationTypeRL">
+              <v-col cols="auto" class="sub-header" v-if="registrationType !== registrationTypes['REPAIRERS_LIEN']">
```
kialj876: can you use the enum here instead of 'REPAIRERS_LIEN'?

kialj876 (approved):
Hey Steph this looks good, just one comment

ppr-ui/src/views/LengthTrust.vue
jordiwes: I'm certainly becoming less good with the options API... I just reverted back to the way Doug had it to keep it simple.

kialj876 (approved):

PR #150, opened 2021-08-11
remaining changes for marriage / mh

jordiwes:
*Issue #:* /bcgov/entity#7860

*Description of changes:*
- remaining changes for marriage / mh

### DIFF
--- a/ppr-ui/src/components/collateral/EditCollateral.vue
+++ b/ppr-ui/src/components/collateral/EditCollateral.vue
@@ -34,7 +34,7 @@
                   filled
                   label="Vehicle Type"
                   v-model="currentVehicle.type"
-                  id="txt-type"
+                  id="txt-type-drop"
                   :error-messages="
                     errors.type.message ? errors.type.message : ''
                   "
--- a/ppr-ui/src/components/collateral/EditCollateral.vue
+++ b/ppr-ui/src/components/collateral/EditCollateral.vue
@@ -176,6 +176,7 @@ import { defineComponent, onMounted } from '@vue/composition-api'
 
 import { useCollateralValidation } from './composables/useCollateralValidation'
 import { useVehicle } from './composables/useVehicle'
+import { APIVehicleTypes } from '@/enums'
 
 export default defineComponent({
   props: {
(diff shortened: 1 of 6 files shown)


ppr-ui/src/components/collateral/EditCollateral.vue
```diff
@@ -205,6 +215,10 @@ export default defineComponent({
 
     onMounted(() => {
       getVehicle()
+      if (mustHaveManufacturedHomeCollateral()) {
+        // set the current vehicle type to motor home
+        currentVehicle.value.type = 'MH'
```
kialj876: enum pls

ppr-ui/tests/unit/EditCollateral.spec.ts
```diff
@@ -71,7 +72,7 @@ describe('Collateral add tests', () => {
   })
 
   it('adds a vehicle to the store', async () => {
-    wrapper.find('#txt-type').setValue('MV')
+    wrapper.find('#txt-type-drop').setValue('MV')
```
kialj876: I know this is existing code, but we should be using the enums here too

ppr-ui/tests/unit/EditCollateral.spec.ts
```diff
@@ -88,6 +89,30 @@ describe('Collateral add tests', () => {
...
+  })
+
+  it('renders with readonly type and manufactured home input', async () => {
+    expect(wrapper.findComponent(EditCollateral).exists()).toBe(true)
+    expect(wrapper.find('#txt-type').exists()).toBe(true)
+    // show read only text box
+    expect(wrapper.find('#txt-man').exists()).toBe(true)
+
+    // no drop down
+    expect(wrapper.find('#txt-type-drop').exists()).toBe(false)
```
kialj876: Don't worry about this for now, but I think it's nice to define these given ids/classes for selectors at the top of the test suite (not for vuetify classes, but for the ones we define ourselves) so that if we change it we only need to do it in one place and I think it's nicer for someone new to the area of code

kialj876 (approved):
Hey Steph just some clean up comments, but looks good

ppr-ui/src/components/collateral/EditCollateral.vue
jordiwes: done

ppr-ui/tests/unit/EditCollateral.spec.ts
jordiwes: gotcha, I did add one for the drop down

PR #152, opened 2021-08-11
8004 drop down update

kialj876:
*Issue #:* /bcgov/entity#8004

*Description of changes:*
see ticket for details


ppr-ui/src/interfaces/store-interfaces/state-interfaces/account-information-interface.ts
```diff
@@ -1,4 +1,4 @@
-// Account information interface
+import { AccountProductSubscriptionIF } from '@/interfaces'
```
jordiwes: why this import?

ppr-ui/tests/unit/RegistrationBarTypeAhead.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    wrapper = createComponent()
+  })
+  afterEach(() => {
+    wrapper.destroy()
+  })
+
+  it('renders registration autocomplete list', async () => {
+    expect(wrapper.findComponent(RegistrationBarTypeAheadList).exists()).toBe(true)
+    expect(wrapper.find(registrationTypeAhead).exists()).toBe(true)
+    // check autocomplete displayed
```
jordiwes, line 85: should you check that autocomplete is displayed here :)

ppr-ui/tests/unit/test-data/mock-registration-responses.ts
```diff
@@ -29,22 +27,22 @@ export const mockedRegistration2: RegistrationSummaryIF = {
...
-    documentId: 'D9000338',
-    baseRegistrationNumber: '',
-    registrationType: APIRegistrationTypes.MISCELLANEOUS_OTHER,
-    registrationDescription: '',
-    path: '/path/to/doc',
-    createDateTime: '2021-08-01T17:21:17+00:00',
-    clientReferenceId: ''
+  documentId: 'D9000338',
+  baseRegistrationNumber: '',
+  registrationType: APIRegistrationTypes.OTHER,
```
jordiwes, line 43: Should we add a description here? I guess this begs a bigger picture question... if the user fills out the dialog with their own registration description, how does this get sent to the API?

ppr-ui/tests/unit/RegistrationBarTypeAhead.spec.ts
kialj876: Oh I will move that comment above (the *registrationTypeAhead* is the autocomplete). I still couldn't access the actual list via jest though :(

ppr-ui/tests/unit/test-data/mock-registration-responses.ts
kialj876: so when they enter the name in the modal it is saved to the store in the new *registrationTypeOtherDesc* . Doug said he would add *otherDescription* as a field in the call that the api is expecting so we can add that in to the post (I didn't touch that since this code is supposed to just be for the dropdown)

kialj876: I didn't add it to the registrationType object because it is a bit of an edge case since it only applies to 1 out of the 20 or so types

ppr-ui/src/interfaces/store-interfaces/state-interfaces/account-information-interface.ts
kialj876: oops, I was going to put this in here and then decided to have it as it's own section in the store. Will remove

jordiwes:
resolve the conflicts then good to go

jordiwes (approved):

PR #153, opened 2021-08-12
7681 secured party code ux

jordiwes:
*Issue #:* /bcgov/entity#7681

*Description of changes:*
- some popup changes based on ux

### DIFF
--- a/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
+++ b/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
@@ -10,13 +10,13 @@
       <v-row no-gutters>
         <v-col cols="11">
           <v-row no-gutters>
-            <v-col class="text-md-center">
-              <v-icon style="font-size: 30px" color="red">mdi-information-outline</v-icon>
+            <v-col class="text-md-center ml-8">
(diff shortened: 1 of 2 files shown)


ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -224,7 +228,7 @@ export default defineComponent({
 <style lang="scss" module>
 @import '@/assets/styles/theme.scss';
 .addressText {
-  font-size: 12px;
+  font-size: 14px;
```
kialj876: I've been using px too in a lot of places, but I think it is better for accessibility if we use the rem conversion for font size (it allows the user to adjust it to their own default if they prefer i.e. enlarge it if they can't see well). Thoughts? I don't have a lot of experience with this, but I think we are doing that in the other UIs

kialj876 (approved):
Looks good, one possible change depending on if you agree or not

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
jordiwes: yes you are probably right, made the change

PR #158, opened 2021-08-17
API minor bug fixes, model change for Bob

doug-lovett:

*Issue #:* /bcgov/entity#8557

*Description of changes:*
- Change db table previous_financing_statements date data types from timestamp to varchar for Bob.
- 8498 correct API renewal expiry date value.
- 8557 correct debtor individual name search with middle name exact matching and report search results display.
- 7858 UX assurance changes to the financing statement pdf.


ppr-api/src/ppr_api/models/previous_financing_statement.py
```diff
@@ -27,14 +26,18 @@ class PreviousFinancingStatement(db.Model):  # pylint: disable=too-many-instance
                              primary_key=True, nullable=False)
     # Free text description
     registration_type = db.Column('registration_type', db.String(30), nullable=False)
-    # cb is companies branch
-    cb_date = db.Column('cb_date', db.DateTime, nullable=True)
+    # From Bob: need to change the data type from date for the 3 columns to varchar(7) as I am not able
```
kialj876, line 29: thank you for the comments

ppr-api/src/ppr_api/models/search_request.py
```diff
@@ -225,15 +225,26 @@ def search_by_business_name(self):
...
         max_results_size = int(current_app.config.get('ACCOUNT_SEARCH_MAX_RESULTS'))
-        result = db.session.execute(search_utils.INDIVIDUAL_NAME_QUERY,
-                                    {'query_last': last_name.strip().upper(),
-                                     'query_first': first_name.strip().upper(),
-                                     'max_results_size': max_results_size})
+        if middle_name is not None and middle_name.strip() != '':
+            result = db.session.execute(search_utils.INDIVIDUAL_NAME_MIDDLE_QUERY,
+                                        {'query_last': last_name.strip().upper(),
+                                         'query_first': first_name.strip().upper(),
+                                         'query_middle': middle_name.strip().upper(),
```
kialj876, line 241: will middle name always be populated? I see that you check the field is there, but is it possible it is None and could cause an error here?

kialj876 (approved):
Looks good to me. Just one comment to consider

ppr-api/src/ppr_api/models/search_request.py
doug-lovett: Thanks Kial, I have added another check in a second commit.

kialj876 (approved):

PR #160, opened 2021-08-18
8397 - related secured party code dropdown

jordiwes:
*Issue #:* /bcgov/entity#8397

*Description of changes:*
- for other types of registration, only allow related secured party code


ppr-ui/src/components/parties/SecuredParties.vue
```diff
@@ -298,30 +430,46 @@ export default defineComponent({
...
 }
 .v-remove:hover {
   background-color: white important;
 }
+
+.party-search {
+  background-color: white;
+}
+.auto-complete-row {
+  font-size: 14px;
```
kialj876: use rem for font pls

ppr-ui/src/components/parties/SecuredParties.vue
```diff
@@ -298,30 +430,46 @@ export default defineComponent({
...
+}
+
+::v-deep .party-search .v-select__selections {
+  color: $gray7 !important;
+}
+
+::v-deep .v-list-item--active {
+  padding-left: 30px;
+  color: $primary-blue !important;
+  font-size: 14px;
```
kialj876: same here

ppr-ui/tests/unit/test-data/mock-new-registration.ts
```diff
@@ -55,6 +55,16 @@ export const mockedMarriageMH: RegistrationTypeIF = {
   text: ''
 }
 
+export const mockedOtherCarbon: RegistrationTypeIF = {
```
kialj876: is there a reason for these mocked registration types other than getting the value we already have defined elsewhere? Could you grab this object from the list of registration types in resources instead? That way we don't have more than one place to change it if we need to alter it later

kialj876 (commented):
Looks good, just a couple small things

ppr-ui/tests/unit/test-data/mock-new-registration.ts
jordiwes: We're using this in tests to set the store to test the other registration type changes... it needs to be a RegistrationTypeIF to set the store... if that makes sense

kialj876:
hey steph you can probably fix that build error with the npm run lint cmd ?

doug-lovett (approved):

ppr-ui/src/components/parties/SecuredParties.vue
jordiwes: done

ppr-ui/src/components/parties/SecuredParties.vue
jordiwes: done

ppr-ui/tests/unit/test-data/mock-new-registration.ts
jordiwes: done

kialj876 (approved):
thanks for making those changes!

PR #179, opened 2021-08-25
style adjustments

jordiwes:
*Issue #:* /bcgov/entity#7681

*Description of changes:*
- a few style adjustments as per ux

### DIFF
--- a/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
+++ b/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
@@ -11,7 +11,7 @@
         <v-col cols="11">
           <v-row no-gutters>
             <v-col class="text-md-center ml-8">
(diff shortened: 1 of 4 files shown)


ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -291,7 +291,7 @@ export default defineComponent({
 }
 
 .iconRed {
-  color: error !important;
+  color: #D3272C !important;
```
kialj876: Is the theme not coming through? 'error' is set to be #D3272C for our app, but I guess since it's set in the style it doesn't get it. Would be good to add a $red or $error colour equal to this since we use it in a few places

kialj876 (approved):
Looks good! One comment for now or later or not (it doesn't matter that much haha)

doug-lovett (approved):

PR #182, opened 2021-08-26
8397 table UI again

jordiwes:
*Issue #:* /bcgov/entity#8397

*Description of changes:*
- more updates for the table ui

### DIFF
--- a/ppr-ui/src/assets/styles/overrides.scss
+++ b/ppr-ui/src/assets/styles/overrides.scss
@@ -253,7 +253,7 @@ tbody {
 }
 
 // list items
-.theme--light.v-list-item:not(.v-list-item--active):not(.v-list-item--disabled) .registration-list-item {
+.theme--light.v-list-item .registration-list-item {
   font-size: 0.875rem;
   padding-left: 16px;
 }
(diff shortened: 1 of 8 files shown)


doug-lovett (approved):
I just scanned RegistrationTable.vue as there are many changes. Looks good.

ppr-ui/src/assets/styles/overrides.scss
```diff
@@ -251,7 +253,7 @@ tbody {
 }
 
 // list items
-.theme--light.v-list-item:not(.v-list-item--active):not(.v-list-item--disabled) .registration-list-item {
```
kialj876, line 254: ~hey steph, I need this last bit on there for the dropdown (the headers are disabled so it will give them this styling otherwise)~

kialj876: ~'.registration-list-item' is a class that I put in for the dropdown registration list, I don't think it's used anywhere else so it shouldn't effect the table~

kialj876: ah okay I see you are using it in the table, and since it has the '.registration-list-item' class it should be fine actually. Do you mind checking it against the dev dropdown before merging?

kialj876 (changes requested):
Hey steph this looks very good, just one thing to verify if you have time. Also can you add a test for the autocomplete list showing/working in the table?

ppr-ui/src/assets/styles/overrides.scss
jordiwes: looks fine

jordiwes:
I added a test for the registration type ahead

kialj876 (approved):

PR #184, opened 2021-08-27
ux updates

kialj876:

*Issue #:* /bcgov/entity#8454 /bcgov/entity#8004

*Description of changes:*
- small ux tweaks

### DIFF
--- a/ppr-ui/src/assets/styles/overrides.scss
+++ b/ppr-ui/src/assets/styles/overrides.scss
@@ -256,18 +256,10 @@ tbody {
   padding-left: 16px;
 }
 
-.theme--light.v-list-item:not(.v-list-item--active):not(.v-list-item--disabled) {
-  color: $gray7 !important;
-  min-height: 0;
-}
-
+// needed for registration bar type ahead list, but can be added to scoped styling if needed
 .theme--light.v-list-item:not(.v-list-item--active):not(.v-list-item--disabled):hover {
   background-color: $gray1;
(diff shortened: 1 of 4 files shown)


ppr-ui/src/assets/styles/overrides.scss
```diff
@@ -256,20 +256,12 @@ tbody {
   padding-left: 16px;
 }
 
-.theme--light.v-list-item:not(.v-list-item--active):not(.v-list-item--disabled) {
-  color: $gray7 !important;
-  min-height: 0;
-}
-
+// needed for registration bar type ahead list, but can be added to scoped styling if needed
```
jordiwes, line 259: We are overriding each other here... I just made a change to this for the table drop down... we should probably add some sub selectors or something. Disadvantage of the module css

kialj876: wait so you want the styling I deleted ? I removed that to prevent the issues with the registration table dropdown I introduced before haha

kialj876: If you want me to add that back in it is fine with me

doug-lovett (approved):

ppr-ui/src/assets/styles/overrides.scss
jordiwes: actually keep it in as I'm all confused, and I'll rebase my stuff and check it out

jordiwes (approved):

PR #188, opened 2021-08-30
secure party change modal and ux items

jordiwes:
*Issue #:* /bcgov/entity#7864

*Description of changes:*
- some ux updates for other registrations
- secured party change modal

### DIFF
--- a/ppr-ui/src/components/parties/SecuredParties.vue
+++ b/ppr-ui/src/components/parties/SecuredParties.vue
@@ -382,6 +382,7 @@ export default defineComponent({
       localState.loading = true
       if (parties.registeringParty) {
         // go to the service and see if there are similar secured parties
+        parties.registeringParty.businessName = 'Bank of Montreal'
         const response: [SearchPartyIF] = await partyCodeSearch(
           parties.registeringParty.businessName,
           true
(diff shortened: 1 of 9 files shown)


ppr-ui/src/components/parties/SecuredParties.vue
```diff
@@ -365,6 +381,7 @@ export default defineComponent({
     const fetchOtherSecuredParties = async () => {
       localState.loading = true
       if (parties.registeringParty) {
+        parties.registeringParty.businessName = 'Bank of Montreal'
```
doug-lovett: Should this be removed?

jordiwes: whoops!

doug-lovett (approved):

doug-lovett (approved):
Adding OT other type description looks good.

kialj876 (changes requested):
Looks good to me, but can you add in some tests please? New dialogue should have a test suite and we should probably test it coming up wherever its needed (should be pretty similar to existing tests we have for the confirmation dialog).
I also think you could've used the confirmation dialog for this one (would've needed an 'if' around the checkbox is the only change right?), but it doesn't really matter. Would've been less test setup for ya though!

jordiwes:
I added some tests :)

kialj876 (approved):
Thanks for adding those in!

PR #191, opened 2021-08-30
caution box component added

kialj876:

*Issue #:* /bcgov/entity#8455

*Description of changes:*
- added tiny caution box component

### DIFF
--- a/ppr-ui/src/components/common/CautionBox.vue
+++ b/ppr-ui/src/components/common/CautionBox.vue
@@ -1,0 +1,41 @@
+<template>
+  <div :class="$style['caution-box']" class="caution-box">
+    <p class="ma-0 pa-4">
+      <b>Important:</b> {{ msg }}
+    </p>
+  </div>
+</template>
+
+<script lang="ts">
+// external
+import {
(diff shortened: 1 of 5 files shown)


doug-lovett (approved):

ppr-ui/src/components/common/CautionBox.vue
```diff
@@ -0,0 +1,41 @@
...
+  toRefs
+} from '@vue/composition-api'
+
+export default defineComponent({
+  name: 'CautionBox',
+  props: {
+    setMsg: String
+  },
+  setup (props) {
+    const localState = reactive({
```
jordiwes, line 23: quick question, could the message change, ie. is the component reused during the course of the app? If so you will need a watch on this...

kialj876: ha yes I thought about this because of our earlier convo. It will always be static, but it may get initialized with a different message by the parent component

jordiwes: 10-4

jordiwes (approved):

PR #192, opened 2021-08-31
7858 fix 2 RL issues from QA.

doug-lovett:

*Issue #:* /bcgov/entity#7858

*Description of changes:*
- See the ticket: allow $ sign with RL lien amount; correct display of surrender date.
- 8080:  Resolve PPR logout not working: common component SBCHeader redirectOnLogout property is no longer working.

### DIFF
--- a/ppr-ui/src/utils/date-helper.ts
+++ b/ppr-ui/src/utils/date-helper.ts
@@ -23,7 +23,7 @@ export function convertDate (date: Date, includeTime: boolean, includeTz: boolea
 
   // add 'Pacific Time' to end if pacific timezone
   let timezone = ''
-  if ((date.toString()).includes('Pacific')) timezone = 'Pacific Time'
+  if ((date.toString()).includes('Pacific')) timezone = 'Pacific time'
 
   const datetime = format12HourTime(date)
 
--- a/ppr-ui/src/utils/date-helper.ts
+++ b/ppr-ui/src/utils/date-helper.ts
@@ -18,11 +18,12 @@ export function format12HourTime (date: Date): string {
 
 /** Converts date to display format. */
 export function convertDate (date: Date, includeTime: boolean, includeTz: boolean): string {
-  if (!includeTime) return moment(date).format('MMMM D, Y')
+  // For just dates use the UTC timestamp otherwise the day of month is incorrect (1 day earlier).
+  if (!includeTime) return moment(date).utc().format('MMMM D, Y')
 
   // add 'Pacific Time' to end if pacific timezone
   let timezone = ''
-  if ((date.toString()).includes('Pacific')) timezone = 'Pacific time'
+  if ((date.toString()).includes('Pacific')) timezone = 'Pacific Time'
 
   const datetime = format12HourTime(date)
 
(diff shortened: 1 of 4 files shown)


ppr-ui/src/utils/date-helper.ts
```diff
@@ -18,11 +18,12 @@ export function format12HourTime (date: Date): string {
...
 /** Converts date to display format. */
 export function convertDate (date: Date, includeTime: boolean, includeTz: boolean): string {
-  if (!includeTime) return moment(date).format('MMMM D, Y')
+  // For just dates use the UTC timestamp otherwise the day of month is incorrect (1 day earlier).
+  if (!includeTime) return moment(date).utc().format('MMMM D, Y')
 
   // add 'Pacific Time' to end if pacific timezone
   let timezone = ''
-  if ((date.toString()).includes('Pacific')) timezone = 'Pacific time'
+  if ((date.toString()).includes('Pacific')) timezone = 'Pacific Time'
```
kialj876: Change this back pls. Lower case 't' is correct (I did the same thing before too and was corrected)

ppr-ui/src/utils/date-helper.ts
```diff
@@ -18,11 +18,12 @@ export function format12HourTime (date: Date): string {
 
 /** Converts date to display format. */
 export function convertDate (date: Date, includeTime: boolean, includeTz: boolean): string {
-  if (!includeTime) return moment(date).format('MMMM D, Y')
+  // For just dates use the UTC timestamp otherwise the day of month is incorrect (1 day earlier).
+  if (!includeTime) return moment(date).utc().format('MMMM D, Y')
```
kialj876: Hey Doug, do you mean the date was getting converted incorrectly to pacific time or it was actually converting the utc date directly before? Just verifying: will this return the date in UTC or Pacific haha? We should always be showing pacific time in the UI by default I think

kialj876 (changes requested):
Hey Doug, this looks good. I just have one small change request and a question

ppr-ui/src/utils/date-helper.ts
doug-lovett: Sorry, restored to Pacific time

ppr-ui/src/utils/date-helper.ts
doug-lovett: Moment incorrectly formats the day of month without this change. 2021-08-30 was formatted as August 29, 2021.

kialj876: Right, but is that because it is 2021-08-30T00:00... in the db in UTC? Wouldn't that mean it is actually correct to show one day behind in pacific time?

doug-lovett: Right, thanks. I have restored convertDate and added time to the format call to force the correct local day.

kialj876 (approved):

ppr-ui/src/utils/date-helper.ts
kialj876: awesome 👍

PR #195, opened 2021-08-31
secured party code - ui items based on latest feedback

jordiwes:
*Issue #:* /bcgov/entity#7681

*Description of changes:*
- ever evolving secured party code lookup

### DIFF
--- a/ppr-ui/src/components/parties/PartySearch.vue
+++ b/ppr-ui/src/components/parties/PartySearch.vue
@@ -6,9 +6,9 @@
           filled
           label="Secured Party Code or Name"
           id="txt-code"
-          :hide-details="true"
           v-model="searchValue"
           persistent-hint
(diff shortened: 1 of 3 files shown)


ppr-ui/src/components/parties/PartySearch.vue
```diff
@@ -85,7 +85,7 @@ export default defineComponent({
   setup (props, context) {
     const localState = reactive({
       searchValue: '',
-      autoCompleteResults: [],
+      autoCompleteResults: null,
```
kialj876, line 88: will this cause an error when initialized since you check *autoCompleteResults.length* inside party-autocomplete?

jordiwes: I popped in the null check

kialj876 (approved):

PR #197, opened 2021-09-01
save registration number to store

jordiwes:
*Issue #:* /bcgov/entity#8452

*Description of changes:*


doug-lovett (approved):

kialj876:
Hey Steph, I don't think we need the registration number saved here do we? We need the debtor name that they enter saved though

PR #207, opened 2021-09-02
8535 discharge folio

jordiwes:
*Issue #:* /bcgov/entity#8535

*Description of changes:*
- add folio to discharge page with validity check

### DIFF
--- a/ppr-ui/src/components/common/FolioNumberSummary.vue
+++ b/ppr-ui/src/components/common/FolioNumberSummary.vue
@@ -14,7 +14,7 @@
       </v-row>
 
       <v-row class="px-1">
-        <v-col cols="12" class="pa-0">
+        <v-col cols="12" class="pa-0" :class="showErrors && !valid ? 'border-error-left': ''">
           <v-card flat>
             <v-row no-gutters>
               <v-col cols="3" class="generic-label pt-10 px-8"
@@ -49,6 +49,11 @@ import {
 import { useGetters, useActions } from 'vuex-composition-helpers'
 
 export default defineComponent({
+  props: {
+    setShowErrors: {
+      default: false
+    }
+  },
   setup (props, { emit }) {
     const { setFolioOrReferenceNumber } = useActions<any>([
       'setFolioOrReferenceNumber'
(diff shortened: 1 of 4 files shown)


ppr-ui/tests/unit/ConfirmDischarge.spec.ts
```diff
@@ -91,6 +92,8 @@ describe('ConfirmDischarge registration view', () => {
     expect(wrapper.findComponent(RegistrationFee).vm.registrationType).toBe('Total Discharge')
     // buttons
     expect(wrapper.findComponent(ButtonsStacked).exists()).toBe(true)
+    // folio
+    expect(wrapper.findComponent(FolioNumberSummary).exists()).toBe(true)
```
kialj876, line 96: can you add a test in here for when folio is invalid it blocks the submit?

kialj876 (commented):
Hey Steph, this looks good just one comment. I gotta say though I am not a big fan of most of the formatting changes. Whatever you have running to auto fix passes our linter, but it does make it different from what we auto fix across the other apps. The main thing is when it makes unnecessary line length changes, the unneeded brackets and what it did with line 28/29/30 in ConfirmDischarge is annoying to look at for me, is it possible to adjust it for those things?

ppr-ui/src/components/common/FolioNumberSummary.vue
```diff
@@ -1,36 +1,40 @@
 <template>
   <v-container flat class="pa-0" id="folio-summary">
```
kialj876, line 2: I also think you'll need to add the red border class somewhere in here when needed (similar to the confirm section)

ppr-ui/tests/unit/ConfirmDischarge.spec.ts
jordiwes: submit doesn't do anything right now.... so it's hard to tell ;)
I did put in a test for the show error property

ppr-ui/src/components/common/FolioNumberSummary.vue
jordiwes: added

jordiwes:
I fixed/reverted some of the changes that the linter made

ppr-ui/tests/unit/ConfirmDischarge.spec.ts
kialj876: ah but it does

kialj876 (approved):
thanks for updating!

doug-lovett (approved):

PR #210, opened 2021-09-03
8806 API update query by party code logic.

doug-lovett:

*Issue #:* /bcgov/entity#8806

*Description of changes:*
Query on branch id instead of head office id. Match on string values that start with the query parameter.

### DIFF
--- a/ppr-api/src/ppr_api/models/client_code.py
+++ b/ppr-api/src/ppr_api/models/client_code.py
@@ -83,7 +83,7 @@ class ClientCode(db.Model):  # pylint: disable=too-many-instance-attributes
         return party
 
     @classmethod
-    def find_by_head_office_code(cls, head_office_id: str = None):
+    def find_by_head_office_code(cls, head_office_id: str):
         """Return a list of client parties belonging to a head office searching by code."""
         party_codes = []
         if head_office_id and len(head_office_id) <= 4 and head_office_id.strip().isdigit():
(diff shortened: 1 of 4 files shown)


ppr-api/src/ppr_api/models/client_code.py
```diff
@@ -83,7 +83,7 @@ def find_by_code(cls, code: str = None):
         return party
 
     @classmethod
-    def find_by_head_office_code(cls, head_office_id: str = None):
+    def find_by_head_office_code(cls, head_office_id: str):
```
jordiwes, line 86: do we still need this? keep just in case?

ppr-api/src/ppr_api/models/client_code.py
```diff
@@ -95,6 +95,41 @@ def find_by_head_office_code(cls, head_office_id: str = None):
 
         return party_codes
 
+    @classmethod
+    def find_by_account_number(cls, account_number: int):
```
jordiwes, line 99: where is this used?

ppr-api/src/ppr_api/models/client_code.py
```diff
@@ -95,6 +95,41 @@ def find_by_head_office_code(cls, head_office_id: str = None):
...
+    def find_by_branch_start(cls, branch_code: str):
+        """Return a list of client parties matching on branch ids that start with a query number."""
+        party_codes = []
+        if not branch_code or not branch_code.strip().isdigit():
+            return party_codes
+        query_num = branch_code.strip()
+        if int(query_num) < 100:  # Exact match
+            party = cls.find_by_code(query_num)
+            if party:
+                party_codes.append(party)
```
jordiwes, line 121: do you need to append party.json here?

ppr-api/src/ppr_api/models/client_code.py
doug-lovett: No harm in keeping.

ppr-api/src/ppr_api/models/client_code.py
doug-lovett: It's not...yet. It will be in a coming ticket split from this one.

ppr-api/src/ppr_api/models/client_code.py
doug-lovett: Yes, this condition is an exact match returning either a single entry or an empty array.

jordiwes (approved):

kialj876 (approved):
Looks good to me

PR #215, opened 2021-09-07
8451 ux for drop down

jordiwes:
*Issue #:* /bcgov/entity#8451

*Description of changes:*
- ux for drop down


kialj876 (changes requested):
Hey Steph, this looks good, but do you know why the package-lock is so huge and have so much changed in there? When I add the mdi/font upgrade and reinstall it only adds a couple changes ?

-- just want to make sure this is okay before we merge it in

kialj876 (approved):

PR #216, opened 2021-09-07
8818 8727 api correct expiry timestamps

doug-lovett:
*Issue #:* /bcgov/entity#8818

*Description of changes:*
- Store expiry timestamps correctly as UTC regardless of local time of day.
- Display expiry timestamps correctly in reports as local date and time, allowing for daylight savings.
- Update unit tests.
- For renewals, add to existing expiry for all applicable registration types including repairer's lien.


jordiwes (approved):
I had a look and looks good to me... but may want some more expertise

PR #219, opened 2021-09-09
8557 display middle name in search query results.

doug-lovett:

*Issue #:* /bcgov/entity#8557

*Description of changes:*
Bug fix to display the debtor middle name in the individual debtor search results.
Background:
There is an inconsistency in the representation of a middle name in the API. Individual Debtor search criteria uses second name; everywhere else in the API uses middle name. The Search History table displays criteria values ("second" name); the search query results uses "middle" name.

### DIFF
--- a/ppr-ui/src/components/tables/SearchedResults.vue
+++ b/ppr-ui/src/components/tables/SearchedResults.vue
@@ -70,7 +70,10 @@
             {{ item.vehicleCollateral.make }} {{ item.vehicleCollateral.model }}
           </template>
           <template v-slot:[`item.debtor.personName`]="{ item }">
-            {{ item.debtor.personName.first }} {{ item.debtor.personName.second }} {{ item.debtor.personName.last }}
+            {{ item.debtor.personName.first }}
+            {{ item.debtor.personName.second }}
+            {{ item.debtor.personName.middle }}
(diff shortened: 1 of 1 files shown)


ppr-ui/src/components/tables/SearchedResults.vue
```diff
@@ -70,7 +70,10 @@
             {{ item.vehicleCollateral.make }} {{ item.vehicleCollateral.model }}
           </template>
           <template v-slot:[`item.debtor.personName`]="{ item }">
-            {{ item.debtor.personName.first }} {{ item.debtor.personName.second }} {{ item.debtor.personName.last }}
+            {{ item.debtor.personName.first }}
+            {{ item.debtor.personName.second }}
```
kialj876, line 74: is personName.second a real value? Seems odd that we have 'second' and 'middle' ?

kialj876 (approved):

ppr-ui/src/components/tables/SearchedResults.vue
doug-lovett: The Search History table is populating second name. The Search Results table uses middle name. There is an inconsistency in the API with the search criteria using second name. I updated the description for this issue.

PR #236, opened 2021-09-15
ux assurance for renewals

jordiwes:
*Issue #:* /bcgov/entity#8724

*Description of changes:*
- changes based on ux assurance

### DIFF
--- a/ppr-ui/src/components/registration/RegistrationLengthTrust.vue
+++ b/ppr-ui/src/components/registration/RegistrationLengthTrust.vue
@@ -1,7 +1,12 @@
 <template>
   <v-container fluid no-gutters class="white pa-0" v-if="summaryView">
     <v-card flat id="length-trust-summary">
-      <v-row no-gutters class="summary-header pa-2">
+      <v-row class="py-2" v-if="isRenewal" style="background-color: #f1f3f5">
+        <v-col cols="auto">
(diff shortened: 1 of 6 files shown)


doug-lovett (approved):
looks good

ppr-ui/src/components/registration/RegistrationLengthTrust.vue
```diff
@@ -1,7 +1,12 @@
 <template>
   <v-container fluid no-gutters class="white pa-0" v-if="summaryView">
     <v-card flat id="length-trust-summary">
-      <v-row no-gutters class="summary-header pa-2">
+      <v-row class="py-2 renewal-title" v-if="isRenewal">
```
kialj876: do you need no-gutters here? Also, are you getting anything from nesting the `<h2>` within the `v-row / v-col`? Might be able to just have the `<h2>` with the `v-if` and the `class`

kialj876 (approved):
Looks good! One comment to consider, but it might not be needed. Also needs the test fixed!

ppr-ui/src/components/registration/RegistrationLengthTrust.vue
jordiwes: changed it to just the h2

PR #242, opened 2021-09-17
8980 general collateral status flag changes

doug-lovett:

*Issue #:* /bcgov/entity#8980

*Description of changes:*
- Update search result, financing statement, and amendment/change statement report templates.
- For financing statement consolidated/current view return general collateral.description in financing statement and legacy general collateral. For change and amendment registration collateral return one or both of descriptionAdd, descriptionDelete.
- 8981 include gc added timestamp as addedDateTime in API gc response data.
- Unit test updates
- Fix for report large draft text being partially obscured by the header when running over multiple pages.  

### DIFF
--- a/ppr-api/report-templates/amendmentStatement.html
+++ b/ppr-api/report-templates/amendmentStatement.html
@@ -38,7 +38,7 @@
       {% if environment != '' %}TEST DATA |  {% endif %}Amendment Registration #{{ amendmentRegistrationNumber }}:
     </div>
 
-    <div class="business-details-container">  
+    <div class="registration-details-container">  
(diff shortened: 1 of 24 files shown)


jordiwes (approved):
LGTM!

kialj876 (commented):
Looks good to me, but we may need to include a deleted description within each one -- will ask for clarification at standup

doug-lovett:
Kial, the amendment registration has a mandatory description property. My understanding is that this property will capture the reason for the change including the deletion of general collateral.

kialj876 (approved):

kialj876 (approved):

PR #243, opened 2021-09-17
folio - remove rules for characters

jordiwes:
*Issue #:* /bcgov/entity#8729

*Description of changes:*

### DIFF
--- a/ppr-ui/tests/unit/FolioNumberSummary.spec.ts
+++ b/ppr-ui/tests/unit/FolioNumberSummary.spec.ts
@@ -56,7 +56,11 @@ describe('Folio number on the summarty page', () => {
 
   it('sets the validity to false for > 15 characters', async () => {
     wrapper.find('#txt-folio').setValue('MY TEST THAT IS VERY LONG IN FACT TOO LONG')
-    expect(wrapper.emitted().folioValid).toBeTruthy()
+    await Vue.nextTick()
+    await Vue.nextTick()
+    await Vue.nextTick()
     expect(wrapper.vm.$data.isValid).toBeFalsy()
+    expect(wrapper.emitted().folioValid).toBeTruthy()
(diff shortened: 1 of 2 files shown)


ppr-ui/tests/unit/FolioNumberSummary.spec.ts
```diff
@@ -56,7 +56,11 @@ describe('Folio number on the summarty page', () => {
 
   it('sets the validity to false for > 15 characters', async () => {
     wrapper.find('#txt-folio').setValue('MY TEST THAT IS VERY LONG IN FACT TOO LONG')
-    expect(wrapper.emitted().folioValid).toBeTruthy()
+    await Vue.nextTick()
+    await Vue.nextTick()
+    await Vue.nextTick()
     expect(wrapper.vm.$data.isValid).toBeFalsy()
+    expect(wrapper.emitted().folioValid).toBeTruthy()
```
kialj876, line 63: shouldn't this emit false? Try await flushPromises and use the getLastEvent fn from utils (it will get the last emitted event -- this will have emitted true before setting the value and then false after I think)

doug-lovett (approved):
No comment in addition to Kial's.

ppr-ui/tests/unit/FolioNumberSummary.spec.ts
jordiwes: So, the fact that it emitted an event called folio valid is true. does that make sense? It's not the value of the emit, but that fact that an event was emitted

kialj876: oh that's right! Sorry, this is good then. Probably better to use that util fn in the future though since here we don't know if its emitting the right value or not

kialj876 (approved):

PR #248, opened 2021-09-20
8883 - keep track of registration flow type and party state... part 1

jordiwes:
*Issue #:* /bcgov/entity#8883

*Description of changes:*
added "action type" to party state, and registration flow type to registration state
next step will be implementing the action for debtors (part 2)


ppr-ui/src/store/getters/state-getters.ts
```diff
@@ -88,6 +88,11 @@ export const getRegistrationType = (state: StateIF): RegistrationTypeIF => {
   return state.stateModel.registration.registrationType
 }
 
+/** The selected registration flow type object. */
+export const getRegistrationFlowType = (state: StateIF): string => {
```
kialj876: this should be type: RegistrationFlowType

ppr-ui/src/store/state/state-model.ts
```diff
@@ -39,6 +40,7 @@ export const stateModel: StateModelIF = {
     },
     registrationNumber: '',
     registrationType: null,
+    registrationFlowType: RegistrationFlowType.NEW,
```
kialj876, line 43: can you set it to this inside *mutateNewRegistration* as well? Just in case they do a discharge or a renewal and then do a new registration afterwards

kialj876 (commented):
Hey Steph, looks good I just have two small things

ppr-ui/src/store/getters/state-getters.ts
jordiwes: done

ppr-ui/src/store/state/state-model.ts
jordiwes: done

doug-lovett (approved):
Looks good, no comments in addition to Kial's.

kialj876 (approved):

PR #254, opened 2021-09-23
8979 general collateral

kialj876:
*Issue #:* /bcgov/entity#8979

*Description of changes:*
- see ticket for details


doug-lovett (approved):
Scanned, refactoring looks good.

ppr-ui/src/components/collateral/Collateral.vue
```diff
@@ -95,376 +54,167 @@
...
-                  :error-messages="generalCollateralError"
-                >
-                </v-textarea>
-              </v-col>
-            </v-row>
-          </v-container>
-        </v-card>
-      </v-col>
-    </v-row>
+    <vehicle-collateral :isSummary="false" :showInvalid="showInvalid && !valid" />
```
jordiwes, line 57: Do we only want to show this if hasVehicleCollateral() ?

ppr-ui/src/components/collateral/generalCollateral/GenColSummary.vue
```diff
@@ -0,0 +1,118 @@
...
+
+    const localState = reactive({
+      showingHistory: false,
+      baseGenCollateralIndex: computed(() => {
+        return (localState.generalCollateral?.length || 0) - 1
+      }),
+      generalCollateral: computed((): GeneralCollateralIF[] => {
+        return getGeneralCollateral.value as GeneralCollateralIF[] || []
+      }),
+      registrationFlowType: computed((): RegistrationFlowType => {
```
jordiwes, line 80: nit: but I don't think we need this computed in the localState...
just const registrationFlowType = getRegistrationFlowType.value 
and then return it below

ppr-ui/src/components/collateral/generalCollateral/GeneralCollateral.vue
```diff
@@ -0,0 +1,64 @@
...
+  components: {
+    GenColEdit,
+    GenColSummary
+  },
+  props: {
+    isSummary: {
+      type: Boolean,
+      default: false
+    },
+    setRegistrationType: String as () => APIRegistrationTypes,
```
jordiwes, line 36: Is this used?

ppr-ui/src/interfaces/store-interfaces/state-model-interface.ts
```diff
@@ -15,10 +15,8 @@ export interface StateModelIF {
   folioOrReferenceNumber: string
   registration: {
     collateral: AddCollateralIF
-    confirmDebtorName: DebtorNameIF // Required for registrations.
+    confirmDebtorName: DebtorNameIF // Required for actions on existing registrations.
     creationDate: string
-    currentStep: number,
```
jordiwes, line 20: No impact to removing this?

kialj876: nope

kialj876: it isn't used anywhere

ppr-ui/src/components/collateral/generalCollateral/GeneralCollateral.vue
kialj876: yep

kialj876: it needs to get passed to the child. I think ideally none of the children would need to access the store at all, but that would've been too complicated to alter this time

ppr-ui/src/components/collateral/generalCollateral/GenColSummary.vue
kialj876: for now that would be okay since we only use the summary in a situation where the flow is static, but doing it that way would prevent it from updating if it was ever in a situation where it changed after it first rendered

ppr-ui/src/components/collateral/Collateral.vue
kialj876: That's handled within the vehicle collateral component

ppr-ui/src/components/collateral/generalCollateral/GenColSummary.vue
kialj876: Since this component will be re-used in the future I think it's better to keep it the way it is unless there's a reason this is worse?

jordiwes (approved):

PR #263, opened 2021-09-28
Partially complete changes to save/load draft amendment.

doug-lovett:

*Issue #:* /bcgov/entity#9102

*Description of changes:*
- API setup to save draft.
- Store changes
- More unit tests to come.

### DIFF
--- a/ppr-ui/src/utils/registration-helper.ts
+++ b/ppr-ui/src/utils/registration-helper.ts
@@ -24,11 +24,25 @@ import {
 } from '@/utils'
 import { RegistrationTypes } from '@/resources'
 
+/** Set the amendment add/delete lists depending on the registration list actions */
+function setAmendmentList (baseList:Array<any>, addList:Array<any>, deleteList:Array<any>) {
+  for (let i = 0; i < baseList.length; i++) {
+    if (baseList[i].action) {
+      if (baseList[i].action === ActionTypes.ADDED || baseList[i].action === ActionTypes.EDITED) {
+        addList.push(JSON.parse(JSON.stringify(baseList[i])))
+      }
+      if (baseList[i].action === ActionTypes.REMOVED || baseList[i].action === ActionTypes.EDITED) {
+        deleteList.push(JSON.parse(JSON.stringify(baseList[i])))
+      }
+    }
+  }
+}
+
 /** Setup the amendment registration for the API call. All data to be saved is in the store state model. */
 function setupAmendmentStatement (stateModel:StateModelIF): AmendmentStatementIF {
   const registrationType: RegistrationTypeIF = stateModel.registration.registrationType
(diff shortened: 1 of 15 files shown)


ppr-ui/src/utils/registration-helper.ts
```diff
@@ -22,6 +24,147 @@ import {
...
+    apiCall = 'update'
+    draftResponse = await updateDraft(draft)
+  } else {
+    apiCall = 'create'
+    draftResponse = await createDraft(draft)
+  }
+  if (draftResponse !== undefined && draftResponse.error === undefined) {
+    console.log('saveAmendmentStatementDraft ' + apiCall + ' draft successful for documentId ' +
+                draftResponse.amendmentStatement.documentId)
+  } else if (draftResponse !== undefined) {
```
jordiwes: nit: I find this a little confusing, and would be tempted to check draftResponse.error instead... and is there a case where draftResponse would be blank? exception?

jordiwes (approved):
One nit, otherwise good!

ppr-ui/src/utils/registration-helper.ts
```diff
@@ -22,6 +24,147 @@ import {
...
+          parties.securedParties[i].action === ActionTypes.EDITED) {
+        statement.deleteSecuredParties.push(JSON.parse(JSON.stringify(parties.securedParties[i])))
+      }
+    }
+  }
+
+  const collateral:AddCollateralIF = stateModel.registration.collateral
+  // Conditionally set draft vehicle collateral
+  if (collateral.vehicleCollateral && collateral.vehicleCollateral.length > 0) {
+    for (let i = 0; i < collateral.vehicleCollateral.length; i++) {
```
kialj876: hey doug, do you think you could make this for loop into it's own fn and use it for parties/debtors as well? Maybe pass in the the list your iterating, the statement.add... list and the statement.delete.. list. I can see someone coming in here to edit something later in ops and messing it up with a copy/paste error haha

ppr-ui/src/utils/registration-helper.ts
kialj876: if you change this to ```if (!draftResponse) ...``` it will catch it if it is undefined, null, empty etc. Would that be better here and just above too?

kialj876 (commented):
Hey Doug, looks good just a couple comments to consider

ppr-ui/src/utils/registration-helper.ts
doug-lovett: Updated in commit, allowing for case where draftResponse is null.

ppr-ui/src/utils/registration-helper.ts
doug-lovett: Good idea, added function in second commit.

kialj876 (approved):
Awesome, thanks for the changes 👍

PR #270, opened 2021-09-30
8886 plug in secured party validation

jordiwes:
*Issue #:* /bcgov/entity#8886

*Description of changes:*
- secured party validation plugged into page
- tests

### DIFF
--- a/ppr-ui/src/components/parties/SecuredParties.vue
+++ b/ppr-ui/src/components/parties/SecuredParties.vue
@@ -165,7 +165,19 @@
                     class="actions__more-actions"
                     :disabled="addEditInProgress"
                   >
+                  <v-list-item
+                    v-if="registrationFlowType === RegistrationFlowType.AMENDMENT
+                      && row.item.action === ActionTypes.REMOVED"
+                      class="v-remove"
+                      @click="undo(row.index)"
+                    >
+                      <v-list-item-subtitle>
+                        <v-icon small>mdi-undo</v-icon>
+                        <span class="ml-1">Undo</span>
+                      </v-list-item-subtitle>
+                    </v-list-item>
                     <v-list-item
+                      v-else
                       class="v-remove"
                       @click="removeParty(row.index)"
                     >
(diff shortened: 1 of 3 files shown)


ppr-ui/tests/unit/SecuredParties.spec.ts
```diff
@@ -201,4 +251,34 @@ describe('Secured party amendment tests', () => {
...
+    await Vue.nextTick()
+    expect(wrapper.vm.getSecuredPartyValidity()).toBe(true)
+    // remove said secured party
+    // click the drop down arrow
+    wrapper.find('.v-data-table .party-row .actions__more-actions__btn').trigger('click')
+    await Vue.nextTick()
+    //click remove
+    wrapper.find('.actions__more-actions .v-list-item__subtitle').trigger('click')
+    await Vue.nextTick()
+    expect(wrapper.vm.getSecuredPartyValidity()).toBe(false)
```
kialj876, line 281: can you please update this to include ```expect(wrapper.findAll('.invalid-message').length).toBe(0)``` before you click the remove btn and then ```expect(wrapper.findAll('.invalid-message').length).toBe(1)``` after? It should fail right now due to my comment below, but it is the desired functionality I think

kialj876 (changes requested):

ppr-ui/src/components/parties/SecuredParties.vue
```diff
@@ -108,7 +108,7 @@
       <v-col>
         <v-data-table
           class="party-table"
-          :class="{ 'invalid-message': showErrorSecuredParties }"
+          :class="{ 'invalid-message': showErrorSecuredParties && !getSecuredPartyValidity() }"
```
kialj876, line 111: hey steph, I think this will only check ```!getSecuredPartyValidity()``` when it's first rendered since it's a fn. So it will calculate it when the page loads, but if the validity changes afterwards this won't update unless you change it to a computed or reactive value

kialj876 (commented):
ah sorry I didn't press the add to review btn for this comment earlier

ppr-ui/tests/unit/SecuredParties.spec.ts
jordiwes: added tests

kialj876 (approved):

doug-lovett (approved):

PR #277, opened 2021-10-05
Discharge verification ui design changes.

doug-lovett:

*Issue #:* /bcgov/entity#9083

*Description of changes:*
- General verification report content change to include current/consolidated view information at the time of the registration.
- Update verification general collateral content to include all changes at the time of the registration.
- Discharge verification updates from UI design
- Fix API response for the current view general collateral when adding/deleting in the same registration: include deleted collateral. 

### DIFF
--- a/ppr-api/src/ppr_api/models/financing_statement.py
+++ b/ppr-api/src/ppr_api/models/financing_statement.py
@@ -209,9 +209,11 @@ class FinancingStatement(db.Model):  # pylint: disable=too-many-instance-attribu
             if party.party_type == party_type or \
                (party_type == Party.PartyTypes.DEBTOR_COMPANY.value and
                 party.party_type == Party.PartyTypes.DEBTOR_INDIVIDUAL.value):
-                if party.registration_id == registration_id and \
-                    (not party.registration_id_end or \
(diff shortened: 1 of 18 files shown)


ppr-api/src/ppr_api/models/financing_statement.py
```diff
@@ -209,9 +209,18 @@ def party_json(self, party_type, registration_id):
             if party.party_type == party_type or \
                (party_type == Party.PartyTypes.DEBTOR_COMPANY.value and
                 party.party_type == Party.PartyTypes.DEBTOR_INDIVIDUAL.value):
-                if not self.current_view_json and party.registration_id == registration_id:
+                if party.registration_id == registration_id and \
```
jordiwes: just wondering if you can add a few comments on this if block, I'm having trouble following it ;)

jordiwes (approved):
Just added a few comments, but otherwise looks good

kialj876 (approved):

ppr-api/src/ppr_api/models/financing_statement.py
doug-lovett: Comments added.

jordiwes: thanks!

PR #280, opened 2021-10-06
9134 court order component

jordiwes:
*Issue #:* /bcgov/entity#9134
*Description of changes:*
-court order component for amendment 
- adding to pages are separate tickets


ppr-ui/src/components/common/CourtOrder.vue
```diff
@@ -0,0 +1,293 @@
...
+          ? convertDate(
+            new Date(
+              getCourtOrderInformation.value.orderDate + 'T09:00:00Z'
+            ),
+            false,
+            false
+          )
+          : ''
+      }),
+      showErrors: props.setShowErrors
```
kialj876: fyi if you make this a computed value that returns the props.setShowErrors you can remove the watch on it

ppr-ui/src/components/common/CourtOrder.vue
```diff
@@ -0,0 +1,293 @@
...
+      isValidCourtOrderForm
+    } = useCourtOrderValidation()
+    const modal = false
+    const localState = reactive({
+      courtName: '',
+      courtRegistry: '',
+      fileNumber: '',
+      orderDate: '',
+      effectOfOrder: '',
+      courtOrderInfo: getCourtOrderInformation.value as CourtOrderIF,
```
kialj876: I think this would be better as a computed value so that it updates with the state. Right now it won't matter, but if anything outside this component were to change it (i.e. a reset function or something) then this component wouldn't get the change

kialj876 (commented):
Hey Steph, looks good. Just a couple things to consider

doug-lovett (approved):
Nothing to add in addition to Kial's comments.

ppr-ui/src/components/common/CourtOrder.vue
jordiwes: good point

ppr-ui/src/components/common/CourtOrder.vue
jordiwes: I made these changes

kialj876 (approved):

PR #282, opened 2021-10-07
API new endpoint for ui add registration.

doug-lovett:

*Issue #:* /bcgov/entity#8999

*Description of changes:*
- New endpoint required by UI to get summary endpoint information before adding a registration from another account.
- Endpoint path: /ppr/api/v1/financing-statements/registrations/{registration_num}
- Return unauthorized 401 instead of 400 bad request when account cannot access a registration request (belongs to another account).


kialj876 (approved):
Looks good! Thanks for making this change so quickly :)

jordiwes (approved):
Looks good!

PR #313, opened 2021-10-18
API add/delete my registrations include user account registrations.

doug-lovett:

*Issue #:* /bcgov/entity#8999

*Description of changes:*
- Allow registrations created by the user account to be added/removed from the my registrations list.
- Update GET individual registration summary to include new "inUserList" boolean property. The property value is true if the registration is already in the user my registrations list.


kialj876 (approved):
Looks good

PR #317, opened 2021-10-19
API user profile changes for UI reg table.

doug-lovett:

*Issue #:* /bcgov/entity#9112

*Description of changes:*
- Update API UserProfile schema to add registrationsTable object type. Value can be any valid JSON.
- UI can use GET/PATCH /ppr/api/v1/user-profile to fetch/update user preferences for the display of the registrations table.  
- 9653 bug fix for financing statement verification report when original secured party, debtor has been updated.


ppr-api/tests/unit/api/test_user_profile.py
```diff
@@ -100,6 +122,7 @@ def test_update_user_profile(session, client, jwt, desc, staff, include_account,
                       headers=headers,
                       content_type='application/json')
     # check
+    print(rv.json)
```
kialj876: not sure if you meant to keep this print

kialj876 (approved):
Looks good, thanks for allowing the generic column names

ppr-api/tests/unit/api/test_user_profile.py
doug-lovett: Thanks for catching this, print statement removed.

jordiwes (approved):
LGTM

PR #345, opened 2021-11-01
Updates to setup of save registration API call.

doug-lovett:

*Issue #:* /bcgov/entity#9143

*Description of changes:*
- Fix validation check of court order information and setup of API call.
- Fix conditional check on setup to add general collateral to an amendment registration.

### DIFF
--- a/ppr-ui/src/views/amendment/ConfirmAmendment.vue
+++ b/ppr-ui/src/views/amendment/ConfirmAmendment.vue
@@ -382,6 +382,11 @@ export default class ConfirmAmendment extends Vue {
   private get courtOrderValid (): boolean {
     const courtOrder: CourtOrderIF = this.getCourtOrderInformation
     return (!courtOrder ||
+            (courtOrder.courtName.length === 0 &&
+            courtOrder.courtRegistry.length === 0 &&
+            courtOrder.fileNumber.length === 0 &&
+            courtOrder.orderDate.length === 0 &&
+            courtOrder.effectOfOrder.length === 0) ||
             (courtOrder.courtName.length > 0 &&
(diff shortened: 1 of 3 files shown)


ppr-ui/src/views/amendment/ConfirmAmendment.vue
```diff
@@ -382,6 +382,11 @@ export default class ConfirmAmendment extends Vue {
   private get courtOrderValid (): boolean {
     const courtOrder: CourtOrderIF = this.getCourtOrderInformation
     return (!courtOrder ||
+            (courtOrder.courtName.length === 0 &&
+            courtOrder.courtRegistry.length === 0 &&
+            courtOrder.fileNumber.length === 0 &&
+            courtOrder.orderDate.length === 0 &&
+            courtOrder.effectOfOrder.length === 0) ||
```
kialj876, line 389: Hey Doug can you just have the conditionals below this as >= instead here?

doug-lovett: Kial, all court order data elements are mandatory: a character length of 0 will fail API validation and a database insert constraint.

jordiwes (approved):
Looks good to me... there's a test that needs fixing though

doug-lovett:
Thanks Steph, it's been that kind of a day.

ppr-ui/src/views/amendment/ConfirmAmendment.vue
kialj876: ah I see sorry -- I was thinking it would be the same check, but I see you're validating that every value is empty or every value has something

kialj876 (approved):

PR #348, opened 2021-11-02
Updates to resolve UI registrations table issues.

doug-lovett:

*Issue #:* /bcgov/entity#9197

*Description of changes:*
- Account registrations summary return empty string instead of None for registeringName and clientReferenceId
- Relax other accounts access restriction to financing statement/base registration data requests. Only enforce the restriction with pdf requests.


kialj876 (approved):

kialj876:
Linter is failing, but code looks good to me

doug-lovett:
This may be a problem with the linter. Identical code in a different endpoint does not generate these errors (lines 781 forward), and I do not see them locally. And the endpoint matching the linter errors works without error locally.

jordiwes (approved):

PR #369, opened 2021-11-09
Update save renewal API call for RL.

doug-lovett:

*Issue #:* /bcgov/entity#9919

*Description of changes:*
- Update save renewal API call for RL: add court order info, no lifeYears or lifeInfinite.
- 9989 UI fix account add registration spinning when reg type is an unrecognized legacy registration type.

### DIFF
--- a/ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
+++ b/ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
@@ -107,7 +107,8 @@ export interface RenewRegistrationIF {
   createDateTime?: string, // Included in a successful response.
   payment?: PaymentIF, // Included in a successful response.
   error?: ErrorIF,
-  authorizationReceived?: boolean // Always true for the UI: API requirement.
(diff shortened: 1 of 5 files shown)


ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
```diff
@@ -107,7 +107,8 @@ export interface RenewRegistrationIF {
   createDateTime?: string, // Included in a successful response.
   payment?: PaymentIF, // Included in a successful response.
   error?: ErrorIF,
-  authorizationReceived?: boolean // Always true for the UI: API requirement.
+  authorizationReceived?: boolean, // Always true for the UI: API requirement.
+  courtOrderInformation?: CourtOrderIF, // Only populated if RL renewal and all court order elements present.
```
jordiwes: does it let you have a trailing comma there?

doug-lovett: It does, but I removed the comma anyway.

PR #372, opened 2021-11-12
changes for repairers lien table rows

jordiwes:
*Issue #:* /bcgov/entity#9105

*Description of changes:*
- table row changes for repairers lien

### DIFF
--- a/ppr-ui/src/composables/useRegistration.ts
+++ b/ppr-ui/src/composables/useRegistration.ts
@@ -108,6 +108,18 @@ export const useRegistration = () => {
     return false
   }
 
+  const hasRenewal = (
+    item: RegistrationSummaryIF
+  ): boolean => {
+    // check each child
+    for (let i = 0; i < item.changes?.length || 0; i++) {
+      if ((item.changes[i] as any)?.registrationClass) {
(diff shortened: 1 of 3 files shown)


ppr-ui/src/composables/useRegistration.ts
```diff
@@ -108,6 +108,18 @@ export const useRegistration = () => {
     return false
   }
 
+  const hasRenewal = (
```
doug-lovett, line 111: Is this checking if there has been a previous renewal registration?

doug-lovett (approved):
Looks good, just one question.

ppr-ui/src/composables/useRegistration.ts
jordiwes: yep. it goes through the children to see if there has been a previous renewal

doug-lovett: So should it check that registrationClass === 'RENEWAL' or registrationType === 'RE'?

jordiwes: yes, good point. I've made the fix

PR #374, opened 2021-11-16
do not show invalid fees

jordiwes:
*Issue #:* /bcgov/entity#9041

*Description of changes:*
- do not show invalid fees

### DIFF
--- a/ppr-ui/src/composables/fees/FeeSummary.vue
+++ b/ppr-ui/src/composables/fees/FeeSummary.vue
@@ -101,7 +101,7 @@ export default defineComponent({
       feeType: props.setFeeType,
       registrationType: props.setRegistrationType,
       registrationLength: computed((): RegistrationLengthI => {
-        if (localState.isValid !== true) {
+        if ((localState.isValid !== true) && (props.setRegistrationLength)) {
           props.setRegistrationLength.lifeYears = 0
         }
         return props.setRegistrationLength
--- a/ppr-ui/src/composables/fees/FeeSummary.vue
+++ b/ppr-ui/src/composables/fees/FeeSummary.vue
@@ -75,6 +75,7 @@ import {
   toRefs,
   watch
 } from '@vue/composition-api'
+import { useGetters } from 'vuex-composition-helpers'
 // local
 import { UIRegistrationTypes } from '@/enums'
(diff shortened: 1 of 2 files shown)


kialj876 (approved):

kialj876 (changes requested):
Looks good to me but test needs to get fixed

ppr-ui/src/composables/fees/FeeSummary.vue
```diff
@@ -95,10 +96,17 @@ export default defineComponent({
...
   },
   setup (props) {
+    const { getLengthTrust } = useGetters<any>(['getLengthTrust'])
     const localState = reactive({
       feeType: props.setFeeType,
       registrationType: props.setRegistrationType,
-      registrationLength: props.setRegistrationLength,
+      registrationLength: computed((): RegistrationLengthI => {
+        if (localState.isValid !== true) {
+          props.setRegistrationLength.lifeYears = 0
```
kialj876, line 105: since props.setRegistrationLength can be null you will need to check it exists I think or use ? notation

doug-lovett (approved):

jordiwes:
tests fixed :)

ppr-ui/src/composables/fees/FeeSummary.vue
kialj876: Hey sorry Steph, I think this will still err in cerrtain situations where it's not set in our app. I think changing the prop in the child component is also a bad practice because it breaks the sync with the parent value. Do you think it would be better to put this validation in the length trust component and updating the life years there instead?

jordiwes: This change was made because Scott wanted to show "not valid" versus unset.... happy to discuss

kialj876: ah so you probably need the actual life years value somewhere else I guess. Okay well that's fine, but this will still throw errors in the app where props.setRegistrationLength is null won't it? Amendment + Discharge?

jordiwes: added a check for null in there.

kialj876 (approved):

PR #380, opened 2021-11-19
10046 staff payment

jordiwes:
*Issue #:* /bcgov/entity#10046

*Description of changes:*
- add staff payment dialog and button for staff
- another task is to "make it work"


doug-lovett (approved):
Looks good but some linter errors.

ppr-ui/src/components/dialogs/StaffPaymentDialog.vue
```diff
@@ -0,0 +1,230 @@
+<template>
+  <v-dialog v-model="setDisplay" width="720px" persistent :attach="attach">
+    <v-card id="staff-payment-dialog" class="pl-5 pr-1 pt-5">
+      <v-row no-gutters>
+        <v-col cols="11">
+          <h3 class="dialog-title pb-6">Staff Payment</h3>
+          <div>
```
kialj876: Hey steph can you please update this to use the base dialog component? You can use slots to replace the body with the staff component / checkbox and props for the title / buttons. You can see I did something similar for the confirmation dialogs. This way all the formatting etc. will be the same / change with anything we alter in the future and we wont have to refactor this with the remaining dialogs

ppr-ui/tests/unit/StaffPaymentDialog.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  it('renders with staff payment component', () => {
+    expect(wrapper.findComponent(StaffPaymentDialog).exists()).toBe(true)
+    expect(wrapper.findComponent(StaffPaymentComponent).exists()).toBe(true)
+    expect(wrapper.findAll(title).length).toBe(1)
+    expect(wrapper.find(title).text()).toBe('Staff Payment')
+    
+  })
+
+  it('Emits the cancel action', async () => {
```
kialj876, line 66: these will need to be tweaked to emit the action from the base dialog similar to the confirmation dialog tests

ppr-ui/tests/unit/StaffPaymentDialog.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  afterEach(() => {
+    wrapper.destroy()
+  })
+
+  it('renders with staff payment component', () => {
+    expect(wrapper.findComponent(StaffPaymentDialog).exists()).toBe(true)
+    expect(wrapper.findComponent(StaffPaymentComponent).exists()).toBe(true)
+    expect(wrapper.findAll(title).length).toBe(1)
+    expect(wrapper.find(title).text()).toBe('Staff Payment')
+
```
kialj876, line 63: can you add another test in here to test that the staff payment comp is passed the correct data (i.e. check the props) and that the payment data gets successfully updated to the store and one for validity? I.e. emit the data from the staff component and check the values. There should also be a test in here that ensures the entered data gets cleared for when they do a second one :)

kialj876 (changes requested):
Hey Steph, looks good I just have some comments I think are important. Let me know if you want to chat / have questions about anything :)

ppr-ui/tests/unit/StaffPaymentDialog.spec.ts
jordiwes: My next task will be to make this dialog "work" :) ie save to store etc

ppr-ui/tests/unit/StaffPaymentDialog.spec.ts
jordiwes: done

ppr-ui/src/components/dialogs/StaffPaymentDialog.vue
jordiwes: done

kialj876 (approved):
Awesome, looks great 👍

PR #388, opened 2021-11-25
10088 pass staff data to api

jordiwes:
*Issue #:* /bcgov/entity#10088
/bcgov/entity#10047

*Description of changes:*
-finalize staff payment search

### DIFF
--- a/ppr-ui/src/utils/ppr-api-helper.ts
+++ b/ppr-ui/src/utils/ppr-api-helper.ts
@@ -1,6 +1,7 @@
 // Libraries
 import { axios } from '@/utils/axios-ppr'
 import { StatusCodes } from 'http-status-codes'
+import { StaffPaymentOptions } from '@bcrs-shared-components/enums'
 
 // Interfaces
 import {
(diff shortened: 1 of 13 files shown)


kialj876 (approved):
Looks good to me as long as Doug is on the same page for how the staff payment is getting sent -- not sure if he's planning to get it from the payload or not

ppr-ui/tests/unit/StaffPaymentDialog.spec.ts
```diff
@@ -76,4 +76,28 @@ describe('Payment component', () => {
...
+      bcolAccountNumber: '',
+      datNumber: '',
+      folioNumber: '',
+      isPriority: false
+    })
+    await flushPromises()
+    wrapper.findComponent(BaseDialog).vm.$emit(proceed, true)
+    await flushPromises()
+    expect(wrapper.vm.$store.state.stateModel.staffPayment.routingSlipNumber).toBe('999888777')
+  })
```
kialj876, line 102: also would be good to have a test that checks the staff payment info is reset on a new one

ppr-ui/src/utils/ppr-api-helper.ts
```diff
@@ -89,6 +91,29 @@ export async function search (
...
+    extraParams = extraParams + 'certified=True'
+  }
+  // filled out staff payment parameters
+  if (staffPayment) {
+    if (extraParams.length > 0) {
+      extraParams = extraParams + '&'
+    }
+    extraParams = extraParams + 'routingSlipNumber=' + staffPayment.routingSlipNumber
+    extraParams = extraParams + '&bcolAccountNumber=' + staffPayment.bcolAccountNumber
+    extraParams = extraParams + '&datNumber=' + staffPayment.datNumber
```
doug-lovett: The parameters are conditional on payment method:
If FAS payment method only include routingSlipNumber.
If BCOL payment only include bcolAccountNumber and datNumber.
If both a routing slip and a bcol account number are submitted the API will return an error.

doug-lovett (commented):
See my ppr-api-helper comment. I'm trying to avoid putting it in the payload so the schemas do not need to be modified as they are exposed in the spec and staff payment content may be confusing to external developers.

ppr-ui/src/utils/ppr-api-helper.ts
jordiwes: Made this change

ppr-ui/tests/unit/StaffPaymentDialog.spec.ts
jordiwes: added a test to search bar (it's where it resets)

doug-lovett (approved):
Looks good.

PR #417, opened 2021-12-15
finalized cronjob code

kialj876:
Signed-off-by: Kial <kialj876@gmail.com>

*Issue #:* /bcgov/entity#9834

*Description of changes:*
- added new config options for Bob
- added legacy pull from table instead of csv
- minor updates

### DIFF
--- a/cronjobs/search-tester/run_search_tester.py
+++ b/cronjobs/search-tester/run_search_tester.py
@@ -22,6 +22,7 @@ from flask import Flask
 from ppr_api.models import db, SearchRequest, TestSearch, TestSearchBatch, TestSearchResult
 
 from search_tester import create_app
+from search_tester.utils.db_utils import QUERY_LEGACY_RESULTS_DATE, QUERY_LEGACY_RESULTS_DATE_TIME, QUERY_LEGACY_RESULTS_MOST_RECENT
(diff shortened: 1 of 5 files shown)


cronjobs/search-tester/run_search_tester.py
```diff
@@ -64,6 +65,65 @@ def add_api_results(search: TestSearch, results: List[dict]):
...
+                'index': row[row_keys[6]]
+            })
+    return batch_searches
+
+
+def get_batch_searches_csv(app: Flask, batch_searches: dict) -> dict:
+    """Get batches from csv."""
+    # read csv + prep for search tests
+    filename = app.config['FILE_NAME']
+    with open(f'csvs/{filename}', 'r') as csvfile:
```
jordiwes, line 106: Does this need a try/catch?

jordiwes (approved):
One question, otherwise looks good

doug-lovett (approved):
LGTM.

cronjobs/search-tester/run_search_tester.py
kialj876: The fn is called within a try catch so I think it's okay. It's a temp job so I don't think we need to worry about it too much

PR #450, opened 2022-01-07
9508 date picker ux

kialj876:
*Issue #:* /bcgov/entity#9508

*Description of changes:*
- plugged in shared date picker
- needed some dependency files (mixins/interface)
- awaiting change to shared repo (I have in review) for hint props and then I will update the shared pkg version to .10


kialj876:
fyi the linting + testing are failing on a git permissions thing -- I don't think it has anything to do with my pr

doug-lovett (approved):
Very nice.

PR #452, opened 2022-01-11
sbc account changes

kialj876:

*Issue #:* /bcgov/entity#10440

*Description of changes:*
- sbc account changes

### DIFF
--- a/ppr-api/src/ppr_api/resources/financing_statements.py
+++ b/ppr-api/src/ppr_api/resources/financing_statements.py
@@ -26,7 +26,7 @@ from ppr_api.models import AccountBcolId, EventTracking, FinancingStatement, Reg
 from ppr_api.models import utils as model_utils
 from ppr_api.reports import ReportTypes, get_pdf
 from ppr_api.resources import utils as resource_utils
-from ppr_api.services.authz import authorized, is_reg_staff_account, is_sbc_office_account, is_staff_account, \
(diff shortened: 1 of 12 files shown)


doug-lovett (approved):
Very nice.

ppr-api/src/ppr_api/resources/financing_statements.py
```diff
@@ -886,7 +886,17 @@ def pay_and_save(req: request,  # pylint: disable=too-many-arguments,too-many-lo
...
     if not is_reg_staff_account(account_id):
-        pay_account_id: str = account_id if not is_sbc_office_account(account_id) else None
+        pay_account_id: str = account_id
+        # if gov user then check if sbc staff
+        if is_gov_account(jwt):
+            # if SBC staff user (authy, api call) then set payment account id to None
+            is_sbc = is_sbc_office_account(jwt.get_token_auth_header(), account_id)
+            if is_sbc:
+                pay_account_id = None
+            elif is_sbc == None:
```
jordiwes, line 896: I just had a question about this... if the user isn't sbc does it raise an exception?

jordiwes (approved):
Looks good

ppr-api/src/ppr_api/resources/financing_statements.py
kialj876: if the call returns successfully and they are not sbc staff then it will return false. None happens if there was an issue or if there's a case where they didn't have an account id

PR #458, opened 2022-01-11
Search result exact match display reg number.

doug-lovett:

*Issue #:* /bcgov/entity#10343

*Description of changes:*
Search result exact match display reg number.


kialj876 (approved):
Looks good to me 👍

PR #459, opened 2022-01-11
fixed reg table date sort

kialj876:

*Issue #:* /bcgov/entity#10060

*Description of changes:*
- date-picker fix
- add registrations ux fix
- filter fix
- chrome autocomplete fix

### DIFF
--- a/ppr-ui/src/composables/useRegistration.ts
+++ b/ppr-ui/src/composables/useRegistration.ts
@@ -123,8 +123,8 @@ export const useRegistration = () => {
   const filterResults = (originalData: RegistrationSummaryIF[]): any[] => {
     const newTableData = [] // filtered list of base registrations
     for (let i = 0; i < originalData.length; i++) {
-      // start by setting expand to false if relevant
-      if (originalData[i].changes) originalData[i].expand = false
+      // start by setting expand to false if relevant (unless it is a newly added registration)
+      if (originalData[i].changes && !originalData[i].new) originalData[i].expand = false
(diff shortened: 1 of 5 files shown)


ppr-ui/src/composables/useRegistration.ts
```diff
@@ -206,15 +206,17 @@ export const useRegistration = () => {
       if (localState.submittedStartDate && localState.submittedEndDate) {
         const sDateFilter = (item: any): boolean => {
           if (item.createDateTime) {
-            const created = item.createDateTime.substring(0, 10)
-            if (created < localState.submittedStartDate) return false
+            const createdDate = new Date(item.createDateTime)
+            const localDateStr = createdDate.toLocaleDateString('en-CA')
```
jordiwes, line 210: Do we want to set the time zone on this?

jordiwes (approved):
One question, looks good

doug-lovett (approved):

doug-lovett (approved):

ppr-ui/src/composables/useRegistration.ts
kialj876: good question, let me look

kialj876: alright so I actually took out the timezone in the datepicker so that it's for whatever the users local timezone is. This is because that's also how it displays in the table and we don't show the timezone anywhere for now (we may come back to this if the ux team wants to though)

PR #488, opened 2022-01-19
9973 reg table ux items

kialj876:
*Issue #:* /bcgov/entity#9973

*Description of changes:*
- reg table ux


jordiwes (approved):
Interesting solution there, I like it

PR #504, opened 2022-01-21
always display something for cert name

kialj876:

*Issue #:* /bcgov/entity#10750

*Description of changes:*
updated cert name logic to always display something if expected fields are not in the token


doug-lovett (approved):
else we've got nothing

jordiwes (approved):
Good stuff

PR #530, opened 2022-01-27
leap year issue fixed

kialj876:

*Issue #:* /bcgov/entity#10845

*Description of changes:*
- leap year bug fixed

### DIFF
--- a/ppr-api/src/ppr_api/models/utils.py
+++ b/ppr-api/src/ppr_api/models/utils.py
@@ -19,9 +19,9 @@ between the API and the database in both directions.
 from datetime import date  # noqa: F401 pylint: disable=unused-import
 from datetime import datetime as _datetime
 from datetime import time, timedelta, timezone
-from dateutil.relativedelta import relativedelta
 
 import pytz
+from datedelta import datedelta
 from flask import current_app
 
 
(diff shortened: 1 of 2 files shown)


doug-lovett (approved):
Thanks, I see locally there is a renewal for the GRIFCO base reg 611196G that calculates the renewal expiry date as 2022-03-01.

ppr-api/src/ppr_api/models/utils.py
```diff
@@ -19,6 +19,7 @@
 from datetime import date  # noqa: F401 pylint: disable=unused-import
 from datetime import datetime as _datetime
 from datetime import time, timedelta, timezone
+from dateutil.relativedelta import relativedelta
```
thorwolpert: Can we try to move to 1 date library?
datedelta has the expected behaviour

There are two date arithmetic traps in the Gregorian calendar:

Leap years. Problems arise when adding years to a February 29th gives a result in a non-leap year.
Variable number of days in months. Problems arise when adding months to a 29th, 30th or 31st gives a result in a month where that day doesn’t exist.
In both cases, datedelta changes the result to the first day of the next month.

thorwolpert (commented):
Can we use 1 library for this?

ppr-api/src/ppr_api/models/utils.py
kialj876: sure I can change it to that, but we don't use datedelta anywhere else that I can see? dateutil is also an internal python library so it might be better to use that? I'm just thinking about for when we do upgrades to python 4, 5 etc. it will probably always be stable with the python datetime library instead of us having to manage the correct version of datedelta for each python version

kialj876: ~Oh hang on did you mean timedelta? I will try using that~ is there a reason we want datedelta over dateutil? We don't use it anywhere currently and it looks like it does the exact same thing (for year and month etc.) except that its not an internal python library?

kialj876: changing to datedelta to keep the same as other projects (lear, etc.)

jordiwes (approved):

thorwolpert (approved):

PR #538, opened 2022-01-31
10223- load secured parties

jordiwes:
*Issue #:* /bcgov/entity#10223

*Description of changes:*
- unrelated to this ticket, but don't condition load of secured parties on reg party

### DIFF
--- a/ppr-ui/src/components/parties/party/SecuredParties.vue
+++ b/ppr-ui/src/components/parties/party/SecuredParties.vue
@@ -541,7 +541,7 @@ export default defineComponent({
 
     const fetchOtherSecuredParties = async () => {
       localState.loading = true
-      if (parties.registeringParty) {
+      if (localState.loading) {
         // go to the service and see if there are similar secured parties
         const response: [SearchPartyIF] = await partyCodeAccount()
         // check if any results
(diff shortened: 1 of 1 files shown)


ppr-ui/src/components/parties/party/SecuredParties.vue
```diff
@@ -541,7 +541,7 @@ export default defineComponent({
 
     const fetchOtherSecuredParties = async () => {
       localState.loading = true
-      if (parties.registeringParty) {
+      if (localState.loading) {
```
kialj876: this seems wrong since loading is set to true right above this statement haha? Did you mean to take out this 'if' entirely? Or is there a case where the 'else' clause should be happening?

jordiwes: well, in a recent change, registering Party is set to null, so that no longer works, but it's not really needed anyways. If you look at the function, it tries to load the secured parties and if successful, it sets loading to false, otherwise sets a timeout for 3 seconds and calls the function again.

doug-lovett (approved):

ppr-ui/src/components/parties/party/SecuredParties.vue
kialj876: Sorry for the late response! Okay sounds good, can you take out the 'if' then? I can see that confusing us later haha

jordiwes: removed the if

kialj876 (approved):
👍

PR #539, opened 2022-02-01
Improve large account reg history behaviour

doug-lovett:

*Issue #:* /bcgov/entity#NA

*Description of changes:*
Limit account registration history results to 1000, but apply the limit after excluding the account removed registrations.
Currently the limit is applied before the account removed registrations (that were created by the account) are excluded.


thorwolpert (approved):

jordiwes (approved):
LGTM

PR #544, opened 2022-02-02
repr doesn't always return a string, and therefore concate will fail,…

thorwolpert:
*Issue #:* /bcgov/entity#10995

*Description of changes:*
repr doesn't always return a string, and therefore concatenation will, fail changing this one which account for a few hundred messages a day.


doug-lovett (approved):
Thanks

PR #545, opened 2022-02-02
10741 address region fix

jordiwes:
*Issue #:* /bcgov/entity#10741

*Description of changes:*
- force the app to register the region rules change

### DIFF
--- a/ppr-ui/src/composables/address/BaseAddress.vue
+++ b/ppr-ui/src/composables/address/BaseAddress.vue
@@ -99,7 +99,6 @@
             v-model="addressLocal.region"
           />
           <v-text-field v-else
-            autocomplete="new-password"
             filled
             class="item address-region"
             :label="regionLabel"
@@ -108,7 +107,6 @@
             :rules="[...schemaLocal.region]"
           />
           <v-text-field
-            autocomplete="new-password"
             filled
             class="item postal-code"
             :label="postalCodeLabel"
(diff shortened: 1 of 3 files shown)


doug-lovett (approved):

ppr-ui/src/composables/address/BaseAddress.vue
```diff
@@ -210,7 +208,10 @@ export default defineComponent({
         schemaLocal.value.region = origRegionRules
       } else {
         schemaLocal.value.postalCode = origPostalCodeRules.concat([baseRules.maxLength(15)])
-        schemaLocal.value.region = [baseRules.maxLength(2), ...spaceRules]
+        for (let i = 0; i < schemaLocal.value.region.length; i++) {
```
kialj876, line 211: hey steph, this should be `schemaLocal.value.region.length - 1` I think? Might not matter depending on how the array pop() works on an empty list

kialj876 (approved):
Just one question, but looks good

ppr-ui/src/composables/address/BaseAddress.vue
jordiwes: if the length is 3, it's going to pop 0, 1, and 2 , which are 3 entries.

I did discover that the concat wasn't behaving nicely though, so I made a change there

kialj876: Oh right yeah definitely haha. Uhhh but without the concat we're back to the same thing as before no?

kialj876: I guess if it works it works. Maybe since it does the 'pops' it triggers the change this way

kialj876 (approved):

PR #553, opened 2022-02-04
infinite scroll, api filter

kialj876:

*Issue #:* /bcgov/entity#10902

*Description of changes:*
- dependent on api changes

### DIFF
--- a/ppr-ui/src/utils/ppr-api-helper.ts
+++ b/ppr-ui/src/utils/ppr-api-helper.ts
@@ -19,11 +19,13 @@ import {
   RegistrationSummaryIF,
   RenewRegistrationIF,
   ErrorIF,
-  BaseHeaderIF
+  BaseHeaderIF,
+  RegistrationSortIF
 } from '@/interfaces'
 import { SearchHistoryResponseIF } from '@/interfaces/ppr-api-interfaces/search-history-response-interface'
 import { StaffPaymentIF } from '@bcrs-shared-components/interfaces' // eslint-disable-line no-unused-vars
-import { ErrorCodes, SettingOptions } from '@/enums'
+import { SettingOptions } from '@/enums'
+import { getRegistrationNumber } from '@/store/getters'
 
 /**
  * Actions that provide integration with the ppr api.
@@ -61,6 +63,18 @@ function staffPaymentParameters (staffPayment: StaffPaymentIF) {
     }
   }
   return paymentParams
+}
+
+const UIFilterToApiFilter = {
+  endDate: 'endDateTime',
+  folNum: 'clientReferenceId',
+  orderBy: 'sortCriteriaName',
+  orderVal: 'sortDirection',
+  regBy: 'registeringName',
+  regNum: 'regNumber',
+  regType: 'registrationType',
+  startDate: 'startDateTime',
+  status: 'statusType'
 }
 
 export const successfulPPRResponses = [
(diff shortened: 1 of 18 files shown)


jordiwes (approved):
Looks good to me.... this table observer stuff is all new to me!

kialj876:
> Looks good to me.... this table observer stuff is all new to me!

me too!

ppr-ui/tests/unit/Dashboard.spec.ts
```diff
@@ -488,9 +491,9 @@ describe('Dashboard add registration tests', () => {
     const getStub = sandbox.stub(axios, 'get')
     const getSearchHistory = getStub.withArgs('search-history')
     getSearchHistory.returns(new Promise(resolve => resolve({ data: { searches: [] } })))
-    const getMyRegDrafts = getStub.withArgs('drafts')
+    const getMyRegDrafts = getStub.withArgs('drafts?fromUI=true&sortCriteriaName=createDateTime&sortDirection=desc')
```
doug-lovett: Drafts and registrations use the same parameter set, please use startDateTime instead of createDateTime. Same for below.

ppr-ui/tests/unit/Dashboard.spec.ts
```diff
@@ -86,9 +88,9 @@ describe('Dashboard component', () => {
     getSearchHistory.returns(new Promise(resolve => resolve({ data: { searches: [] } })))
     const getDraft = getStub.withArgs(`drafts/${draftDocId}`)
     getDraft.returns(new Promise(resolve => resolve({ data: mockedDraftFinancingStatementAll })))
-    const getMyRegDrafts = getStub.withArgs('drafts')
+    const getMyRegDrafts = getStub.withArgs('drafts?fromUI=true&sortCriteriaName=createDateTime&sortDirection=desc')
```
doug-lovett: From 10903 notes, please use startDateTime instead of createDateTime for draft and registration timestamp sorting. Same for request parameters below.

ppr-ui/src/utils/ppr-api-helper.ts
```diff
@@ -63,6 +65,18 @@ function staffPaymentParameters (staffPayment: StaffPaymentIF) {
   return paymentParams
 }
 
+const UIFilterToApiFilter = {
+  endDate: 'endDateTime',
+  folNum: 'clientReferenceId',
+  orderBy: 'sortCriteriaName',
+  orderVal: 'sortDirection',
+  regBy: 'registeringName',
+  regNum: 'regNumber',
```
doug-lovett: 10903: regNumber should be registrationNumber

doug-lovett (changes requested):
Looks great just requesting some changes on API request parameter names: regNum changed to registrationNumber; createDateTime changed to startDateTime.

kialj876:
> Looks great just requesting some changes on API request parameter names: regNum changed to registrationNumber; createDateTime changed to startDateTime.

Ah darn -- currently the orderBy is based on the api field names completely. The header value needs to be the same as the api field in the response and then that is just passed through without any conversion. I'll need to write something to alter all the values for orderBy not just createDateTime

doug-lovett:
I think it is only createDateTime that has a different name than the request query parameters? I was following the instruction in 10903. If this is a hassle I can change it in the API.

kialj876:
> I think it is only createDateTime that has a different name than the request query parameters? I was following the instruction in 10903. If this is a hassle I can change it in the API.

Oh you're right! Sorry looks like I've got Monday brain haha

ppr-ui/src/utils/ppr-api-helper.ts
kialj876: done

ppr-ui/tests/unit/Dashboard.spec.ts
kialj876: done

ppr-ui/tests/unit/Dashboard.spec.ts
kialj876: done

doug-lovett (approved):

PR #558, opened 2022-02-04
10869 duplicate regs

kialj876:

*Issue #:* /bcgov/entity#10869

*Description of changes:*
- throttle on all submit calls (save/submit for all reg stuff + search), also overlay the page with the spinner so the user isn't confused when things are a bit slow
- added some spinner loading for loading the reg info on these workflows as well


doug-lovett (approved):
Looks great.

jordiwes (approved):
LGTM

PR #562, opened 2022-02-07
10521 - has amendment really changed

jordiwes:
*Issue #:* /bcgov/entity#10521

*Description of changes:*
- amendment changed badge fixes

### DIFF
--- a/ppr-ui/src/components/parties/composables/useDebtor.ts
+++ b/ppr-ui/src/components/parties/composables/useDebtor.ts
@@ -5,7 +5,7 @@ import { Months } from '@/resources/months'
 import { PartyAddressSchema } from '@/schemas'
 import { useParty } from '@/composables/useParty'
 import { ActionTypes, RegistrationFlowType } from '@/enums'
-import { checkAddress } from '@/composables/address/factories/address-factory'
+import { checkAddress, formatAddress } from '@/composables/address/factories/address-factory'
 import { cloneDeep, isEqual } from 'lodash'
 
 const initPerson = { first: '', middle: '', last: '' }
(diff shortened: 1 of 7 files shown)


doug-lovett (approved):

ppr-ui/src/components/parties/composables/useDebtor.ts
```diff
@@ -115,7 +115,14 @@ export const useDebtor = (props, context) => {
...
       }
     }
-
+    // format the original address before the compare
+    if (localState.originalDebtor) {
+      localState.originalDebtor.address = formatAddress(localState.originalDebtor.address)
+    }
+    // format the original birthdate the exact same before the compare
+    if (localState.originalDebtor.birthDate && localState.originalDebtor.birthDate.length > 10) {
+      localState.originalDebtor.birthDate = localState.originalDebtor.birthDate.substring(0, 10) + 'T00:00:00-08:00'
```
kialj876: will this register as a change once the time zone changes to 7 hours?

kialj876: just wondering if you'll need to use the date library to add the utc tz

jordiwes: changed it to the timezone thingy

ppr-ui/src/components/parties/composables/useDebtor.ts
```diff
@@ -115,7 +115,16 @@ export const useDebtor = (props, context) => {
...
-
+    // format the original address before the compare
+    if (localState.originalDebtor) {
+      localState.originalDebtor.address = formatAddress(localState.originalDebtor.address)
+    }
+    // format the original birthdate the exact same before the compare
+    if (localState.originalDebtor.birthDate && localState.originalDebtor.birthDate.length > 10) {
+      const originalDateOfBirth = new Date(localState.originalDebtor.birthDate)
+      localState.originalDebtor.birthDate = originalDateOfBirth.toLocaleDateString(
+        'en-CA', { timeZone: 'America/Vancouver' }).substring(0, 10) + 'T00:00:00-08:00'
```
kialj876, line 126: Oh no, if you do this I think you overwrite the timezone incorrectly still. Won't -08:00 put this in a different timezone when we go back to only 7 hours behind utc? I think you can either convert it to a full timestamp or get the offset with `getTimezoneOffset` and set the tz that way

jordiwes: It's the original date of birth, so doesn't actually affect anything... it's just for comparison vs the date of birth in component, just so it doesn't appear that it changed, when it hasn't. Does that make sense?

kialj876: Yeah definitely. So where is the original datetime set then? If it's the api and the api sets it correctly (i.e. -07:00 when the timezone changes) won't this register as a change? Otherwise never mind haha

kialj876: Sorry for dragging this out haha -- I am going to approve and let you make the call since you're more familiar with this piece

kialj876 (approved):

PR #575, opened 2022-02-10
6925 user error

jordiwes:
*Issue #:* /bcgov/entity#6925

*Description of changes:*
- confirm user has auth role


kialj876 (approved):

kialj876:
Just need to add the ppr role to the token in that test I think?

PR #576, opened 2022-02-10
11097 repairer lien table fix and secured party validation fix

jordiwes:
*Issue #:* /bcgov/entity#11097
/bcgov/entity#11062

*Description of changes:*

### DIFF
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -166,7 +166,7 @@
             <span>Edit</span>
           </v-btn>
           <v-btn
-            v-else-if="isRepairersLien(item) && !isDischarged(item)"
+            v-else-if="isRepairersLien(item) && isActive(item)"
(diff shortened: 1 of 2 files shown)


ppr-ui/src/components/tables/common/TableRow.vue
```diff
@@ -166,7 +166,7 @@
             <span>Edit</span>
           </v-btn>
           <v-btn
-            v-else-if="isRepairersLien(item)"
+            v-else-if="isRepairersLien(item) && !isDischarged(item)"
```
kialj876: I think this should be `isRepairersLien(item) && isActive(item)` because that way it covers if it is expired as well ?

doug-lovett (approved):

ppr-ui/src/components/tables/common/TableRow.vue
jordiwes: updated to isActive

kialj876 (approved):

PR #588, opened 2022-02-15
11113 report gc add delete fix

doug-lovett:
*Issue #:* /bcgov/entity#10903
                 /bcgov/entity#11113

*Description of changes:*
- 10903 API account reg filtering, querying fixes:
1. Trim reg number filter value, fix not setting expand=True when matching on change registration number
2. Sort change registrations in descending order by registration timestamp.
3. Fix to include all change registrations when paging (beyond the first page).

- 11113 reports gc section display delete before add
- 10796 improve RL renewal unit tests


kialj876 (approved):
Looks good!

jordiwes (approved):

PR #590, opened 2022-02-16
6925 error handling

kialj876:
*Issue #:* /bcgov/entity#6925

*Description of changes:*
- error handling centralized in app
- added category in error obj


ppr-ui/src/App.vue
```diff
@@ -362,46 +360,67 @@ export default class App extends Mixins(AuthMixin) {
...
       const settings: UserSettingsIF = await getPPRUserSettings()
       userInfo.settings = settings
       this.setUserInfo(userInfo)
-      if (!settings || settings?.error) {
-        // error popup -> user may still continue
-        throw new Error('Invalid user settings')
+      if (settings?.error) {
+        message = 'Unable to get user settings.'
+        statusCode = settings.error.statusCode
       }
```
jordiwes, line 424: Should we return out of here if no error?

ppr-ui/src/mixins/auth-mixin.ts
```diff
@@ -14,6 +14,7 @@ export default class AuthMixin extends Vue {
   async fetchCurrentUser (): Promise<any> {
     const authUrl = sessionStorage.getItem('AUTH_API_URL')
     const config = { baseURL: authUrl }
+    console.log(config)
```
jordiwes: remove console?

ppr-ui/src/views/amendment/ConfirmAmendment.vue
```diff
@@ -650,16 +653,7 @@ export default class ConfirmAmendment extends Vue {
...
-      await this.loadRegistration()
-    } catch (error) {
-      const errorMsg = error as string
-      console.error(errorMsg)
-      this.emitError({
-        statusCode: 500,
-        message: errorMsg
-      })
-    }
+    await this.loadRegistration()
```
jordiwes, line 645: do you want to handle any error returned from here?

ppr-ui/src/App.vue
kialj876: Nope -- if successful the obj returned be: `{ category:..., message: '', statusCode: 200 }`. If you returned `{ statusCode: 200 }` here in an 'else' it would also work, but I think that's just unneeded extra code / more confusing

ppr-ui/src/views/amendment/ConfirmAmendment.vue
kialj876: it's already handled within the fn so there shouldn't ever be an error at this level

jordiwes (approved):
Cool LGTM

doug-lovett (approved):
Wow. Great work.

PR #610, opened 2022-02-23
11252 - bcol registering party code

jordiwes:
*Issue #:* /bcgov/entity#11252

*Description of changes:*
hard code bcol registering party code

### DIFF
--- a/ppr-ui/src/utils/auth-helper.ts
+++ b/ppr-ui/src/utils/auth-helper.ts
@@ -42,9 +42,9 @@ export function getKeycloakRoles (): Array<string> {
 }
 
 export async function getStaffegisteringParty (isBcOnline: boolean): Promise<PartyIF> {
-  let partyCode = '99990008'
+  let partyCode = sessionStorage.getItem(SessionStorageKeys.PprStaffPartyCode)
   if (isBcOnline) {
-    partyCode = '99980001'
+    partyCode = sessionStorage.getItem(SessionStorageKeys.BcolStaffPartyCode)
   }
   const response: [SearchPartyIF] = await partyCodeSearch(
     partyCode,
--- a/ppr-ui/src/utils/auth-helper.ts
+++ b/ppr-ui/src/utils/auth-helper.ts
@@ -42,9 +42,9 @@ export function getKeycloakRoles (): Array<string> {
 }
 
 export async function getStaffegisteringParty (isBcOnline: boolean): Promise<PartyIF> {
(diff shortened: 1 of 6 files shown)


ppr-ui/src/utils/auth-helper.ts
```diff
@@ -41,65 +41,68 @@ export function getKeycloakRoles (): Array<string> {
   throw new Error('Error getting Keycloak roles')
 }
 
+export async function getStaffegisteringParty (isBcOnline: boolean): Promise<PartyIF> {
+  let partyCode = '99990008'
```
kialj876: Can you add these codes as config values? That way if they're ever changed in dev/test/prod we don't need to rerelease. Once you add them into the code I can update 1pass to include them in the deploy

doug-lovett (approved):
Looks good, I like Kial's suggestion.

ppr-ui/src/utils/auth-helper.ts
jordiwes: done, let me know if it makes sense

ppr-ui/src/utils/auth-helper.ts
```diff
@@ -41,65 +41,68 @@ export function getKeycloakRoles (): Array<string> {
   throw new Error('Error getting Keycloak roles')
 }
 
+export async function getStaffegisteringParty (isBcOnline: boolean): Promise<PartyIF> {
+  let partyCode = sessionStorage.getItem(SessionStorageKeys.PprStaffPartyCode)
```
kialj876: SessionStorageKeys is just for the common components (it doesn't include our own config stuff that we put in session storage). I don't think we have an enum or anything for it so you just need to do the string *PPR_STAFF_PARTY_CODE*

Same for below

ppr-ui/src/utils/auth-helper.ts
kialj876: yeah looks good, you just need to update how you get the session storage value which I mentioned below

kialj876:
fyi I updated the config in 1pass to include PPR_STAFF_PARTY_CODE + BCOL_STAFF_PARTY_CODE in dev/test/prod, let me know if the code for those changes in any of the envs

ppr-ui/src/utils/auth-helper.ts
jordiwes: gotcha, made the change

kialj876 (approved):
thanks for the changes!

PR #611, opened 2022-02-23
non-billable fees update

kialj876:

*Issue #:* /bcgov/entity#10743

*Description of changes:*
- check if non billable + store it + their custom service fee
- update visible fees based on ^ on all flows
- secure with tests


ppr-ui/src/composables/fees/FeeSummary.vue
```diff
@@ -145,17 +149,25 @@ export default defineComponent({
         if (localState.feeType === FeeSummaryTypes.RENEW) {
           feeSummary.processingFee = 5
         }
+        if (props.setFeeOverride) {
+          feeSummary.feeAmount = props.setFeeOverride.feeAmount
+        }
+        if (props.setFeeOverride && feeSummary.serviceFee !== 0) {
```
jordiwes, line 155: Is there a circumstance where we want to override the service fee to 0?

ppr-ui/src/composables/fees/FeeSummary.vue
```diff
@@ -145,17 +149,25 @@ export default defineComponent({
...
         return props.setStaffReg || props.setStaffSBC
       }),
       hintFee: computed((): string => {
-        return getFeeHint(
+        const hint = getFeeHint(
           localState.feeType,
           localState.registrationType,
           localState.registrationLength
         )
+        if (props.setFeeOverride) return hint.replace('$5', '$0')
```
jordiwes, line 169: If the hint changes from $5?

ppr-ui/src/utils/axios-pay.ts
```diff
@@ -0,0 +1,28 @@
+import Axios from 'axios'
+import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
+import * as Sentry from '@sentry/browser'
+
+const axios = Axios.create()
+
+axios.interceptors.request.use(
```
jordiwes, line 9: what is this used for? just a quick comment would help :)

ppr-ui/src/composables/fees/FeeSummary.vue
kialj876: I don't think that's a thing, but even if it was then it would be fine here because the result is a service fee of 0. This is preventing the override where the normal service fee is 0 (business rules = override it to their custom service fee except when the service fee is 0)

ppr-ui/src/composables/fees/FeeSummary.vue
kialj876: this is just changing the *$5.00 / year* to *$0.00 / year*. The *$5.00 / year* never changes regardless of how many years they select, except for when it's a different hint (i.e. infinite.., etc.) and in that case we don't need anything to change

ppr-ui/src/utils/axios-pay.ts
kialj876: sure, I'll add a quick explanation

jordiwes (approved):

doug-lovett (approved):

PR #613, opened 2022-02-24
Report party former name fix.

doug-lovett:

*Issue #:* /bcgov/entity#11180
                 /bcgov/entity#11090

*Description of changes:*
- 11180 fix logic to check if an amendment party change is a name change.
- 11090 update search results sort order.
- 

### DIFF
--- a/ppr-api/src/ppr_api/models/registration.py
+++ b/ppr-api/src/ppr_api/models/registration.py
@@ -910,10 +910,29 @@ class Registration(db.Model):  # pylint: disable=too-many-instance-attributes, t
                     elif party.business_name:
                         former_name = party.business_name
                     else:
-                        former_name = party.last_name + ', ' + party.first_name
-                        if party.middle_initial:
-                            former_name += ' ' + party.middle_initial
-                    return former_name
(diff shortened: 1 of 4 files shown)


ppr-api/src/ppr_api/models/registration.py
```diff
@@ -910,10 +910,29 @@ def get_former_party_name(self, new_party: Party):
...
                     else:
-                        former_name = party.last_name + ', ' + party.first_name
-                        if party.middle_initial:
-                            former_name += ' ' + party.middle_initial
-                    return former_name
+                        # match if only 1 name is different in addition to same address.
+                        former_name = self.__get_matching_party_name(new_party, party)
+                    if former_name:
+                        return former_name
+        return former_name
```
jordiwes, line 917: Are there any instances where the first name and last name change?

doug-lovett: Then it's not a name change: cannot distinguish between a different person at the same address.

jordiwes (approved):

kialj876 (approved):
lgtm

PR #615, opened 2022-02-25
11083 vc tooltip

jordiwes:
*Issue #:* /bcgov/entity#11083

*Description of changes:*
- show tooltip for too many vehicle collateral

### DIFF
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -132,7 +132,7 @@
         <span class="pl-1">PDF</span>
       </v-btn>
       <v-tooltip
-        v-else-if="!isDraft(item) && !item.registeringParty"
+        v-else-if="!isDraft(item) && !item.registeringName"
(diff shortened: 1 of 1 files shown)


doug-lovett (approved):

ppr-ui/src/components/tables/common/TableRow.vue
```diff
@@ -146,6 +146,23 @@
           {{ tooltipTxtPdf }}
         </div>
       </v-tooltip>
+      <v-tooltip
+        v-else-if="!isDraft(item) && item.registeringName"
```
kialj876: hey Steph, instead of adding a new tooltip in the html I think this would be a lot nicer if you just make *tooltipTxtPdf* a computed with this condition / new text or a function that returns it based on the item. What do you think?

jordiwes: Done

kialj876 (approved):
Awesome, thanks!

PR #616, opened 2022-02-25
reg table ux update

kialj876:

*Issue #:* /bcgov/entity#11114

*Description of changes:*
- reg table info added to store to support all flows (can be used later for other things)
- adding a reg/draft will scroll to new reg with ux stuff
- tests updated


ppr-ui/src/store/state/state-model.ts
```diff
@@ -18,6 +18,29 @@ export const stateModel: StateModelIF = {
     certified: false,
     legalName: ''
   },
+  folioOrReferenceNumber: '',
```
kialj876, line 21: just moved this here so its alphabetical to make the store a little easier to navigate now that we have quite a few sections

ppr-ui/src/store/state/state-model.ts
```diff
@@ -18,6 +18,29 @@ export const stateModel: StateModelIF = {
...
+  // orig reg party used for discharge/renew/amend, other for amend only
+  originalRegistration: {
```
kialj876, line 23: same thing here

ppr-ui/src/store/state/state-model.ts
```diff
@@ -57,27 +80,9 @@ export const stateModel: StateModelIF = {
...
-      lienAmount: '',
-      surrenderDate: ''
-    },
-    parties: {
-      valid: true,
-      registeringParty: null,
-      securedParties: [],
-      debtors: []
-    }
+  registrationTable: {
```
kialj876, line 83: this is the only new part of the store I added

ppr-ui/src/store/state/state-model.ts
```diff
@@ -89,6 +94,7 @@ export const stateModel: StateModelIF = {
     searching: false,
     searchCertified: false
   },
+  staffPayment: null,
```
kialj876, line 97: same thing here as above (alphabetical move)

jordiwes (approved):
Phew! Good stuff!

doug-lovett (approved):
Looks good.

PR #626, opened 2022-03-02
cancel/exit flows update

kialj876:

*Issue #:* /bcgov/entity#10582

*Description of changes:*
- new dialogs + popups for all exit flows

### DIFF
--- a/ppr-ui/src/views/amendment/ConfirmAmendment.vue
+++ b/ppr-ui/src/views/amendment/ConfirmAmendment.vue
@@ -9,10 +9,9 @@
       <v-progress-circular color="primary" size="50" indeterminate />
     </v-overlay>
     <base-dialog
-      setAttach="#confirm-amendment"
       :setOptions="options"
       :setDisplay="showCancelDialog"
-      @proceed="cancel($event)"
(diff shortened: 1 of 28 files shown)


ppr-ui/src/views/amendment/ConfirmAmendment.vue
```diff
@@ -571,6 +574,7 @@ export default class ConfirmAmendment extends Vue {
     if (draft.error) {
       this.emitError(draft.error)
     } else {
+      this.setUnsavedChanges(false)
```
jordiwes, line 568: If they are on the confirm page, don't they have unsaved changes?

kialj876: ha - not if it's a draft they loaded in (they would never get here unless they had bookmarked the page and returned here directly, otherwise all the info would already be loaded in the store and the code above would skip this step)

jordiwes: ah! good point

jordiwes (approved):
LGTM

doug-lovett (approved):
Looks great.

PR #645, opened 2022-03-07
Parties.first_name_key length increase.

doug-lovett:

*Issue #:* /bcgov/entity#11393

*Description of changes:*
Increase db parties.first_name_key length to work with long individual debtor names.


ppr-api/src/ppr_api/models/party.py
```diff
@@ -65,7 +65,7 @@ class PartyTypes(Enum):
     email_id = db.Column('email_address', db.String(250), nullable=True)
 
     # Search keys
-    first_name_key = db.Column('first_name_key', db.String(50), nullable=True, index=True)
+    first_name_key = db.Column('first_name_key', db.String(100), nullable=True, index=True)
```
jordiwes, line 68: Generally curious why we are increasing the first name and not the last... I just put some validation in the UI to validate for 50 chars on all 3 names (first/middle/last)

doug-lovett: As a change within a few days of the prod release, Bob changed this key value from being on the debtor first name to being on the combination of first name and last name. As a consequence the column name is a little misleading.

kialj876 (approved):

jordiwes (approved):

PR #700, opened 2022-03-28
11601 mhr drop down

jordiwes:
*Issue #:* /bcgov/entity#11601

*Description of changes:*


doug-lovett (approved):
Looks good to me.

kialj876 (commented):
Looking good! Just needs some test fixes

ppr-ui/src/components/search/SearchBarList.vue
```diff
@@ -0,0 +1,194 @@
...
+      searchTypeLabel: computed((): string => {
+        if (localState.selectedSearchType?.searchTypeUI) {
+          return localState.selectedSearchType.searchTypeUI
+        }
+        return 'Select a search category'
+      }),
+      origItems: computed((): Array<SearchTypeIF> => {
+        const allSearchTypes = []
+        if (isRoleStaffReg.value) {
+          if (getFeatureFlag('bcregistry-ui-ppr-mhr-staff-only')) {
```
kialj876: hey steph this feature flag should be something like 'mhr_enabled' and then in launch darkly we specify who has access (staff, specific staff, specific users, etc.). This way we can turn off MHR for users at any time too while keeping ppr open

ppr-ui/src/components/search/SearchBarList.vue
```diff
@@ -0,0 +1,194 @@
...
+          }
+        }
+        if (hasPprRole.value) {
+          allSearchTypes.push.apply(allSearchTypes, SearchTypes)
+          // we can pop the title off if there is only one search type
+          if (!hasMhrRole.value) {
+            allSearchTypes.shift()
+          }
+        }
+        if (hasMhrRole.value) {
```
kialj876, line 114: same thing here, you can use the same feature flag instead of the role so that we can turn it off

kialj876 (commented):
Looks good steph I just have a couple small comments and test needs a fix

ppr-ui/src/components/search/SearchBarList.vue
jordiwes: I updated the feature flag name

PR #702, opened 2022-03-28
update common component

jordiwes:
*Issue #:* /bcgov/entity#11653

*Description of changes:*

### DIFF
--- a/ppr-ui/src/components/dialogs/StaffPaymentDialog.vue
+++ b/ppr-ui/src/components/dialogs/StaffPaymentDialog.vue
@@ -79,7 +79,6 @@ export default defineComponent({
     ])
     const { getStaffPayment } = useGetters<any>(['getStaffPayment'])
     const localState = reactive({
-      validate: true,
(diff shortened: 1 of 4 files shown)


doug-lovett (approved):
LGTM.

ppr-ui/src/components/dialogs/StaffPaymentDialog.vue
```diff
@@ -7,7 +7,7 @@
     <template v-slot:content>
       <staff-payment-component
         :staffPaymentData="staffPaymentData"
-        :validate="validate"
+        :validate="true"
```
kialj876, line 10: I think you meant to put this back to 'validate' right? Otherwise it doesn't look like 'validate' is being used for anything so you can probably remove it from the code below?

jordiwes: Removed it

kialj876 (approved):
LGTM

PR #708, opened 2022-03-29
update evaluation so that it shows in ld dash

kialj876:
Signed-off-by: Kial Jinnah <kial@Kials-MBP.vn.shawcable.net>

*Issue #:* /bcgov/entity#11397

*Description of changes:*
- functionality is still the same, but launch darkly dashboard tracks when it is evaluated it this way


doug-lovett (approved):
Nice.

PR #714, opened 2022-03-30
case insensitive for folio + reg name filter on drafts

kialj876:
Signed-off-by: Kial Jinnah <kial@Kials-MBP.vn.shawcable.net>

*Issue #:* /bcgov/entity#11397

*Description of changes:*
- added case insensitivity to drafts for folio and reg name
- protected by ff
- small update to tests that would've broken


doug-lovett (approved):
Looks good.

PR #731, opened 2022-04-11
11758 - search anywhere in name

jordiwes:
*Issue #:* /bcgov/entity#11758

*Description of changes:*
- search anywhere in registration name, not only starts with


doug-lovett (approved):
LGTM

PR #734, opened 2022-04-12
11758 - draft search same as financing statements

jordiwes:
*Issue #:* /bcgov/entity#11758

*Description of changes:*


doug-lovett (approved):
Sorry, I should have caught this with the first change.

PR #754, opened 2022-04-25
MHR Results and Review Updates

cameron-eyds:
*Issue #:* /bcgov/entity#11600 & /bcgov/entity#11597

*Description of changes:*
* Repurposed RESULTS table to include a review mode. 
* Replaced previous REVIEW component with Results table in Review mode to meet updated designs.
* Updated data handling of 'Selected' MHR Results.
* Table functionality and design updates includes style changes and individual checkbox selectors and Select All selectors.
* Updated Unit tests


cameron-eyds:
Will ask Patrick to have a peek at the CodeCov configurations, appears to be having an issue uploading the report.

doug-lovett (approved):
Looks good to me.

ppr-ui/src/components/tables/mhr/SearchedResults.vue
```diff
@@ -121,55 +167,96 @@ export default defineComponent({
   components: {
     FolioNumber
   },
+  props: {
+    isReviewMode: { default: false }
```
jordiwes, line 171: we usually have to load this into a computed value as the props values sometimes don't register the update... not sure if you will find this to be true :)

jordiwes (approved):
One small comment, but looks great :)

ppr-ui/src/components/tables/mhr/SearchedResults.vue
cameron-eyds: Will look into this, although i didn't notice this in development. It could be perhaps since it's re-mounted when we mount this component in Review Mode. 
But yea, thanks for the heads up here. Will look at this! 

Specifically, were you finding that the components were not being reactive to props in general?

cameron-eyds:
> One small comment, but looks great :)

Thank you!

PR #759, opened 2022-04-28
Missed fix to computed registration expiry with GETs.

doug-lovett:

*Issue #:* /bcgov/entity#11944

*Description of changes:*
- Fix with GET financing statement when expiry is computed from registration timestamp.


cameron-eyds (approved):
Looks good!

PR #766, opened 2022-05-02
Updated Regex to filter zero space hidden characters

cameron-eyds:
*Issue #:* /bcgov/entity#11255

*Description of changes:*
* Updated regex expression


doug-lovett (approved):
Very nice.

PR #774, opened 2022-05-09
MHR api setup cloud storage for reports.

doug-lovett:

*Issue #:* /bcgov/entity#12063

*Description of changes:*
Set up for MHR api reports document storage.


cameron-eyds (approved):
Looks good to me! 👍

mhr_api/src/mhr_api/services/document_storage/storage_service.py
```diff
@@ -0,0 +1,211 @@
...
+from mhr_api.services.gcp_auth.token_service import GoogleStorageTokenService
+from mhr_api.services.utils.exceptions import StorageException
+
+
+HTTP_DELETE = 'delete'
+HTTP_GET = 'get'
+HTTP_POST = 'post'
+
+
+class DocumentTypes(str, Enum):
```
thorwolpert: I think this should use our BaseEnum, so that the same interfaces is there for all Registry Enums.

mhr_api/src/mhr_api/services/document_storage/storage_service.py
```diff
@@ -0,0 +1,211 @@
...
+    """Render an Enum of storage document types."""
+
+    SEARCH_RESULTS = 'SEARCH_RESULTS'
+    REGISTRATION = 'REGISTRATION'
+
+
+class StorageService(ABC):  # pylint: disable=too-few-public-methods
```
thorwolpert: In mine I separated out the abstract class so that the "interface" could be imported without dragging along the implementation specific libraries.

mhr_api/src/mhr_api/services/gcp_auth/token_service.py
```diff
@@ -0,0 +1,79 @@
...
+"""This maintains access tokens for API calls."""
+import os
+from abc import ABC, abstractmethod
+
+import google.auth.transport.requests
+from google.oauth2 import service_account
+from flask import current_app
+
+
+class TokenService(ABC):  # pylint: disable=too-few-public-methods
```
thorwolpert: Do you want this to be explicit, or pick up the service account assigned to this. We're not consistent here and should choose a singular approach.

thorwolpert (commented):
Just some discussion items

mhr_api/src/mhr_api/services/document_storage/storage_service.py
doug-lovett: Refactored to use BaseEnum

mhr_api/src/mhr_api/services/document_storage/storage_service.py
doug-lovett: Separated out the abstract class.

mhr_api/src/mhr_api/services/gcp_auth/token_service.py
doug-lovett: Refactored to follow your approach on obtaining service account credentials either as one env var or from the google env.

PR #787, opened 2022-05-16
keep folio field for staff

garyfan1:
*Issue #:* /bcgov/entity#12074

*Description of changes:*
keep folio or reference number text filed for staff.
image

### DIFF
--- a/ppr-ui/src/components/search/SearchBar.vue
+++ b/ppr-ui/src/components/search/SearchBar.vue
@@ -322,9 +322,6 @@ export default defineComponent({
       isStaffBcolReg: computed((): boolean => {
         return isRoleStaffBcol.value || isRoleStaffReg.value
       }),
-      isStaffSbc: computed((): boolean => {
-        return isRoleStaffSbc.value
-      }),
       searching: computed((): boolean => {
         return isSearching.value
       }),
--- a/ppr-ui/src/components/search/SearchBar.vue
+++ b/ppr-ui/src/components/search/SearchBar.vue
@@ -40,7 +40,7 @@
           </v-tooltip>
         </div>
       </v-col>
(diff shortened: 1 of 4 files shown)


cameron-eyds (commented):
Looks good so far.
Will need to update the unit testing to verify it's ALWAYS shown. 

Also, see comments in our Assets channel, as we've now moved to node version `16.14.2`, all the front end contributors will need to be on the same version.

dimak1 (changes requested):
Agree with Cameron. Please add some tests.

ppr-ui/src/components/search/SearchBar.vue
```diff
@@ -40,7 +40,7 @@
           </v-tooltip>
         </div>
       </v-col>
-      <v-col v-if="!isStaffBcolReg && !isStaffSbc" align-self="end" cols="3">
+      <v-col align-self="end" cols="3">
```
cameron-eyds, line 43: Was there further commits here to render for specific types of Staff or is the requirement to display for all, client and staff?

cameron-eyds (approved):
Thanks for updating the test.
Looks good to me, just the one question. 
You mentioned previously that it might be rendered for only specific types of staff but if no further change is required, you are g2g 👍

ppr-ui/src/components/search/SearchBar.vue
dimak1: From the story, the requirement says it should appear for all staff users (in addition to clients). Right @garyfan1 ?

garyfan1: I checked with Melissa and the field is shown in all cases.

dimak1 (approved):

PR #796, opened 2022-05-20
Check SBC role for Tombstone and Breadcrumbs titles

dimak1:
*Issue #:* /bcgov/entity#11644

*Description of changes:*

Raised issue by QA that Tombstone and Breadcrumbs titles were not showing as expected. After an investigation it appeared we were not checking the SBC flag when showing titles. SBC flag is an indicator (outside of `authRoles`) to identify an SBC Staff, therefore titles should show as 'Staff ...' 

- Check SBC role when rendering Tombstone and Breadcrumbs titles
- Update tests


ppr-ui/src/store/actions/actions-model.ts
```diff
@@ -56,14 +57,14 @@ export const setOriginalAddCollateral: ActionIF = ({ commit }, addCollateral: Ad
...
 
-  if (authRoles.includes('staff')) accessRole.push('STAFF')
+  if (auth.authRoles.includes('staff') || auth.isSbc) accessRole.push('STAFF')
   else accessRole.push('CLIENT')
 
-  if (authRoles.includes('ppr')) accessRole.push('PPR')
-  if (authRoles.includes('mhr')) accessRole.push('MHR')
+  if (auth.authRoles.includes('ppr')) accessRole.push('PPR')
+  if (auth.authRoles.includes('mhr')) accessRole.push('MHR')
   commit('mutateUserAccessRole', accessRole.join('_'))
```
cameron-eyds: In state getters line 276, we have a computed value which identifies `isRoleStaffSbc`. 
Is that no longer required by utilizing the same flag and adding it into this array?

If that computed value is no long used in the App could we remove it?
If it is being used, we might need to look closer, as we don't want to have two means of identifying the same role/condition.

ppr-ui/src/App.vue
```diff
@@ -388,17 +388,18 @@ export default class App extends Mixins(AuthMixin) {
...
-          const isSbc = await getSbcFromAuth()
+          isSbc = await getSbcFromAuth()
           this.setRoleSbc(isSbc)
         }
         if (!authRoles.includes('ppr') && !authRoles.includes('mhr')) {
           throw new Error('No access to Assets')
         }
         this.setAuthRoles(authRoles)
-        this.setUserAccessRole(authRoles)
+        this.setUserAccessRole({ authRoles: authRoles, isSbc })
```
cameron-eyds: Sort of same question as below, would the state getter that exists already for identifying the SBC role not suffice?
Just curious why we apply it to another array when it's already defined in store as `stateModel.authorization.isSbc`

cameron-eyds (commented):
Looks good so far, just a few questions. 
Also, check the linter. Seems to be some small spacing issue in the `actions` 👍

ppr-ui/src/store/actions/actions-model.ts
dimak1: That computed property is used throughout the app. 

In this case, I'm only passing down an SBC boolean flag, and not reusing it. I didn't want to call a getter in a setter file. Open for a discussion.

ppr-ui/src/App.vue
dimak1: It would suffice, I thought calling a `isRoleStaffSbc` in `setUserAccessRole` setter didn't look clean to me, so I thought just to pass it down as a prop.

ppr-ui/src/store/actions/actions-model.ts
cameron-eyds: For sure, i agree with the latter. 
The question is more with regards to using the getter itself than adding it to a reconstructed roles array. 

It feels like a duplication if we already have a global computed property indicating said role, which is something we'd like to avoid if possible.

If there is a good reason, i'm all for it. Was just curious the reason why we pass it down vs using the pre existing flag?

ppr-ui/src/App.vue
cameron-eyds: For sure, I definitely agree with that aspect but same as above, not referring to its use of a prop here, more a question do we need to add it to this array at all, since there already is a global getter which will identify this role.


It seems like we have `UserAccessRole` driving some conditionals in the UI and those other global getters driving conditions elsewhere.

Is the end goal to have this array drive all our role based conditionals?

doug-lovett (approved):

dimak1:
@cameron-freshworks refactored to remove `UserAccessRole`, add utils instead, add more unit tests.

ppr-ui/src/App.vue
dimak1: Cleaned up.

ppr-ui/src/store/actions/actions-model.ts
dimak1: Cleaned up.

ppr-ui/src/utils/roles-helper.ts
```diff
@@ -0,0 +1,15 @@
+export function getDescriptiveUserRole (authRoles: Array<string>): string {
+  const accessRole = []
+
+  if (authRoles.includes('ppr_staff')) return 'STAFF_PPR'
+  if (authRoles.includes('mhr_staff')) return 'STAFF_MHR'
+  if (authRoles.includes('staff') && authRoles.includes('helpdesk')) return 'STAFF_PPR'
```
dimak1, line 6: Not sure about this combination. But I saw it in other tests, so added it here. @cameron-freshworks

cameron-eyds (approved):
Great job sir!

cameron-eyds:
> @cameron-freshworks refactored to remove `UserAccessRole`, add utils instead, add more unit tests.

Really nice clean up, thank you!

PR #803, opened 2022-05-25
12074 folio field for staff

garyfan1:
*Issue #:* /bcgov/entity#12074

*Description of changes:*
fix alignment issue

### DIFF
--- a/ppr-ui/src/components/search/SearchBar.vue
+++ b/ppr-ui/src/components/search/SearchBar.vue
@@ -30,7 +30,7 @@
       </v-col>
       <v-col align-self="end" cols="1" class="pl-3"/>
     </v-row>
-    <v-row v-if="typeOfSearch!=''">
+    <v-row v-if="typeOfSearch">
       <v-col>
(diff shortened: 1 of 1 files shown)


ppr-ui/src/components/search/SearchBar.vue
```diff
@@ -30,7 +30,7 @@
       </v-col>
       <v-col align-self="end" cols="1" class="pl-3"/>
     </v-row>
-    <v-row>
+    <v-row v-if="typeOfSearch!=''">
```
cameron-eyds: Would `v-if="typeOfSearch"` or `v-if="!!typeOfSearch"` work? Preferably one of these conditionals.

If not, you could keep the expression but i would use `!==` (type and value comparison operator) and also add some spacing.

garyfan1: just fixed it!

dimak1 (approved):

cameron-eyds (approved):
Looks good, thanks for the update 👍

PR #813, opened 2022-06-02
12139 serial num too long

garyfan1:
*Issue #:* /bcgov/entity#12139

*Description of changes:*
validate serial number even with MHRnumber entered

### DIFF
--- a/ppr-ui/src/utils/validators/serial-number.validator.ts
+++ b/ppr-ui/src/utils/validators/serial-number.validator.ts
@@ -16,8 +16,6 @@ export const serialNumberValidator = ({ values }) => {
     valueToValidate = values.manufacturedHomeRegistrationNumber.trim()
   }
 
-  console.log(valueToValidate)
-
   switch (values.type) {
     case 'AC':
       message = 'Maximum 25 characters'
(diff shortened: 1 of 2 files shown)


ppr-ui/src/utils/validators/serial-number.validator.ts
```diff
@@ -6,6 +6,7 @@ export const serialNumberValidator = ({ values }) => {
   let message = 'Maximum 25 characters'
   let emptyMessage = ''
   let valueToValidate = values.serialNumber?.trim()
+  const serialNumber = values.serialNumber?.trim()
```
cameron-eyds: I see what you're doing here but this might be misleading to reference the same property with two different references.

If you only want to validate `values.serialNumber` explicitly and always, i don't see the need for the second reference here. You can just validate `valueToValidate ` before a potential mutation. 

Or just validate `values.serialNumber` maxLen down below. 

That being said, by doing this you are excluding `valueToValidate = values.manufacturedHomeRegistrationNumber.trim()` from any maxLen validations, is that intended?

garyfan1: 1. ok I'll just validate `values.serialNumber` with maxLen
2. validation for mhrNumber (6 digits only) is taken care in `validateInput()` (which is a wrapper around `validateField()`)
image

cameron-eyds (approved):

dimak1 (approved):

dimak1:
@garyfan1 pls sync with `main` branch and I will merge it

doug-lovett (approved):

PR #814, opened 2022-06-03
Fix empty General Collateral on Review page

dimak1:
*Issue #:* /bcgov/entity#12430

Deleting all text from General Collateral description (wysiwyg) does not clear HTML styling tags, which in turn displays empty Collateral Summary

*Description of changes:*
- Fix the issue by stripping out HTML from the GC description and checking for valid text
- Add tests

### DIFF
--- a/ppr-ui/src/components/collateral/Collateral.vue
+++ b/ppr-ui/src/components/collateral/Collateral.vue
@@ -84,7 +84,8 @@ import {
   onMounted,
   reactive,
   toRefs,
-  watch
+  watch,
+  onUnmounted
 } from '@vue/composition-api'
 import { useActions, useGetters } from 'vuex-composition-helpers'
 // local components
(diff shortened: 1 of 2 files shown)


ppr-ui/src/components/collateral/Collateral.vue
```diff
@@ -192,6 +193,18 @@ export default defineComponent({
...
+      if (getAddCollateral.value.generalCollateral[0]?.description?.replace(/(<([^>]+)>)/ig, '').length === 0) {
+        setGeneralCollateral([])
+        // clear collateral step check mark if there are no vehicles
+        // (this resets check mark that was set by general collateral description)
+        if (getAddCollateral.value.vehicleCollateral.length === 0) {
+          getAddCollateral.value.valid = false
+        }
+      }
+    })
+
```
cameron-eyds, line 207: Very cool, just curious, perhaps you already tried this as well but what if we made this conditional a computed property(so a reactive boolean in this case) and we set up a watcher on that property to clear the store?

Also, with regards to the `getAddCollateral.value.valid` is this computed currently based on have some char length or set manually by ui?

cameron-eyds: I'm just a little wary of using unmounted lifecycle hook in a SPA, although i'm sure it's probably fine here..

dimak1: I tried to add a condition [here](https://github.com/bcgov/ppr/blob/cf72314e0ad3a3278ce8f7dc77bb72d26a12f3cd/ppr-ui/src/components/collateral/generalCollateral/GenColEdit.vue#L146) in a watcher to check if the stripped value had any leftover text but it was lagging (delay on each keystroke). So then I wanted to check only once as the user navigates away from that route. 
`Valid` is set manually.

cameron-eyds: Ok, fair enough, thanks!

cameron-eyds (approved):
Looks good 👍

doug-lovett (approved):

PR #818, opened 2022-06-08
Use Subscribed Products instead of Roles for MHR/PPR

dimak1:
*Issue #:* /bcgov/entity#12436

*Description of changes:*
- Refactor to use Products instead of Roles
- Add new and update existing tests
- Cleanup tech debt


dimak1:
There could be a few improvements, like adding a state prop for active Subscribed Products codes only (i.e. ['PPR', 'MHR']). But wanted to send it off for initial review. 
@cameron-freshworks @doug-lovett

dimak1:
@cameron-freshworks @doug-lovett marked as ready for review.

ppr-ui/src/utils/products-helper.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @returns String to describe user's roles and products. e.g. STAFF_PPR
+ */
+export function getRoleProductCode (
+  authRoles: Array<string>,
+  subscribedProducts: Array<UserProductSubscriptionIF>): string {
+  let role: string
+
+  if (authRoles.includes('staff') || authRoles.includes('sbc') ||
+    authRoles.includes('helpdesk') || authRoles.includes('ppr_staff') ||
+    authRoles.includes('mhr_staff')) role = 'STAFF'
```
cameron-eyds: I believe this role is now obsolete (`mhr_staff`)

ppr-ui/src/views/Dashboard.vue
```diff
@@ -431,6 +431,20 @@ export default class Dashboard extends Vue {
...
+      .includes(ProductCode.PPR)
+  }
+
+  private get hasMHRProduct (): boolean {
+    return this.getUserProductSubscriptions
+      .filter(product => product.subscriptionStatus === ProductStatus.ACTIVE)
+      .map(product => product.code)
+      .includes(ProductCode.MHR)
+  }
+
```
cameron-eyds: Do you think these would be useful globally?

doug-lovett (approved):
Scanned the change, looks good.

cameron-eyds (commented):
Looks good so far! 

Just a few minor questions..
Do we still consider purely the roles for this text templating? As Staff will not have Products.
Also is the intention here to use roles ONLY for the text templating specifically? 

ie we use the roles (+ feature flag) for feature accessibility, the ability to do MHR Searches or start MHR Registrations, for clients we will need to consider products + roles and staff only roles, is that the idea?

Maybe we should have a quick discussion, just hmu whenever 👍

ppr-ui/src/utils/products-helper.ts
dimak1: Removed.

ppr-ui/src/views/Dashboard.vue
dimak1: Added active Product Codes to the store.

ppr-ui/src/App.vue
```diff
@@ -401,6 +405,18 @@ export default class App extends Mixins(AuthMixin) {
...
+      const subscribedProducts = await fetchAccountProducts(this.getAccountId)
+      if (subscribedProducts) {
+        this.setUserProductSubscriptions(subscribedProducts)
+
+        const activeProductCodes = subscribedProducts
+          .filter(product => product.subscriptionStatus === ProductStatus.ACTIVE)
+          .map(product => product.code)
+        this.setUserProductSubscriptionsCodes(activeProductCodes)
+      } else {
+        throw new Error('Unable to get Products for the User')
```
cameron-eyds: Will this always throw for staff?

ppr-ui/src/store/actions/actions-model.ts
```diff
@@ -44,6 +44,14 @@ export const setAccountProductSubscribtion: ActionIF = (
...
 }
 
+export const setUserProductSubscriptions: ActionIF = ({ commit }, products: Array<UserProductSubscriptionIF>): void => {
+  commit('mutateUserProductSubscriptions', products)
+}
+
+export const setUserProductSubscriptionsCodes: ActionIF = ({ commit }, activeProducts: Array<ProductCode>): void => {
+  commit('mutateUserProductSubscriptionsCodes', activeProducts)
+}
+
```
cameron-eyds, line 54: Is the latter the abbreviated version of the products? For easier access?

cameron-eyds (commented):
Looks good, nice job! 

Just curious, in the event of a staff user (ie no account id or products) that they still have access based on their token roles? Is that still the case?

ppr-ui/src/store/actions/actions-model.ts
dimak1: Yes, correct. The latter is will hold only the Active Product Codes (for easier access) vs. All Subscribed Products (Active+Inactive).

dimak1:
@cameron-freshworks thanks for the review! I missed that part when Staff will not have an Account Id (it wasn't mentioned). I will update it accordingly and push a new commit.

ppr-ui/src/App.vue
dimak1: Will need to update it to take into consideration that Staff might not have an account id.

cameron-eyds: Yessir good call, i don't believe they do. They have `BCOL_STAFF_PARTY_CODE` afaik, SBC staff might have an account, not 100% sure. 
But yes, you are correct, we will need to consider staff in these conditions.

ppr-ui/src/store/actions/actions-model.ts
cameron-eyds: Very cool, thanks!

cameron-eyds:
> @cameron-freshworks thanks for the review! I missed that part when Staff will not have an Account Id (it wasn't mentioned). I will update it accordingly and push a new commit.

No prob, right on!

ppr-ui/src/App.vue
dimak1: Added a check for `accountId`. If a user doesn't have an id (ie staff) then don't fetch the products.

cameron-eyds: Excellent 👍

cameron-eyds (approved):
Looks good! 
Due to our pre-release Dev code freeze, hold off on a merge until Divya gives the green light (Thurs i believe)

doug-lovett (approved):
That's a lot of work, thanks Dima and Cameron for the feedback. I think I followed it.

PR #828, opened 2022-06-14
Fix for MH vehicle collateral schema: allow MHR and serial numbers.

doug-lovett:

*Issue #:* /bcgov/entity#NA

*Description of changes:*
- Fix to recent PPR vehicle collateral MH type schema change: allow both MHR number and serial number.


cameron-eyds (approved):
Looks good!

PR #833, opened 2022-06-16
MHR Registration - Other Information

dimak1:
*Issue #:* /bcgov/entity#12449

*Description of changes:*

- Add the **Other Information** section for the 'Describe Your Home' step
- Add MHR Validation composable
- Show Other Info section on the Review step
- Add unit tests
- Also adding 'ManufacturerMakeModel.vue' as part of this PR because I added this prior and didn't want to remove it (this is WIP still).

**Screenshots**

Screen Shot 2022-06-21 at 5 29 35 PM

Review page 
Screen Shot 2022-06-21 at 5 29 46 PM

Review page (when nothing was entered)
Screen Shot 2022-06-21 at 5 29 00 PM


ppr-ui/src/components/mhr/index.ts
```diff
@@ -0,0 +1 @@
+
```
cameron-eyds: I think this folder/index might be obsolete now. 
We can add it back in, if we find there is some common components between Mhr Registration and Search perhaps but no purpose at the moment.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+  <div>
+    <v-card flat class="white pb-6 pt-6 pr-10 pl-8 rounded">
```
cameron-eyds: Are all these helpers required?

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
```diff
@@ -0,0 +1,133 @@
...
+      () => localState.manufacturerName,
+      (val: string) => {
+        setMhrRegistrationManufacturerName(val)
+      }
+    )
+
+    watch(
+      () => localState.yearOfManufacture,
+      (val: string) => {
+        setMhrRegistrationYearOfManufacture(val)
```
cameron-eyds: I believe this property is a `number`, according to the specs.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
```diff
@@ -0,0 +1,133 @@
...
+        setMhrRegistrationHomeMake(val)
+      }
+    )
+
+    watch(
+      () => localState.model,
+      (val: string) => {
+        setMhrRegistrationHomeModel(val)
+      }
+    )
```
cameron-eyds: It's a cool way to do this, seeing how it's not a form with standard submission. 

That being said, correct me if i'm wrong, but this will apply the values to store immediately regardless of validations no?
Is there field level validations or is the plane to intercept here prior to setting properties to store if `invalid`?

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <div>
```
cameron-eyds: Never a bad idea to add an `id` to your components. 
Sometimes useful for unit testing/automation and/or scroll to effects that we will include with our future app-wide validations.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
```diff
@@ -0,0 +1,69 @@
...
+    <v-card flat class="white pb-6 pt-6 pr-10 pl-8 rounded">
```
cameron-eyds: `pb` and `pt` can be shortened to `py` if the px count is the same.
`py` and `px` is padding top/bottom, left/right respectively.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
cameron-eyds: Can be shortened to `py` here as well

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+  <div>
```
cameron-eyds, line 2: Not a bad idea to add an id :)

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
```diff
@@ -0,0 +1,69 @@
...
+      <v-row>
+        <v-col cols="2">
+          <label class="generic-label ml" for="">Other</label>
```
cameron-eyds: What is `ml` short for. Is that a class defined locally or a helper?

Can probably remove `for=""` if its not required

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
```diff
@@ -0,0 +1,69 @@
...
+        </v-col>
+        <v-col cols="10">
+          <v-textarea
+            v-model="otherRemarks"
+            filled
+            :error-messages="errorMessages.otherRemarks"
+            name="name"
+            counter="140"
+            placeholder="Enter other details about the home (Optional)"
```
cameron-eyds: Designs look like they call for a `hint` with a `persistent-hint` property.
I could be wrong but i believe placeholder is gone once you start typing in the input?

ppr-ui/src/utils/validators/mhr-registration-validation.ts
```diff
@@ -0,0 +1,16 @@
...
+      id: 'otherRemarks',
+      validations: (fieldValue: string) => {
+        // All validations would go here, returning the error for each rule
+        if (fieldValue?.length > 140) {
+          return 'Maximum 140 characters'
+        }
+      }
+    }
+  ] as Array<MhrRegistrationFieldValidationIF>
+}
```
cameron-eyds: I appreciate the approach here but this might add unneeded complexity in my opinion. 

Vuetify components that take input have `:rules` prop which can handle the abstract complexity of applying rules and entering an error state/message on the component. 

I believe this could be simplified with a generic input rule in our `useInputRules` composable, something like..
  
Screen Shot 2022-06-21 at 9 10 27 PM

  And just apply the rule to the input `:rules="textAreaRules"`
  
Could probably be more accurately named but yea, i'm not sure we need this abstract method of fetching validations when it's built into vuetify components. 

Thoughts on this?

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
```diff
@@ -0,0 +1,69 @@
...
+    <v-card flat class="white pb-6 pt-6 pr-10 pl-8 rounded">
+      <v-row>
+        <v-col cols="2">
+          <label class="generic-label ml" for="">Other</label>
+        </v-col>
+        <v-col cols="10">
+          <v-textarea
+            v-model="otherRemarks"
+            filled
+            :error-messages="errorMessages.otherRemarks"
```
cameron-eyds: This is a manual approach to applying an error message, see comments below about utilizing `:rules` prop perhaps.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
```diff
@@ -0,0 +1,69 @@
...
+
+    const { setMhrRegistrationOtherInfo } = useActions<any>([
+      'setMhrRegistrationOtherInfo'
+    ])
+
+    const localState = reactive({
+      otherRemarks: getMhrRegistrationOtherInfo.value,
+      errorMessages: {
+        otherRemarks: ''
+      }
```
cameron-eyds: I'm not sure we need a local model of the error message when the component can handle this internally.
We would need a local model for determining validity however. 

Wrapping our input fields in a `v-form` is very useful for a few reasons, we can utilize a  `v-model="isValid" `property which will be our local model for determining the validity state of an input or collection of inputs. 
We can also assign a `ref` in order to prompt validations when we choose too, if requirements call for it at a specific time (ie at File and Pay or something)

ppr-ui/src/store/getters/state-getters.ts
```diff
@@ -610,3 +610,23 @@ export const hasMorePages = (state: StateIF): boolean => {
...
+export const getMhrRegistrationHomeMake = (state: StateIF): string => {
+  return state.stateModel.mhrRegistration.description.baseInformation.make
+}
+
+export const getMhrRegistrationHomeModel = (state: StateIF): string => {
+  return state.stateModel.mhrRegistration.description.baseInformation.model
+}
+
+export const getMhrRegistrationOtherInfo = (state: StateIF): string => {
+  return state.stateModel.mhrRegistration.description.otherRemarks
```
cameron-eyds, line 631: I didn't see `otherRemarks` or `other` in the spec, is this something Doug is still working on perhaps?

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
```diff
@@ -44,10 +44,28 @@
...
+              <h3>Other Information</h3>
+            </v-col>
+            <v-col cols="9">
+              <div>
+                <p>
+                  {{ getMhrRegistrationOtherInfo || '(Not Entered)' }}
+                </p>
+              </div>
+            </v-col>
+          </v-row>
```
cameron-eyds: Nice, i like the style for readability. 
You could probably put the `<p>` tag on one line?
Also is the `<div>` wrapper required for the p tag?

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
```diff
@@ -123,6 +144,18 @@ export default class MhrReviewConfirm extends Vue {
...
 
+#mhr-registration-summary {
+  h3 {
+    line-height: unset;
+  }
+
+  p {
+    margin-bottom: unset;
+    line-height: 24px;
+    color: #495057;
```
cameron-eyds: Same as `$gray7` i think

cameron-eyds (commented):
Nice work, a lot of great stuff here, just a few questions and comments.
I believe we could simplify our validations a bit, hit me up if you would like to discuss 👍

ppr-ui/src/components/mhr/index.ts
dimak1: Agree, removed it.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: I got them from another card but didn't shorten them. Done!

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: Good catch, updated.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: Right now, we are storing values. We can either validate them prior or use stored values for validation. We also discussed in one of the meetings that they wanted to save a draft with (possibly) invalid values.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
dimak1: Added an `id`.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
dimak1: Shortened.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: Added ids for each row.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
dimak1: Cleaned up. Added for attribute `for` attributes for labels.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
dimak1: To me, the design looks like a placeholder. It will be gone after the user starts typing.
Screen Shot 2022-06-22 at 8 26 40 AM

ppr-ui/src/store/getters/state-getters.ts
dimak1: Yea, I believe he is still working on it. He told me he will add them when I pointed out it was missing.

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
dimak1: Correct, updated!

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
dimak1: We can probably omit `div` tag. I think it was prettier to format like that.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
cameron-eyds: Screen Shot 2022-06-22 at 8 30 24 AM

Confirm with Tracey, this comp looks to retain the 'hint'

ppr-ui/src/utils/validators/mhr-registration-validation.ts
dimak1: There are multiple ways we do validation in this app. I will look into vuetify's built-in validation. Thanks!

cameron-eyds: Thank you.
To your point let's do our best to keep it simple, adding in another way to validate is probably not the solution to the inconsistency.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
cameron-eyds: Will leave this comment unresolved but the assumption is that validations will be applied when you continue work on this component.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
cameron-eyds: Sorry i don't follow. Is the `for` prop being used?

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
cameron-eyds: Sounds good, i would remove unless it had a purpose (ie styling etc)

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: I'm adding vuetify :rules to the fields, and it works great. We can keep it open till then, thanks.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
dimak1: It is normally used for accessibility and better user experience, when clicking on the label cursor jumps to the respective input field.

ppr-ui/src/utils/validators/mhr-registration-validation.ts
dimak1: Agree, I like the built-it validation and we should go with that, yea.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
dimak1: Yes, will rework it.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
cameron-eyds: Oh interesting, i did not know that!
Even without a value provided to the prop? (ie an id)
`for=""` alone does this? How does it know what input to jump too 🤔

dimak1: `for` attr should have/point to an id of the input field. I've added them for each field.

ppr-ui/src/components/mhrRegistration/YourHome/OtherInformation.vue
dimak1: Tracey confirmed the text should stay. Will update it.

dimak1:
@cameron-freshworks updated the validation for the Other Information component. For now, I just added a `isOtherInfoValid` flag to v-form, but haven't used it yet. Need to figure out a way to use it properly. There is a way to prevent storing invalid values in the store model but in other forms don't do it and instead have an invalid flag.

ppr-ui/src/composables/useInputRules.ts
```diff
@@ -32,10 +31,15 @@ export const useInputRules = () => {
     ]
   }
 
+  const maxLengthRules = (maxLength: number): Array<Function> => {
+    return [v => v.length <= maxLength || `Maximum ${maxLength} characters`]
+  }
+
```
cameron-eyds, line 37: ⭐

cameron-eyds (approved):
Much much cleaner, very nice. Great job 👍

cameron-eyds:
> @cameron-freshworks updated the validation for the Other Information component. For now, I just added a `isOtherInfoValid` flag to v-form, but haven't used it yet. Need to figure out a way to use it properly. There is a way to prevent storing invalid values in the store model but in other forms don't do it and instead have an invalid flag.

Definitely! This works for now. We will utilize this flag when we tackle app validations, great job 👍

dimak1:
@doug-lovett @garyfan1 please review.

doug-lovett (approved):
Looks good. Thanks for all your input Cameron.

PR #835, opened 2022-06-16
12139 Serial Number too long bug fix

garyfan1:
*Issue #:* /bcgov/entity#12139

*Description of changes:*
fix bug where MH field is showing serial number error

### DIFF
--- a/ppr-ui/src/components/collateral/vehicleCollateral/factories/useCollateralValidation.ts
+++ b/ppr-ui/src/components/collateral/vehicleCollateral/factories/useCollateralValidation.ts
@@ -39,12 +39,8 @@ export const useCollateralValidation = () => {
   const validateCollateralForm = async currentVehicle => {
     const validationResult = await formValidation.validateForm(currentVehicle)
     errors.value = { ...errors.value, ...validationResult.fieldErrors }
-    if ((currentVehicle.type === 'MH') && (errors.value.manufacturedHomeRegistrationNumber.succeeded)) {
-      errors.value.manufacturedHomeRegistrationNumber =
-        validationResult.recordErrors.serialNumber
-    } else {
-      errors.value.serialNumber = validationResult.recordErrors.serialNumber
-    }
+    errors.value.serialNumber = validationResult.recordErrors.serialNumber
+
     return validationResult.succeeded
   }
 


ppr-ui/src/components/collateral/vehicleCollateral/factories/useCollateralValidation.ts
```diff
@@ -39,12 +39,8 @@ export const useCollateralValidation = () => {
...
     const validationResult = await formValidation.validateForm(currentVehicle)
     errors.value = { ...errors.value, ...validationResult.fieldErrors }
-    if ((currentVehicle.type === 'MH') && (errors.value.manufacturedHomeRegistrationNumber.succeeded)) {
-      errors.value.manufacturedHomeRegistrationNumber =
-        validationResult.recordErrors.serialNumber
-    } else {
-      errors.value.serialNumber = validationResult.recordErrors.serialNumber
-    }
+    errors.value.serialNumber = validationResult.recordErrors.serialNumber
+
```
cameron-eyds, line 43: Could you provide a little more detail on this in the PR.

The intention is to remove this condition entirely? Was it a condition you added previously that didn't work as expected?

cameron-eyds: Just to clarify, the serial number field was displaying under the wrong field? 
If so, does the condition need to be moved elsewhere?

garyfan1: The bug was mhr number field showing serial number field's error, caused by line 43-44, where serial number error overwrites mhr number error message. So I removed the whole condition statement,  but the test won't pass, so I added back the else statement for the tests to pass.

garyfan1: oh we need to keep the line that was in the else statement, is because serial number only appears in record check, so when we are adding only the field error to our error object (line 41), we'll miss out serial number's error; and so we need to explicitly add it.

cameron-eyds: Sounds good, thanks for digging into this a bit more!

cameron-eyds (approved):

doug-lovett (approved):

dimak1 (approved):

PR #837, opened 2022-06-16
Fix Search Bar header for Staff

dimak1:
*Description of changes:*

- Check `mhr-ui-enabled` flag for the Search Bar header

### DIFF
--- a/ppr-ui/src/views/Dashboard.vue
+++ b/ppr-ui/src/views/Dashboard.vue
@@ -30,9 +30,10 @@
                   id="search-header"
                   :class="[$style['dashboard-title'], 'pl-6', 'pt-3', 'pb-3', 'soft-corners-top']">
             <v-col cols="auto">
-              <b v-if="hasPPRProduct && hasMHRProduct">Manufactured Home and Personal Property Registry Search</b>
-              <b v-else-if="hasPPRProduct">Personal Property Search</b>
-              <b v-else-if="hasMHRProduct">Manufactured Home Search</b>
+              <b v-if="hasPPR && hasMHR">
+                Manufactured Home and Personal Property Registry Search</b>
+              <b v-else-if="hasPPR">Personal Property Search</b>
+              <b v-else-if="hasMHR">Manufactured Home Search</b>
             </v-col>
           </v-row>
           <v-row no-gutters>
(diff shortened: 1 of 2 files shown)


ppr-ui/src/views/Dashboard.vue
```diff
@@ -30,9 +30,10 @@
                   id="search-header"
                   :class="[$style['dashboard-title'], 'pl-6', 'pt-3', 'pb-3', 'soft-corners-top']">
             <v-col cols="auto">
-              <b v-if="hasPPRProduct && hasMHRProduct">Manufactured Home and Personal Property Registry Search</b>
-              <b v-else-if="hasPPRProduct">Personal Property Search</b>
-              <b v-else-if="hasMHRProduct">Manufactured Home Search</b>
+              <b v-if="hasPPR && hasMHR">
+                Manufactured Home and Personal Property Registry Search</b>
```
cameron-eyds, line 34: too many chars for 1 line?

ppr-ui/src/views/Dashboard.vue
```diff
@@ -431,12 +437,22 @@ export default class Dashboard extends Vue {
     return this.getSearchHistory?.length || 0
   }
 
-  private get hasPPRProduct (): boolean {
-    return this.getUserProductSubscriptionsCodes.includes(ProductCode.PPR)
+  private get hasPPR (): boolean {
+    // For Staff we check roles, for Client we check Products
+    if (this.isRoleStaff || this.isRoleStaffBcol || this.isRoleStaffReg) {
```
cameron-eyds, line 442: For a consolidated staff check like this, is there already a getter? If not, perhaps we could create one.
Seems useful, saves some imports and duplication here if created in the store.

cameron-eyds (approved):
Looks good, just 1 suggestion!

doug-lovett (approved):

ppr-ui/src/views/Dashboard.vue
dimak1: Yea, had to break it

ppr-ui/src/views/Dashboard.vue
dimak1: There is no generic getter to check for all staff options. I could add one, as we use it in multiple places.

dimak1: I will add this getter in another PR. Just want to keep this one as a small fix.

PR #839, opened 2022-06-21
Home Sections Unit Testing

cameron-eyds:
*Issue #:* /bcgov/entity#12532

*Description of changes:*
* Implements some additional unit testing for Home Sections


dimak1 (approved):
Looks good. I'm not used to testing the content, but since it is not going to change often, we should be fine. Approved.

cameron-eyds:
> Looks good. I'm not used to testing the content, but since it is not going to change often, we should be fine. Approved.

Totally agree. 
The value might come into play if future requirements dictates some variation in that content. If we end up templating different values we can slot that in dynamically, in the tests.

doug-lovett (approved):
Looks good.

PR #840, opened 2022-06-21
Search fails when the search parameter contains special characters

garyfan1:
*Issue #:* /bcgov/entity#12171

*Description of changes:*
fix astrophe can't be recognized, added some tests

### DIFF
--- a/ppr-ui/src/components/search/SearchBar.vue
+++ b/ppr-ui/src/components/search/SearchBar.vue
@@ -393,11 +393,22 @@ export default defineComponent({
       })
     })
 
+    const cleanUpInput = (dirtyValue: string|undefined) => {
+      if (dirtyValue === undefined) {
+        return undefined
+      }
+      return dirtyValue
+        .trim()
+        .replaceAll(/[\u200B-\u200D\uFEFF\u200E\u200F]|(?:&#x200E;)/g, '')
+        .replaceAll(/[\u2018\u2019]/g, "'")
+        .replaceAll(/[\u201C\u201D]/g, '"')
+    }
+
     const getCriteria = () => {
       if (localState.isIndividual) {
-        const first = localState.searchValueFirst?.trim()
-        const second = localState.searchValueSecond?.trim()
-        const last = localState.searchValueLast?.trim()
+        const first = cleanUpInput(localState.searchValueFirst)
+        const second = cleanUpInput(localState.searchValueSecond)
+        const last = cleanUpInput(localState.searchValueLast)
 
         if (isPPRSearchType(localState.selectedSearchType.searchTypeAPI)) {
           return { debtorName: { first: first, second: second, last: last } }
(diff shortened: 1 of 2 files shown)


ppr-ui/src/components/search/SearchBar.vue
```diff
@@ -393,11 +393,22 @@ export default defineComponent({
...
+      if (dirtyValue === undefined) {
+        return undefined
+      }
+      return dirtyValue
+        .trim()
+        .replaceAll(/[\u200B-\u200D\uFEFF\u200E\u200F]|(?:&#x200E;)/g, '')
+        .replaceAll(/[\u2018\u2019]/g, "'")
+        .replaceAll(/[\u201C\u201D]/g, '"')
+    }
+
```
cameron-eyds, line 412: Judging by the regex this will clean hidden characters as well?

Could you add a docblock comment to describe the function definition of what it will clean for those unfamiliar with reading regex?

dimak1 (approved):

ppr-ui/src/components/search/SearchBar.vue
```diff
@@ -393,11 +393,22 @@ export default defineComponent({
       })
     })
 
+    const cleanUpInput = (dirtyValue: string|undefined) => {
+      if (dirtyValue === undefined) {
```
cameron-eyds, line 403: Could this be simplified? ie `(!dirtyValue)`
Checking for undefined is pretty explicit. What is the argument was `null`?

ppr-ui/src/components/search/SearchBar.vue
```diff
@@ -393,11 +393,22 @@ export default defineComponent({
...
+    }
+
     const getCriteria = () => {
       if (localState.isIndividual) {
-        const first = localState.searchValueFirst?.trim()
-        const second = localState.searchValueSecond?.trim()
-        const last = localState.searchValueLast?.trim()
+        const first = cleanUpInput(localState.searchValueFirst)
+        const second = cleanUpInput(localState.searchValueSecond)
+        const last = cleanUpInput(localState.searchValueLast)
```
cameron-eyds, line 417: If the function returns `undefined`for any of these properties as it could, do we still want to return the search criteria with `undefined` properties?

ppr-ui/src/components/search/SearchBar.vue
garyfan1: Because the original expression was `localstate.searchValueFirst?.trim()`, which if `localstate.searchValueFirst` is `undefined`, it will short-circuit and return `undefined`. So in order to retain the original behavior, I chose to check for `undefine`.
And because these parameters are returned by other optional chaining, it'll only be either `string` or `undefined`.

ppr-ui/src/components/search/SearchBar.vue
garyfan1: I believe that is the original behavior.
If it works don't touch it?
I'm only making sure `cleanUpInput` is behaving just like its original expression, just with a bit more of string manipulation.

ppr-ui/src/components/search/SearchBar.vue
cameron-eyds: Ok, fair enough! Thanks for the explanation 👍

ppr-ui/src/components/search/SearchBar.vue
cameron-eyds: haha fair enough. You are right! (As long as it worked to begin with 😆 )
Resolving.

ppr-ui/src/components/search/SearchBar.vue
garyfan1: just added!

cameron-eyds: Thanks sir 👍

cameron-eyds: Resolved

cameron-eyds (approved):

doug-lovett (approved):
Nice work.

dimak1 (approved):
Looks good, thanks for clarifications @garyfan1.

PR #842, opened 2022-06-22
12582 mhr callback url

garyfan1:
*Issue #:* /bcgov/entity#12582

*Description of changes:*
add callbackURL for submitSelectedMhr

### DIFF
--- a/ppr-ui/src/utils/mhr-api-helper.ts
+++ b/ppr-ui/src/utils/mhr-api-helper.ts
@@ -100,8 +100,7 @@ export async function submitSelectedMhr (
   searchId: string,
   selected: Array<ManufacturedHomeSearchResultIF>,
   staffPayment: StaffPaymentIF = null,
-  isCertified: boolean = false,
-  shouldCallback: boolean = false
+  isCertified: boolean = false
 ): Promise<number> {
   const url = sessionStorage.getItem('MHR_API_URL')
   // change to application/pdf to get the pdf right away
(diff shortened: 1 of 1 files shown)


doug-lovett (approved):
Looks good.

dimak1 (approved):

ppr-ui/src/utils/mhr-api-helper.ts
```diff
@@ -122,6 +123,16 @@ export async function submitSelectedMhr (
     }
   }
 
+  if (shouldCallback || selected.length >= 75) {
```
cameron-eyds: If we are putting the logic here to enforce the callback based on arr length, do we need the argument to do it manually?

cameron-eyds (approved):
Looks good! 
I'm not against an override argument although if you don't need it now, typically would add it in when you do need it :) 
Less is more right! haha

ppr-ui/src/utils/mhr-api-helper.ts
garyfan1: I leave it there because callback is optional for sizes that are less than 75?

garyfan1:
oh I see your point; I don't need to add things for the future. I can definitely remove those if we really don't want to keep it.

cameron-eyds:
> oh I see your point; I don't need to add things for the future. I can definitely remove those if we really don't want to keep it.

For sure, i would only add if we needed it now, thanks 👍 
If you were building a shared component that was going to accessible for many implementations, it would make sense to include options up front.

cameron-eyds (approved):

PR #847, opened 2022-06-29
Fix amending empty General Collateral

dimak1:
*Issue #:* /bcgov/entity#12130

*Description of changes:*

- Fix amending empty General Collateral (remove HTML tags)
- Remove empty General Collateral (during amend so we don't keep empty ones and can't Undo)

### DIFF
--- a/ppr-ui/src/components/collateral/generalCollateral/GenColAmend.vue
+++ b/ppr-ui/src/components/collateral/generalCollateral/GenColAmend.vue
@@ -207,6 +207,9 @@ export default defineComponent({
         } else {
           newGeneralCollateral.push(amendedGC)
         }
+        if (localState.addDesc === '' && localState.delDesc === '') { // remove empty collateral altogether
+          newGeneralCollateral.pop()
+        }
         setGeneralCollateral(newGeneralCollateral)
       }
       emit('closeGenColAmend', true)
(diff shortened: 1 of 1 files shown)


ppr-ui/src/components/collateral/generalCollateral/GenColAmend.vue
```diff
@@ -207,6 +207,9 @@ export default defineComponent({
         } else {
           newGeneralCollateral.push(amendedGC)
         }
+        if (localState.addDesc === '' && localState.delDesc === '') { // remove empty collateral altogether
+          newGeneralCollateral.pop()
```
cameron-eyds, line 211: What are we removing here from `newGeneralCollateral`?
ie what is the last element in the array? How does this work?

dimak1: Yes, we are removing the last added Gen Col, if it has empty values.

In the logic above, we always push a new GC, even with empty values. And since there are multiple items in `generalCollateral` array, it allows to Undo (which we don't want for empty GC). So the logic here is to check if both Add and Del are empty, we simply remove the newly added GC.

Screen Shot 2022-06-29 at 2 16 54 PM

dimak1:
@cameron-freshworks just fyi, we shared the screen with Divya and it worked as desired.

ppr-ui/src/components/collateral/generalCollateral/GenColAmend.vue
cameron-eyds: Gotcha! Cool, thanks for the explanation sir 👍

cameron-eyds (approved):
Looks good!

doug-lovett (approved):

PR #851, opened 2022-06-30
Other Information and Review page updates (UXA feedback)

dimak1:
*Issue #:* /bcgov/entity#12535

*Description of changes:*

- Updates as per UXA feedback
- Trim empty spaces
- Preserve styling when displaying Other information on the Review page (preserve new line chars)

Screen Shot 2022-06-29 at 5 50 57 PM


ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
```diff
@@ -101,7 +101,7 @@
               <h3>Other Information</h3>
             </v-col>
             <v-col cols="9">
-              <p>{{ getMhrRegistrationOtherInfo || '(Not Entered)' }}</p>
+              <p v-html="getMhrRegistrationOtherInfo.replace(/\r?\n/g, '<br />') || '(Not Entered)'"></p>
```
cameron-eyds: Try `.trim` on the component i.e. ` v-model.trim="otherRemarks"`

If this is the result of a different set of ux feedback, please create a new computed property so we can keep the regex out of the template. 

Also, what is the purpose here? This is purely for review because this formatting wouldn't have any impact on whats sent in the filing, or be reflective in the outputs, if that was the designs intention.

ppr-ui/src/composables/useInputRules.ts
```diff
@@ -6,6 +6,13 @@ export const useInputRules = () => {
     ]
   }
 
+  // check if entire value has empty spaces
+  const isEmpty = (): Array<Function> => {
+    return [
+      v => !/^\s+$/g.test(v) || 'Invalid spaces'
+    ]
+  }
+
```
cameron-eyds, line 21: I believe Tracey posted she preferred to not have validations? I could be wrong but that was last i heard.

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
dimak1: It works nicely without the trim. 
We need to display the HTML on the review page (with new lines, etc.) so I had to convert new line chars to `<br>`. Initially, I had `<br>`'s in the store model (replacing \n with `<br>` as user types) but then changed to show on the review page. 
But maybe I need to update to have <br> in the model for the filing output...

ppr-ui/src/composables/useInputRules.ts
dimak1: Yea, I saw it after and wanted to rework it today.

dimak1:
Changing to draft to rework some of the logic (as per UXA feedback).

ppr-ui/src/composables/useInputRules.ts
dimak1: Removed validation for empty spaces.

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
dimak1: Reworked to add HTML formatted string to the store model (for filing outputs and Review page template).

ppr-ui/src/utils/format-helper.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Helper function to convert string to HTML format
+ * @param stringToFormat
+ */
+export function toHtml (stringToFormat: string) {
+  return stringToFormat.replace(/\r?\n/g, '<br />') // convert new line chars to html break lines
```
cameron-eyds: is `<br />` the same as `<br>` 
Im not sure it needs to be self closing

cameron-eyds (commented):
Looks fine

A little concerned about passing HTML to the back end though, this could be a security concern. 
Also, do we want to set the precedent where we want to preserve everything a user can copy paste into the documents? 
Is this a business requirement to allow the components to preserve hidden chars/spacing etc for these text input areas. 
We might need to discuss this as a team, thoughts on this?

dimak1:
> Looks fine
> 
> A little concerned about passing HTML to the back end though, this could be a security concern. Also, do we want to set the precedent where we want to preserve everything a user can copy paste into the documents? Is this a business requirement to allow the components to preserve hidden chars/spacing etc for these text input areas. We might need to discuss this as a team, thoughts on this?

@cameron-freshworks 
As per UXA feedback, the ask is to preserve new lines. We are not really putting HTML in the backend. Only replacing new line `\n` with a `<br>` for filing output (so it is styled the same) and to display on the Review page. Maybe the function could have a better name.

ppr-ui/src/utils/format-helper.ts
dimak1: Updated as not self closing

dimak1:
@cameron-freshworks as per @tlebedovich I preserved new line characters only on the Review page (for styling) and do not add html to the store model.

dimak1:
@doug-lovett can you please review it?

doug-lovett (approved):

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
dimak1: Removed adding Html to the store. We might need a further discussion on this. For now, I will merge.

PR #852, opened 2022-06-30
Serial number search return duplicates as with legacy.

doug-lovett:

*Issue #:* /bcgov/entity#12626

*Description of changes:*
- Match the legacy application logic, returning all section serial number matches within a single registration as distinct results.


dimak1 (approved):

cameron-eyds (approved):
LGTM!

PR #861, opened 2022-07-05
upgrade sbc-common-component version, add its depending env configs

garyfan1:
*Issue #:* /bcgov/entity#12814

*Description of changes:*
upgrade sbc-common-component version, add its depending env configs

### DIFF
--- a/ppr-ui/src/utils/config-helper.ts
+++ b/ppr-ui/src/utils/config-helper.ts
@@ -136,4 +136,9 @@ export async function fetchConfig (): Promise<any> {
     (<any>window).addressCompleteKey = addressCompleteKey
     console.info('Set Address Complete Key.')
   }
+
+  const siteminderLogoutUrl: string = response.data.SITEMINDER_LOGOUT_URL
+  if (siteminderLogoutUrl) {
+    sessionStorage.setItem('SITEMINDER_LOGOUT_URL', siteminderLogoutUrl)
+  }
 }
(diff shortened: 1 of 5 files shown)


ppr-ui/src/utils/config-helper.ts
```diff
@@ -136,4 +136,9 @@ export async function fetchConfig (): Promise<any> {
     (<any>window).addressCompleteKey = addressCompleteKey
     console.info('Set Address Complete Key.')
   }
+
+  const siteminderLogoutUrl: string = response.data.SITEMINDER_LOGOUT_URL
+  if (siteminderLogoutUrl) {
+    sessionStorage.setItem('SITEMINDER_LOGOUT_URL', siteminderLogoutUrl)
```
dimak1, line 142: Is this used anywhere?

cameron-eyds: I was curious about this as well. 
Gary mentioned this was intentional as it's a dependency of recent changes in the SBC Common Components, although i'm not sure the details.

cameron-eyds (approved):
@garyfan1 Code looks good!

Could you call out the purpose of the siteminder inclusion in both the ticket/pr.
What it does, why we need etc, just so everyone is aware, thanks sir!

garyfan1:
We weren't logging out of siteminder when logging out, and this will cause an error if we try to log in again with another IDIR.
More details please see https://github.com/bcgov/entity/issues/12565.

cameron-eyds:
> We weren't logging out of siteminder when logging out, and this will cause an error if we try to log in again with another IDIR. More details please see [bcgov/entity#12565](https://github.com/bcgov/entity/issues/12565).

Excellent, thanks 👍

dimak1 (approved):
Thanks for clarification @garyfan1. Approved.

PR #862, opened 2022-07-06
Manufacturer Make Model section

dimak1:
*Issue #:* /bcgov/entity#12466

*Description of changes:*
- Manufacturer Make Model section
- Review Page 
- Validations
- Tests

Screenshots:

Screen Shot 2022-07-06 at 4 56 28 PM

Screen Shot 2022-07-06 at 4 57 45 PM

Screen Shot 2022-07-06 at 4 56 14 PM

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
+++ b/ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
@@ -93,15 +93,9 @@ export default defineComponent({
       'getMhrRegistrationHomeModel'
     ])
     const {
-      setMhrRegistrationManufacturerName,
-      setMhrRegistrationYearOfManufacture,
-      setMhrRegistrationHomeMake,
-      setMhrRegistrationHomeModel
+      setMhrHomeRegistration
     } = useActions<any>([
-      'setMhrRegistrationManufacturerName',
-      'setMhrRegistrationYearOfManufacture',
-      'setMhrRegistrationHomeMake',
-      'setMhrRegistrationHomeModel'
+      'setMhrHomeRegistration'
     ])
 
     const localState = reactive({
@@ -114,28 +108,28 @@ export default defineComponent({
     watch(
       () => localState.manufacturerName,
       (val: string) => {
-        setMhrRegistrationManufacturerName(val)
+        setMhrHomeRegistration({ key: 'description.manufacturer', value: val })
       }
     )
 
     watch(
       () => localState.yearOfManufacture,
       (val: number) => {
-        setMhrRegistrationYearOfManufacture(val)
+        setMhrHomeRegistration({ key: 'description.baseInformation.year', value: val })
       }
     )
 
     watch(
       () => localState.make,
       (val: string) => {
-        setMhrRegistrationHomeMake(val)
+        setMhrHomeRegistration({ key: 'description.baseInformation.make', value: val })
       }
     )
 
     watch(
       () => localState.model,
       (val: string) => {
-        setMhrRegistrationHomeModel(val)
+        setMhrHomeRegistration({ key: 'description.baseInformation.model', value: val })
       }
     )
 
--- a/ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
+++ b/ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
@@ -12,7 +12,9 @@
             id="manufacturer-name"
             v-model="manufacturerName"
             filled
+            :rules="maxLength(65)"
             label="Business Legal Name (Optional)"
+            data-test-id="manufacturer-name"
           />
         </v-col>
       </v-row>
@@ -28,17 +30,36 @@
             id="manufacturer-year"
             v-model="yearOfManufacture"
             filled
+            :rules="manufactureYearRules"
             label="Year of Manufacture"
             persistent-hint
             hint="YYYY"
+            data-test-id="manufacture-year"
           />
         </v-col>
         <v-col cols="6">
-          <v-checkbox>
-            <template v-slot:label>
-              This Year of Manufacture is approximate
+          <v-checkbox
+            id="circa-year"
+            label="This Year of Manufacture is approximate"
+            v-model="circa"
+            class="float-left"
+            data-test-id="circa-year-checkbox"
+          />
+           <v-tooltip
+            top
+            content-class="top-tooltip pa-5"
+            transition="fade-transition"
+            data-test-id="circa-year-tooltip"
+          >
+            <template v-slot:activator="{ on }">
+              <v-icon
+                class="circa-tooltip-icon ml-2 mt-n1"
+                color="primary"
+                v-on="on">mdi-information-outline</v-icon>
             </template>
-          </v-checkbox>
+            When the exact year of manufacture is unknown, enter an estimated
+            year and indicate that the year is approximate.
+          </v-tooltip>
         </v-col>
       </v-row>
 
@@ -53,7 +74,9 @@
             id="manufacturer-make"
             v-model="make"
             filled
+            :rules="makeRules"
             label="Make"
+            data-test-id="manufacturer-make"
           />
         </v-col>
       </v-row>
@@ -67,8 +90,12 @@
             id="manufacturer-model"
             v-model="model"
             filled
+            :rules="modelRules"
             label="Model"
-          />
+            counter="65"
+            data-test-id="manufacturer-model"
+          >
+          </v-text-field>
         </v-col>
       </v-row>
     </v-card>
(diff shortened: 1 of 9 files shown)


ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
```diff
@@ -67,83 +90,134 @@
...
+        graterThan(new Date().getFullYear() + 1, 'Year cannot be more than 1 year in the future'),
+        startsWith('Year must begin with 19 or 20', '19', '20')
+      )
+
+    const combinedMakeModelLengthRule = (localState): Array<Function> => {
+      return [
+        () => (0 || localState.model.length) + (0 || localState.make.length) <= 65 ||
+          'Make and Model combined cannot exceed 65 characters'
+      ]
+    }
```
cameron-eyds, line 151: Good job!
Does this work well?

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
```diff
@@ -67,83 +90,134 @@
...
     ])
 
+    const manufactureYearRules: Array<Function> =
+      customRules(
+        required('Enter a year of manufacture'),
+        minLength(4),
+        maxLength(4),
+        graterThan(new Date().getFullYear() + 1, 'Year cannot be more than 1 year in the future'),
+        startsWith('Year must begin with 19 or 20', '19', '20')
+      )
```
cameron-eyds: Could this be `computed` as well?
Just to be consistent with our other computed custom rule sets?

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
```diff
@@ -67,83 +90,134 @@
...
+      (val: boolean) => {
+        setMhrHomeRegistration({ key: 'description.baseInformation.circa', value: val })
       }
     )
 
     watch(
       () => localState.make,
       (val: string) => {
-        setMhrRegistrationHomeMake(val)
+        setMhrHomeRegistration({ key: 'description.baseInformation.make', value: val })
```
cameron-eyds: I think we already have  `setMhrHomeDescription` for `description` nested items.

I'm a little wary to create 1 setter that can be applicable the entirety of the registration, best practices lean towards a setter per layer and to avoid mutating nested complex properties etc. 

Is it possible to repurpose this generic `action` to be specific to the `baseInformation` property?

ppr-ui/src/composables/useInputRules.ts
```diff
@@ -2,7 +2,13 @@
 export const useInputRules = () => {
   const maxLength = (maxLength: number): Array<Function> => {
     return [
-      v => v.length <= maxLength || `Maximum ${maxLength} characters`
+      v => (v || '').length <= maxLength || `Maximum ${maxLength} characters`
```
cameron-eyds: What's the difference here? 
Doesn't count spaces?

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: Thanks, yea it works nicely, but it only validates its own fields. So for instance, if both errors are showing (65+ chars for Make and Model) and I remove characters in one field, it will clear the error only for that field. I think this is expected, as it validates only on keystrokes.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: Yea, sounds good, I will make a computed.

ppr-ui/src/composables/useInputRules.ts
```diff
@@ -42,6 +48,22 @@ export const useInputRules = () => {
     ]
   }
 
+  // Check if string starts with any of the search values
+  const startsWith = (errorMessage: string, ...searchValues: string[]): Array<Function> => {
```
cameron-eyds: Good use of Rest parameter 👍

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: Yea we do have a setter for `description` and I couldn't use it for `baseInformation` as it is nested. I will create a new setter for it.

ppr-ui/src/composables/useInputRules.ts
dimak1: Just preventing throwing a (console) error when `v` is null, so just setting to empty (as per example in the official docs).

ppr-ui/src/store/mutations/mutations-model.ts
```diff
@@ -353,22 +354,10 @@ export const mutateUnsavedChanges = (state: StateIF, unsavedChanges: Boolean) =>
...
-  state.stateModel.mhrRegistration.description.baseInformation.model = model
-}
-
 export const mutateMhrHomeDescription = (state: StateIF, { key, value }) => {
   state.stateModel.mhrRegistration.description[key] = value
 }
+
+export const mutateMhrRegistration = (state: StateIF, { key, value }) => {
+  set(state.stateModel.mhrRegistration, key, value)
+}
```
cameron-eyds: Would prefer to avoid this and create an action/mutation per property layer. 
It can help avoid complicated nested objects and is also in uniform with the other projects historical patterns. 

I'm sure it works and it's a good idea to look into, i think the downside of the current implementation is the means by which it's set. ie calling the action with `rootObject.secondLayer.thirdLayer.arrayLayer` as a string is a little complex. 
We should brainstorm a more consistent way to specify the property layer than we could move towards this.

cameron-eyds (commented):
Nice job! Some really interesting stuff here. 

Just some comments around the singular generic action/mutation.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
cameron-eyds: Right on, nice work!

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
cameron-eyds: Right on, i think you are onto something here though. We just need to figure out how to specify these layers consistently and we could save a whole lot of code. 
Great topic!

ppr-ui/src/composables/useInputRules.ts
cameron-eyds: Ok fair enough, nice work 👍

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: Updated to computed.

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
dimak1: Refactored to have a `baseInformation` setter.

ppr-ui/src/composables/useInputRules.ts
dimak1: I refactored this to be consistent and put the error message as the last parameter. Changed from rest param to an array. Looks cleaner.

ppr-ui/src/store/mutations/mutations-model.ts
dimak1: Refactored.

dimak1:
@cameron-freshworks pls take another quick look as I updated as per comments. Thanks!

ppr-ui/src/components/mhrRegistration/YourHome/ManufacturerMakeModel.vue
cameron-eyds: Right on 👍

ppr-ui/src/composables/useInputRules.ts
```diff
@@ -32,13 +38,30 @@ export const useInputRules = () => {
...
+  const isNumber = (numberType: string = null, maxDigits: number = null, maxValue: number = null): Array<Function> => {
     const maxDigitRule = new RegExp(`^\\d{1,${maxDigits}}$`)
 
     return [
-      v => (v ? /^\d+$/g.test(v) : true) || `${numberType} must be a valid whole number (cannot contain decimals)`,
+      v => ((v && numberType) ? /^\d+$/g.test(v) : true) || `${numberType} must be a valid whole number (cannot contain decimals)`,
       v => ((v && maxDigits) ? maxDigitRule.test(v) : true) || `Maximum ${maxDigits} characters`,
-      v => ((v && maxValue) ? v < maxValue : true) || `${numberType} must be less than ${maxValue}`
+      v => ((v && maxValue) ? v < maxValue : true) || `${numberType} must be less than ${maxValue}`,
+      v => (v ? /^\d+$/g.test(v) : true) || 'Must contain numbers only'
```
cameron-eyds, line 48: This is default is no arguments are passed to the rule?

ppr-ui/src/store/mutations/mutations-model.ts
```diff
@@ -353,22 +354,10 @@ export const mutateUnsavedChanges = (state: StateIF, unsavedChanges: Boolean) =>
...
-export const mutateMhrRegistrationHomeModel = (state: StateIF, model: string) => {
-  state.stateModel.mhrRegistration.description.baseInformation.model = model
-}
-
 export const mutateMhrHomeDescription = (state: StateIF, { key, value }) => {
   state.stateModel.mhrRegistration.description[key] = value
 }
+
+export const mutateMhrBaseInformation = (state: StateIF, { key, value }) => {
+  set(state.stateModel.mhrRegistration.description.baseInformation[key], value)
```
cameron-eyds: is `set` still required here? 
I don't believe it should be

cameron-eyds (approved):
Looks good!

Just the one comment/question 👍

ppr-ui/src/store/mutations/mutations-model.ts
dimak1: You are correct. Missed it.

ppr-ui/src/composables/useInputRules.ts
dimak1: Yes, if nothing is passed, it checks for only numbers.

ppr-ui/src/store/mutations/mutations-model.ts
dimak1: Removed set().

dimak1:
Added Review Page to this PR. Uploaded screenshots for reference.

PR #864, opened 2022-07-08
Submitting Party - Part 1

cameron-eyds:
*Issue #:* /bcgov/entity#12579 & /bcgov/entity#12829

*Description of changes:*
*NOTE:* - This is an initial commit of the Submitting Party. There will be a PART 2 to wrap up Validations/Review Confirm Table by Gary. 

- Implements Submitting Party form, Updated Party Look up
- Updates Party lookup to accommodate MHR requirements
- Refactors to the Review Confirm View, pulled out Steps into their own Sub Review Components
- fixed Mhr Search cancel btn

### DIFF
--- a/ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
+++ b/ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
@@ -1,18 +1,73 @@
 <template>
-  <div>SUBMITTING PARTY PLACEHOLDER</div>
+  <div id="mhr-submitting-party-shim">
+    <section id="mhr-add-submitting-party" class="mt-10">
+      <h2>Submitting Party</h2>
+      <p class="mt-2">
+        Provide the name and contact information for the person or business submitting this registration. You can add
+        the submitting party information manually, or, if the submitting party has a Personal Property Registry party
+        code, you can look up the party code or name.
+      </p>
+
+      <!-- Parties Look Up -->
+      <PartySearch isMhrPartySearch />
+
+      <!-- Mhr Submitting Party Form -->
+      <MhrSubmittingParty />
+    </section>
+
+    <section id="mhr-submitting-party-reference" class="mt-10">
+      <h2>Attention or Reference Number</h2>
+      <p class="mt-2">
+        THIS COPY NEEDS TO BE REWRITTEN Add optional attention or reference number information for this transaction
+        for your own tracking purposes. This information is not used by the Manufactured Home Registry.
+      </p>
+
+      <!-- Insert Attention or Reference Number here -->
+    </section>
+  </div>
 </template>
 
 <script lang="ts">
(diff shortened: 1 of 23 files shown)


ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
```diff
@@ -1,18 +1,56 @@
...
+    <section id="mhr-add-submitting-party" class="mt-10">
+      <h2>Submitting Party</h2>
+      <p class="mt-2">
+        Provide the name and contact information for the person or business submitting this registration. You can add
+        the submitting party information manually, or, if the submitting party has a Personal Property Registry party
+        code, you can look up the party code or name.
+      </p>
+
+      <!-- Parties Look Up -->
+      <PartySearch isMhrPartySearch />
```
dimak1, line 12: `isMhrPartySearch` is not defined here.

dimak1 (approved):
Looks good! Just one comment, not sure if that would be coming in Part 2.

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
cameron-eyds: This is a prop

This is the same as `:isMhrPartySearch="true"`

cameron-eyds: Sorry, to add to that, the prop (all props) are defined on their respective components. 

In this case, it's defined with a default `false` value, so we don't have to refactor this components other implementations with this additional prop and so on.

doug-lovett (approved):
Very nice.

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
dimak1: Yea, I saw the prop, just wanted to check if we need to pass it (as it is always false).

cameron-eyds: Ah i see!
Yessir, as mentioned above, the inclusion of the prop here is the same as saying it's true (`:isMhrPartySearch="true"`)

dimak1: Ok, I think I got it. You anticipate using `PartySearch` component elsewhere with an option to pass different values for `isMhrPartySearch` prop, and in this case, we want `true`.

cameron-eyds: It's currently implemented in other flows, yes. 
We didn't want to impact those other flows but have some unique interactions within `PartySearch` and its subsequent child components for our MHR flow implementation. 

Yessir, absolutely, we want `true` in this case but also want to default too false, so we don't have to include this prop in other implementations.

dimak1 (approved):

PR #870, opened 2022-07-15
12579 submitting party staff

garyfan1:
*Issue #:* /bcgov/entity#12579

*Description of changes:*
add UI for mhr registration step 2
https://projects.invisionapp.com/share/BA12PVEDCQF8#/screens/467720541

image
image
image
image
image

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
+++ b/ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
@@ -22,16 +22,16 @@
           <!-- Insert Review mode of component here -->
           <v-row no-gutters class="px-6 pb-7">
             <v-col cols="3">
-              <h3 class="headers">Name</h3>
+              <h3 class="table-header">Name</h3>
             </v-col>
             <v-col cols="3" class="pl-1">
-              <h3 class="headers">Mailing Address</h3>
+              <h3 class="table-header">Mailing Address</h3>
             </v-col>
             <v-col cols="3" class="pl-3">
-              <h3 class="headers">Email Address</h3>
+              <h3 class="table-header">Email Address</h3>
             </v-col>
             <v-col cols="3" class="pl-4">
-              <h3 class="headers">Phone Number</h3>
+              <h3 class="table-header">Phone Number</h3>
             </v-col>
           </v-row>
 
@@ -43,20 +43,20 @@
             <v-col cols="3">
               <v-row no-gutters>
                 <v-col cols="1" class="mr-2">
-                  <v-icon color="black">
+                  <v-icon class="side-header-icon">
                     {{getMhrRegistrationSubmittingParty.businessName ? 'mdi-domain' : 'mdi-account'}}
                   </v-icon>
                 </v-col>
                 <v-col>
-                  <p class="first-col pt-1 font-weight-bold">
-                    {{ getName() }}
+                  <p class="side-header pt-1 font-weight-bold">
+                    {{ getSubmittingPartyName() }}
                   </p>
                 </v-col>
               </v-row>
             </v-col>
             <v-col cols="3" class="pl-1">
               <base-address
-                v-if="!emptyAddress()"
+                v-if="hasAddress"
                 class="content"
                 :schema="addressSchema"
                 :value="address"
@@ -78,7 +78,7 @@
 
           <v-row no-gutters class="px-6 py-7">
             <v-col cols="3">
-              <p class="first-col">Attention or<br>Reference Number</p>
+              <p class="side-header">Attention or<br>Reference Number</p>
             </v-col>
             <v-col cols="9">
               <p class="content">{{getMhrAttentionReferenceNum || '(Not Entered)'}}</p>
@@ -115,7 +115,8 @@ export default defineComponent({
     const localState = reactive({
       address: computed(() => getMhrRegistrationSubmittingParty.value.address),
       businessName: computed(() => getMhrRegistrationSubmittingParty.value.businessName),
-      personName: computed(() => getMhrRegistrationSubmittingParty.value.personName)
+      personName: computed(() => getMhrRegistrationSubmittingParty.value.personName),
+      hasAddress: computed(() => !Object.values(localState.address).every(val => !val))
     })
 
     const addressSchema = PartyAddressSchema
@@ -130,11 +131,7 @@ export default defineComponent({
       return `${phoneNum}${ext}`
     }
 
-    const emptyAddress = () => {
-      return Object.values(localState.address).every(val => !val)
-    }
-
-    const getName = () => {
+    const getSubmittingPartyName = () => {
       if (!localState.businessName && Object.values(localState.personName).every(val => !val)) {
         return '(Not Entered)'
       } else {
@@ -149,8 +146,7 @@ export default defineComponent({
       addressSchema,
       getMhrRegistrationSubmittingParty,
       getMhrAttentionReferenceNum,
-      getName,
-      emptyAddress,
+      getSubmittingPartyName,
       parsePhoneNumber,
       ...toRefs(localState)
     }
@@ -171,16 +167,22 @@ export default defineComponent({
   font-size: 16px;
 }
 
-.headers{
+.table-header {
   font-size: 14px;
   color: $gray9;
 }
-.first-col{
+
+.side-header {
   font-weight: bold;
   font-size: 16px;
   color: $gray9;
 }
-.content{
+
+.side-header-icon {
+  color: $gray9;
+}
+
+.content {
   margin-bottom: unset;
   line-height: 24px;
   font-size: 14px;
--- a/ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
+++ b/ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
@@ -2,7 +2,7 @@
   <v-card flat id="submitting-party-summary" class="mt-6 pb-6">
     <header class="review-header">
       <v-icon class="ml-2" color="darkBlue">mdi-account</v-icon>
-      <label class="font-weight-bold pl-2">Submitting Party</label>
+      <label class="font-weight-bold pl-2">Submitting Party </label>
     </header>
 
     <div :class="{ 'invalid-section': false }">
@@ -55,7 +55,14 @@
               </v-row>
             </v-col>
             <v-col cols="3">
-              <p class="content" v-html="parseAddress()"></p>
+              <base-address
+                v-if="!Object.values(address).every((val) => { return !val })"
+                class="content"
+                :schema="addressSchema"
+                :value="address"
+              >
+              </base-address>
+              <p v-else class="content"> (Not Entered) </p>
             </v-col>
             <v-col cols="3">
               <p class="content">{{getMhrRegistrationSubmittingParty.emailAddress || '(Not Entered)'}}</p>
@@ -84,14 +91,17 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, reactive, toRefs } from '@vue/composition-api'
+import { computed, defineComponent, reactive, toRefs } from '@vue/composition-api'
 import { RouteNames } from '@/enums'
 import { useGetters } from 'vuex-composition-helpers'
-import { useCountriesProvinces, formatAddress } from '@/composables/address/factories'
+import { BaseAddress } from '@/composables/address'
+import { PartyAddressSchema } from '@/schemas'
 
 export default defineComponent({
   name: 'SubmittingPartyReview',
-  components: {},
+  components: {
+    BaseAddress
+  },
   props: {},
   setup () {
     const {
@@ -104,26 +114,17 @@ export default defineComponent({
 
     const localState = reactive({})
 
-    const parseAddress = () => {
-      let address = getMhrRegistrationSubmittingParty.value.address
-      if (Object.values(address).every((val) => { return !val })) {
-        return '(Not Entered)'
-      }
-      address = formatAddress(address)
-      const street = address.street ? address.street + '<br>' : ''
-      const city = address.city ? address.city + ' ' : ''
-      const region = address.region ? address.region + ' &nbsp;' : ''
-      const postalCode = address.postalCode ? address.postalCode + '<br>' : ''
-      const country = address.country ? useCountriesProvinces().getCountryName(address.country) + '<br>' : ''
-      const details = address.deliveryInstructions ? '<br><i>' + address.deliveryInstructions + '</i>' : ''
-      return `${street}${city}${region}${postalCode}${country}${details}`
-    }
+    const addressSchema = PartyAddressSchema
+
+    const address = computed(() => {
+      return getMhrRegistrationSubmittingParty.value.address
+    })
 
     const parsePhoneNumber = () => {
       const phone = getMhrRegistrationSubmittingParty.value
       const phoneNum = phone.phoneNumber
       const ext = phone.phoneExtension ? ' &nbsp;Ext ' + phone.phoneExtension : ''
-      if (!phoneNum && !ext) {
+      if (!phoneNum || !ext) {
         return '(Not Entered)'
       }
       return `${phoneNum}${ext}`
@@ -131,9 +132,10 @@ export default defineComponent({
 
     return {
       RouteNames,
+      addressSchema,
       getMhrRegistrationSubmittingParty,
       getMhrAttentionReferenceNum,
-      parseAddress,
+      address,
       parsePhoneNumber,
       ...toRefs(localState)
     }
--- a/ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
+++ b/ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
@@ -132,11 +132,13 @@ export default defineComponent({
       if (!localState.businessName && Object.values(localState.personName).every(val => !val)) {
         return '(Not Entered)'
       } else {
-        return localState.businessName
-          ? localState.businessName
-          : localState.personName.middle
+        if (localState.businessName) {
+          return localState.businessName
+        } else {
+          return localState.personName.middle
             ? localState.personName.first + ' ' + localState.personName.middle + ' ' + localState.personName.last
             : localState.personName.first + ' ' + localState.personName.last
+        }
       }
     }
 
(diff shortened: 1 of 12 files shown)


doug-lovett (approved):
Scanning, looks good to me. Just a couple of comments.
1. The api and data model do not have a separate field for the attention/reference info. It is where the folio is also stored. You may want to confirm with Scott/Patrick that the value is captured in two places in the UI, otherwise we need an api change and either attention/ref or folio will not be stored in DB2.
2. Are some unit tests going to be added?

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
```diff
@@ -23,22 +23,55 @@
...
+            <label class="generic-label" :class="{'error-text': false}">Attention or Reference Number</label>
+          </v-col>
+          <v-col cols="12" sm="10" class="px-1">
+            <v-text-field
+              filled
+              id="attention-or-reference-number"
+              class="pr-2"
+              label="Attention or Reference Number (Optional)"
+              v-model="attentionReferenceNum"
+              :rules="attentionReferenceNumRule"
```
dimak1, line 38: If we are only validating for the length (which looks like we are), we can use `maxLength` rule.

ppr-ui/src/components/parties/party/PartyAutocomplete.vue
```diff
@@ -227,6 +248,11 @@ export default defineComponent({
   color: $gray7 !important;
 }
 
+.wasSelected {
```
dimak1: CSS class names usually have dash/hyphens.

ppr-ui/src/components/parties/party/PartyAutocomplete.vue
```diff
@@ -128,6 +136,13 @@ export default defineComponent({
       })
     })
 
+    const wasSelected = (val: SearchPartyIF) => {
+      if (localState.selectedCode === val.code) {
```
dimak1: Could be shortened to `localState.selectedCode === val.code ? 'wasSelected' : ''`

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
```diff
@@ -34,16 +86,55 @@
...
+    }
+
+    const parsePhoneNumber = () => {
+      const phone = getMhrRegistrationSubmittingParty.value
+      const phoneNum = phone.phoneNumber
+      const ext = phone.phoneExtension ? ' &nbsp;Ext ' + phone.phoneExtension : ''
+      if (!phoneNum && !ext) {
+        return '(Not Entered)'
+      }
+      return `${phoneNum}${ext}`
```
dimak1: There might be an error here if the `phoneNum` or `ext` are not defined (you have an `&&` check).

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
```diff
@@ -34,16 +86,55 @@
...
+      getMhrRegistrationSubmittingParty,
+      getMhrAttentionReferenceNum
+    } = useGetters<any>([
+      'getMhrRegistrationSubmittingParty',
+      'getMhrAttentionReferenceNum'
+    ])
+
     const localState = reactive({})
 
+    const parseAddress = () => {
```
dimak1: We can use a `BaseAddress` component here which can display the address with all the fields.
E.g. `<base-address :schema="addressSchema" :value="address" />`

ppr-ui/src/components/mhrRegistration/SubmittingParty/MhrSubmittingParty.vue
```diff
@@ -29,6 +29,13 @@
 
         <v-divider class="my-9 ml-0 mr-2" />
 
+        <v-card v-if="showImportantMsg" outlined id="important-message" class="rounded-0 mb-9">
```
dimak1: Can you check directly for `v-if="businessName"` here instead?

ppr-ui/src/components/mhrRegistration/SubmittingParty/MhrSubmittingParty.vue
```diff
@@ -208,7 +219,31 @@ export default defineComponent({
...
+
+    const middleNameRule = customRules(isStringOrNumber(), maxLength(15), invalidSpaces())
+
+    const lastNameRule = customRules(
+      required('Enter a last name'),
+      isStringOrNumber(),
+      maxLength(25),
+      invalidSpaces())
+
+    const businessNameRule = customRules(required('Business name is required'), invalidSpaces())
```
dimak1: No `maxLength` rules for these fields below?

dimak1 (changes requested):

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
garyfan1: yeah this is because the error text is different, in the design it says "Maximum 40 characters reached", but `maxLength()` can only return "Maximum 40 characters"

ppr-ui/src/components/mhrRegistration/SubmittingParty/MhrSubmittingParty.vue
garyfan1: not really, this message is only shown when we retrieve data from `partySearch`.
This message isn't shown when just switching from individual person to business.

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
```diff
@@ -16,34 +16,142 @@
...
+
+          <div class="px-4">
+            <v-divider />
+          </div>
+
+          <v-row no-gutters class="px-6 py-7">
+            <v-col cols="3">
+              <v-row no-gutters>
+                <v-col cols="1" class="mr-2">
+                  <v-icon color="black">
```
cameron-eyds: Would double check icon colour vs designs. I'm not sure we use plain black too often but correct me if i'm wrong!

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
```diff
@@ -16,34 +16,142 @@
...
+      const ext = phone.phoneExtension ? ' &nbsp;Ext ' + phone.phoneExtension : ''
+      if (!phoneNum) {
+        return '(Not Entered)'
+      }
+      return `${phoneNum}${ext}`
+    }
+
+    const emptyAddress = () => {
+      return Object.values(localState.address).every(val => !val)
+    }
```
cameron-eyds: As we discussed, this would be better suited as a computed property. 
Returning a singular value type without any arguments or potential for error responses, makes sense as a getter.

cameron-eyds: Also, might be better to update to `hasAddress` as true values are a little easier to read than a `!emptyAddress`

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
```diff
@@ -16,34 +16,142 @@
...
+
+    const getName = () => {
+      if (!localState.businessName && Object.values(localState.personName).every(val => !val)) {
+        return '(Not Entered)'
+      } else {
+        return localState.businessName
+          ? localState.businessName
+          : localState.personName.first + ' ' + (localState.personName.middle || '') + ' ' + localState.personName.last
+      }
+    }
```
cameron-eyds: `getName` would assume it's a getter vs a function expression, a little confusing :) 
Like above, would refactor to a getter. 
`getPartyName` or `getSubmittingPartyName` would also help describe it.

ppr-ui/src/components/mhrRegistration/SubmittingParty/MhrSubmittingParty.vue
cameron-eyds: Just to elaborate here, this messages requirements are 2-fold
- is a business
- and have they applied the values using the lookup

So yea, there is a little bit extra to the condition. 

That being said, could we update the property name to explain the property a bit more. 
ie `isBusinessLookup` or something along those lines.

Naming is hard :)

ppr-ui/src/components/mhrRegistration/SubmittingParty/MhrSubmittingParty.vue
```diff
@@ -41,7 +48,7 @@
                   class="pt-4 pr-2"
                   label="First Name"
                   v-model="submittingParty.personName.first"
-                  :rules="[]"
+                  :rules="firstNameRule"
```
cameron-eyds: Sorry, nitpicking here, but rules are plural not singular 😁 Could we update the rule names?

ppr-ui/src/components/parties/party/PartyAutocomplete.vue
```diff
@@ -128,6 +136,10 @@ export default defineComponent({
       })
     })
 
+    const wasSelected = (val: SearchPartyIF) => {
+      return localState.selectedCode === val.code ? 'was-selected' : ''
+    }
+
```
cameron-eyds, line 142: A good example of WHEN to create a function expression, nice work 👍

ppr-ui/src/interfaces/mhr-registration-interfaces/MhrRegistrationIF.ts
```diff
@@ -4,6 +4,11 @@ export interface MhrRegistrationIF {
   clientReferenceId: string,
   declaredValue: string,
   submittingParty: {
+    personName:{
+      first:string,
+      last: string,
+      middle: string
+    }
```
cameron-eyds, line 11: good catch 👍

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
cameron-eyds: Hmm, seems a bit weird to have unique messaging here vs everywhere else. 
Perhaps use our generic message. `Maximum 40 characters` and we push back a bit on the extra verbiage for consistency.

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
```diff
@@ -15,35 +15,68 @@
...
 })
-export default class SubmittingParty extends Vue {}
+export default class SubmittingParty extends Vue {
+  @Action setMhrAttentionReferenceNum : ActionBindingIF
+
+  private attentionReferenceNum = ''
+
+  private attentionReferenceNumRule = [
+    v => (v || '').length <= 40 || 'Maximum 40 characters reached'
+  ]
```
cameron-eyds: As Dima suggested above, we should use the generic rule and push back for messaging consistency.

cameron-eyds (commented):
Looks great Gary, nice job!
Just a few minor comments/suggestions. 

Tests can be in a follow up PR, that way we can get this to UX asap 👍

cameron-eyds (approved):
Thanks for the updates Gary, looks good 👍 

Follow up PR for some basic unit testing would be awesome, nice job.

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
garyfan1: updated, so now the condition only looks at phone number, and whenever it's undefined, the field will show "not entered"

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
garyfan1: updated using the compostable !

ppr-ui/src/components/parties/party/PartyAutocomplete.vue
garyfan1: updated to `was-selected`

ppr-ui/src/components/parties/party/PartyAutocomplete.vue
garyfan1: shortened the if statement to ternary!

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
garyfan1: updated, now using the `baseAddress` component.

ppr-ui/src/components/mhrRegistration/SubmittingParty/MhrSubmittingParty.vue
garyfan1: changed the name to `isBusinessLookup`

ppr-ui/src/components/mhrRegistration/SubmittingParty/MhrSubmittingParty.vue
garyfan1: added `maxLength` to it.

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
dimak1: Great, thanks!

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
```diff
@@ -15,35 +15,67 @@
       <MhrSubmittingParty />
     </section>
 
-    <section id="mhr-submitting-party-reference" class="mt-10">
+    <section id="mhr-submitting-partyy-reference" class="mt-10">
```
dimak1: Typo?

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
```diff
@@ -16,34 +16,138 @@
...
+      return `${phoneNum}${ext}`
+    }
+
+    const getSubmittingPartyName = () => {
+      if (!localState.businessName && Object.values(localState.personName).every(val => !val)) {
+        return '(Not Entered)'
+      } else {
+        return localState.businessName
+          ? localState.businessName
+          : localState.personName.first + ' ' + (localState.personName.middle || '') + ' ' + localState.personName.last
```
dimak1: If there is no _middle_ name - there would be two spaces between _first_ and _last_

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
dimak1: Better, but could be simplified even further: `return phoneNum ?` `${phoneNum}${ext}` `: '(Not Entered)'`

dimak1 (commented):
@garyfan1 looks great, few more small comments, and needs a sync with main branch.

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
garyfan1: oh yes it is, updated

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
garyfan1: updated!

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
```diff
@@ -16,34 +16,137 @@
...
+    }
+
+    const getSubmittingPartyName = () => {
+      if (!localState.businessName && Object.values(localState.personName).every(val => !val)) {
+        return '(Not Entered)'
+      } else {
+        return localState.businessName
+          ? localState.businessName
+          : localState.personName.middle
+            ? localState.personName.first + ' ' + localState.personName.middle + ' ' + localState.personName.last
```
dimak1, line 139: As per best practices, we try to avoid having nested ternary operators. Also, I realized that the middle name will always be defined, as in the `state-model`, `middle: ''`.

dimak1 (approved):
Approved with a comment (if you want to improve it).

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
garyfan1: replaced the first ternary with an if statement!

ppr-ui/src/utils/feature-flags.ts
```diff
@@ -10,7 +10,7 @@ export const defaultFlagSet: LDFlagSet = {
   'search-registration-number': false,
   'search-serial-number': false,
   'mhr-ui-enabled': false,
-  'mhr-registration-enabled': false
+  'mhr-registration-enabled': true
```
dimak1: This is on purpose?

dimak1 (changes requested):

ppr-ui/src/utils/feature-flags.ts
garyfan1: no sorry, this is when I got lazy and did `git commit -am`, reverted the flag to false.

ppr-ui/src/components/mhrRegistration/ReviewConfirm/SubmittingPartyReview.vue
```diff
@@ -16,34 +16,139 @@
...
+      const phoneNum = phone.phoneNumber
+      const ext = phone.phoneExtension ? ' &nbsp;Ext ' + phone.phoneExtension : ''
+      return phoneNum ? `${phoneNum}${ext}` : '(Not Entered)'
+    }
+
+    const getSubmittingPartyName = () => {
+      if (!localState.businessName && Object.values(localState.personName).every(val => !val)) {
+        return '(Not Entered)'
+      } else {
+        if (localState.businessName) {
```
dimak1, line 135: I think this will always be true since our model is initially set as `businessName: ''`. What you think @cameron-freshworks . cc: @garyfan1

cameron-eyds: I believe an empty string in typescript resolves as `falsey` 
So `businessName:  ''` being initialized would still equate too `!localState.businessName` ie `false` 
@garyfan1 Should/could verify though

garyfan1: I was thinking empty string is falsy? so even if `business == ''`, `if(business)` is still false?

dimak1: @cameron-freshworks right, you are correct. Resolving this.

dimak1 (approved):

PR #875, opened 2022-07-19
12974 display owner status in selection list and results

ChasBoucher:
*Issue #:* /bcgov/entity12974

Change status to Owner Status in MHR Search results


ppr-ui/src/components/tables/mhr/SearchedResults.vue
```diff
@@ -124,7 +124,7 @@
             {{ item.mhrNumber }}
           </template>
           <template v-slot:[`item.state`]="{ item }">
-            {{ item.status }}
+            {{ getOwnerStatus(item.ownerStatus) }}
```
cameron-eyds, line 127: Was `status` too  `ownerStatus` a recent property name update done on the BE?

ppr-ui/src/components/tables/mhr/SearchedResults.vue
```diff
@@ -290,6 +290,19 @@ export default defineComponent({
       } else return '-'
     }
 
+    const getOwnerStatus = (ownerStatus: string): string => {
+      if (ownerStatus) {
+       if(ownerStatus=='PREVIOUS')
```
cameron-eyds: Minor spacing issues (linter might complain) 
also the equality operator we use traditionally  is the strict equality (`===`), which also checks for type. 
The `==` equality operator will type coerce properties. 
Super small thing but good for consistency 👍 

`if (ownerStatus === 'PREVIOUS')`

ppr-ui/src/components/tables/mhr/SearchedResults.vue
```diff
@@ -290,6 +290,19 @@ export default defineComponent({
...
+       {
+         return 'HISTORICAL'
+       } 
+       else
+         return ownerStatus
+      }
+      else
+         return ''
```
cameron-eyds: Minor syntax clean up, although not wrong, i would keep the `else` on the same line as the closing bracket. 
Also add brackets for the statement or keep the return on the same line. 

ie `} else return ownerStatus` or `} else { return ownerStatus }`
Some people prefer the brackets for consistency, i'm good with either.

cameron-eyds (commented):
Nice job!
Just a few comments/questions
Also see the lint issues in the ci.

doug-lovett (approved):

ppr-ui/src/components/tables/mhr/SearchedResults.vue
ChasBoucher: Good point.

ppr-ui/src/components/tables/mhr/SearchedResults.vue
ChasBoucher: @cameron-freshworks I am not sure what BE is, but ownerStatus was the field that should have been used.  The api was returning it.  It is the owner.status.  Ali confirmed it.

cameron-eyds: Sounds good, thanks!

cameron-eyds (approved):
Thank you! Looks good 👍

PR #879, opened 2022-07-25
12584 tombstone image and wording

garyfan1:
*Issue #:* /bcgov/entity#12584

*Description of changes:*
change the text of breadcrumb
change the image of tombstone
when logged in as a staff (staff OR ppr_staff OR sbc OR helpdesk)


ppr-ui/src/components/common/Breadcrumb.vue
```diff
@@ -127,6 +127,7 @@ export default defineComponent({
     const handleStaff = (breadcrumbText): string => {
       if (localState.isStaff) {
         breadcrumbText = breadcrumbText.replace('My', 'Staff')
+        if (breadcrumbText === 'BC Registries Dashboard') breadcrumbText = 'Staff Dashboard'
```
dimak1: This is not the best way to handle it. It is not future-proof in case the text gets updated. Is there a way to rework it?

dimak1 (changes requested):

ppr-ui/src/components/common/Breadcrumb.vue
cameron-eyds: For sure, agreed we shouldn't use a string here. 
Perhaps you could define the `Staff Dashboard` as a const and replace `tombstoneBreadcrumbDashboard[1].text` or redefine it etc vs a string swap/replace?

ppr-ui/src/components/common/Breadcrumb.vue
```diff
@@ -86,6 +86,16 @@ export default defineComponent({
...
+        if (localState.isStaff) {
+          const staffBreadcrumbText = 'Staff Dashboard'
+          tombstoneBreadcrumbDashboard[0].text = staffBreadcrumbText
+          tombstoneBreadcrumbDischarge[0].text = staffBreadcrumbText
+          tombstoneBreadcrumbRenewal[0].text = staffBreadcrumbText
+          tombstoneBreadcrumbAmendment[0].text = staffBreadcrumbText
+          tombstoneBreadcrumbRegistration[0].text = staffBreadcrumbText
+          tombstoneBreadcrumbSearch[0].text = staffBreadcrumbText
+          tombstoneBreadcrumbSearchConfirm[0].text = staffBreadcrumbText
+        }
```
cameron-eyds: Not your fault, this is a very verbose computed value. 
But perhaps we refactor the way it returns the `breadcrumbs` value. 

ie create a let property (let breadcrumbs) and assign vs returning it. 
In that case, you would only have to modify the return `breadcrumbs` once at the end.

Sorry for this, we really need to rework the breadcrumb architecture as whole and apply them dynamically.

ppr-ui/src/components/common/Breadcrumb.vue
```diff
@@ -86,15 +86,18 @@ export default defineComponent({
...
         ]
+        const allTombstoneBreadcrumbs = [
+          tombstoneBreadcrumbDashboard,
+          tombstoneBreadcrumbDischarge,
+          tombstoneBreadcrumbRenewal,
+          tombstoneBreadcrumbAmendment,
+          tombstoneBreadcrumbRegistration,
+          tombstoneBreadcrumbSearch,
+          tombstoneBreadcrumbSearchConfirm
+        ]
```
cameron-eyds, line 97: This works too 👍

ppr-ui/src/components/common/Breadcrumb.vue
```diff
@@ -86,6 +86,19 @@ export default defineComponent({
...
+        if (localState.isStaff) {
+          const staffBreadcrumbText = 'Staff Dashboard'
+          allTombstoneBreadcrumbs.forEach((val) => { val[0].text = staffBreadcrumbText })
```
cameron-eyds: Might not need a const if just used locally here,
I was thinking if it was defined in the BreadCrumb resource or something 😁 
All good either way though, no change required 👍

garyfan1: oh sorry didn't see your comments before making some other changes, but I think the updated code has pretty much the same idea, I'm just using `for..of` instead of `foreach`

cameron-eyds: You bet! All good sir 👍

garyfan1: removed the redundant `const`!

cameron-eyds (approved):

dimak1 (approved):

PR #881, opened 2022-07-26
add Circa to year when the approximate checkbox is checked

garyfan1:
*Issue #:* /bcgov/entity#13018

*Description of changes:*
add Circa to manufactured year when the approximate checkbox is checked.

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
+++ b/ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
@@ -31,7 +31,7 @@
             <h3>Year of Manufacture</h3>
           </v-col>
           <v-col cols="9">
-            <p>{{ getMhrRegistrationHomeDescription.baseInformation.year || '(Not Entered)' }}</p>
+            <p>{{ manufactureYear }}</p>
           </v-col>
         </v-row>
         <v-row no-gutters class="pt-3 px-6">
@@ -179,6 +179,16 @@ export default defineComponent({
       }),
       engineerDisplayDate: computed((): string => {
         return yyyyMmDdToPacificDate(getMhrRegistrationHomeDescription.value?.engineerReportDate, true)
+      }),
+      manufactureYear: computed((): string => {
+        const year = getMhrRegistrationHomeDescription.value.baseInformation.year
+        if (!year) {
+          return '(Not Entered)'
+        }
+        if (getMhrRegistrationHomeDescription.value.baseInformation.circa) {
+          return 'Circa ' + year
+        }
+        return year
       })
     })
 
--- a/ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
+++ b/ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
@@ -31,7 +31,13 @@
             <h3>Year of Manufacture</h3>
           </v-col>
           <v-col cols="9">
-            <p>{{ manufactureYear }}</p>
+            <p v-if="getMhrRegistrationHomeDescription.baseInformation.year">
+              {{ getMhrRegistrationHomeDescription.baseInformation.circa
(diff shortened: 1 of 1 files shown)


ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
```diff
@@ -179,6 +179,16 @@ export default defineComponent({
...
+      }),
+      manufactureYear: computed((): string => {
+        const year = getMhrRegistrationHomeDescription.value.baseInformation.year
+        if (!year) {
+          return '(Not Entered)'
+        }
+        if (getMhrRegistrationHomeDescription.value.baseInformation.circa) {
+          return 'Circa ' + year
+        }
+        return year
```
cameron-eyds: Are Base Information / circa / year always initialized before this is processed? 
If not, might want to protect yourself from type errors here by optional chaining. 

These would show up in the console if there was an issue 👍

ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
```diff
@@ -179,6 +179,16 @@ export default defineComponent({
       }),
       engineerDisplayDate: computed((): string => {
         return yyyyMmDdToPacificDate(getMhrRegistrationHomeDescription.value?.engineerReportDate, true)
+      }),
+      manufactureYear: computed((): string => {
```
cameron-eyds: typo in `manufacturer`? 
Or maybe not, unless that is intentional. Could be read both ways i suppose!

cameron-eyds (approved):

ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
dimak1: Not sure if we need a computed prop for this. We can simplify and use at the template level. E.g.
```
<p v-if="getMhrRegistrationHomeDescription.baseInformation.year">
...Check for circa here...
</p>
<p v-else>
(Not Entered)
</p>
```

dimak1 (changes requested):

ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
garyfan1: the model was initialized like this
image
so I think it should be ok?

ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
garyfan1: this is because the text field is called "Year of Manufacture", so I call it "manufactureYear"

cameron-eyds: Makes sense, thanks for checking!

ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
cameron-eyds: Should be fine, as long as you don't see any errors in the console when firing up the Registration 👍

ppr-ui/src/components/mhrRegistration/ReviewConfirm/YourHomeReview.vue
cameron-eyds: Depends on what the check is,
I'm in the camp of keeping logic out of the template but if it can be done within a basic ternary than all good. Your call but less logic in the template the better 👍

garyfan1: oops pushed before seeing your comment again; anyways, let me know if you would like to revert the changes.

dimak1 (approved):

PR #882, opened 2022-07-27
HomeLocation and Review

cameron-eyds:
*Issue #:* /bcgov/entity#12962

*Description of changes:*
* Home Location Step View
* Home Location Selector Component
* Civic Address Placeholder
* Home Location Review Components

* Update enums/store model for Home Location
* Minor Style/Ux updates

*Notes:*
* Unit testing will be in a follow up PR in order to unblock continued work on Step 4 (Home Location)
* Validations are NOT COMPLETE IN VALIDATION MATRIX -- Still Pending
* PID Verification is NOT COMPLETE AS PER REQUIREMENTS -- Still Pending
* Pending some Model changes, local state is subject to adjustments. 
* The above pending tasks are to be completed in follow up Tickets. 


ppr-ui/src/components/common/PidNumber.vue
```diff
@@ -0,0 +1,153 @@
...
+
+    const emitPid = (): void => { context.emit('setPid', localState.pidNumber) }
+
+    watch(() => localState.pidOne, () => {
+      // @ts-ignore - function exists
+      if (localState.pidOne.length === 3) vue.nextTick(() => { context.refs.pidTwoRef.focus() })
+    })
+    watch(() => localState.pidTwo, () => {
+      // @ts-ignore - function exists
+      if (localState.pidTwo.length === 3) vue.nextTick(() => { context.refs.pidThreeRef.focus() })
```
dimak1, line 100: Nice!
Do we need to go back to PID 1 when we delete the PID 2 fields?

cameron-eyds: Thanks! 
No i didn't build that interaction in, current implementation is an auto focus on 1, which proceeds to 2 and 3 subsequently. I didn't see anything saying deletions would jump to the previous field but we could definitely build that in?
Do you recall Tracey mentioning that as a requirement? Should we build it now or maybe wait until after UX?

ppr-ui/src/views/newMhrRegistration/MhrRegistration.vue
```diff
@@ -136,10 +136,20 @@ export default class MhrRegistration extends Vue {
...
+        class: 'registration-list-item',
+        disabled: false,
+        divider: false,
+        group: 3,
+        registrationTypeUI: 'Manufactured Home Registration',
+        registrationTypeAPI: 'MHR',
+        text: 'Manufactured Home Registration (MHR)'
       })
-      return
+      // return
```
cameron-eyds: Removing Dev Code 👍

cameron-eyds: * REMOVED

ChasBoucher (approved):
Looks Good and thanks for putting in some placeholders.. for civic address.

cameron-eyds:
> Looks Good and thanks for putting in some placeholders.. for civic address.

Thanks! No problem at all 👍

ChasBoucher (approved):
LGTM

ppr-ui/src/components/mhrRegistration/HomeLocation/HomeLocationType.vue
```diff
@@ -0,0 +1,236 @@
...
+              />
+              <v-expand-transition>
+                <v-form
+                  v-if="locationTypeOption === HomeLocationTypes.LOT"
+                  ref="lotForm"
+                  v-model="isValidLot"
+                >
+                  <v-text-field
+                    filled
+                    id=""
```
dimak1: Maybe we can clean up empty IDs?

ppr-ui/src/components/common/PidNumber.vue
```diff
@@ -0,0 +1,152 @@
...
+      pidOne: '',
+      pidTwo: '',
+      pidThree: '',
+      enablePidLoader: false,
+      pidNumber: computed(() => {
+        return `${localState.pidOne}-${localState.pidTwo}-${localState.pidThree}`
+      }),
+      isValidPid: computed(() => {
+        return (
+          (localState.pidOne ? /^\d+$/g.test(localState.pidOne) : true) &&
```
dimak1, line 82: Do we need to add a rule to min length of 3 digits?

ppr-ui/src/components/common/PidNumber.vue
```diff
@@ -0,0 +1,152 @@
...
+
+    const emitPid = (): void => { context.emit('setPid', localState.pidNumber) }
+
+    watch(() => localState.pidOne, () => {
+      // @ts-ignore - function exists
+      if (localState.pidOne.length === 3) vue.nextTick(() => { context.refs.pidTwoRef.focus() })
+    })
+    watch(() => localState.pidTwo, () => {
+      // @ts-ignore - function exists
+      if (localState.pidTwo.length === 3) vue.nextTick(() => { context.refs.pidThreeRef.focus() })
```
dimak1, line 100: Nice! What about going backwards when deleting pid numbers (e.g. from pid 3 to 2 to 1)?

dimak1 (approved):
Great job! Approved with some comments. Also, I remember for pasting PID they wanted to distribute the number between three input fields. Is it still a req or it was dropped?

ppr-ui/src/components/common/PidNumber.vue
dimak1: Agree, let's wait for UX to come back with feedback (if any!). Thanks!

cameron-eyds: Good idea, will do!

ppr-ui/src/components/mhrRegistration/HomeLocation/HomeLocationType.vue
cameron-eyds: Definitely, good catch. Will remove until required

ppr-ui/src/components/common/PidNumber.vue
cameron-eyds: Currently there is no validations in the validation matrix, so sounds like this will be applied once those requirements have been figured out. 

The pid as a whole however will most likely be an 11 char min/max to include the 9 digits and 2 dashes.
The validation on this is a bit* weird, we will have to apply the validation error to only the one field but the value would be the result of the 3 fields if that makes sense. 
I've done a little bit of that already with the custom error and hint etc

dimak1: yea ok that makes sense, thanks

ppr-ui/src/components/mhrRegistration/HomeLocation/HomeLocationType.vue
cameron-eyds: Removed! 
Also updated the Radio Btn id's 👍

cameron-eyds:
> Great job! Approved with some comments. Also, I remember for pasting PID they wanted to distribute the number between three input fields. Is it still a req or it was dropped?

Good question! I'm not sure about this requirement. Will have to follow up with Tracey and Patrick.

PR #905, opened 2022-08-16
MHR: Unsaved Changes Handler

cameron-eyds:
*Issue #:* /bcgov/entity#12542

*Description of changes:*
* Implements the handlers for unsaved changes on navigation in the MHR
* Validation Reset in MHR Wrapper


ChasBoucher (approved):
Looks Good..

dimak1 (approved):
Looks good, just two things:
- Saved / Unsaved changes were a bit confusing, as we don't really have a save functionality. Then I read the task, and I guess we can follow the requirement wording.
- Tasks mentions there should be a modal when user is trying to exit prematurely. Is that still coming?

cameron-eyds:
> Looks good, just two things:
> 
> * Saved / Unsaved changes were a bit confusing, as we don't really have a save functionality. Then I read the task, and I guess we can follow the requirement wording.
> * Tasks mentions there should be a modal when user is trying to exit prematurely. Is that still coming?

@dimak1 Agreed, we are not exactly saving are we 😁 But yes, as per requirements it was put in. 
You bet, the dialog is already in, it is a pre-existing one.

doug-lovett (approved):
Looks good.

PR #914, opened 2022-08-20
Document ID for Submitting Party

dimak1:
*Issue #:* /bcgov/entity#13376

*Description of changes:*
- Add Document ID on Submitting Party page
- Add required validation (no look up)
- Show Document ID on the Review page

### Screenshots

Screen Shot 2022-08-22 at 8 17 29 AM

### DIFF
--- a/ppr-ui/src/store/getters/state-getters.ts
+++ b/ppr-ui/src/store/getters/state-getters.ts
@@ -662,6 +662,10 @@ export const getMhrRegistrationHomeOwners = (state: StateIF): MhrRegistrationHom
   return owners
 }
 
+export const getMhrRegistrationDocumentId = (state: StateIF): any => {
+  return state.stateModel.mhrRegistration.documentId
+}
+
 export const getMhrAttentionReferenceNum = (state: StateIF): any => {
   return state.stateModel.mhrRegistration.attentionReferenceNum
 }
--- a/ppr-ui/src/store/getters/state-getters.ts
+++ b/ppr-ui/src/store/getters/state-getters.ts
@@ -662,7 +662,7 @@ export const getMhrRegistrationHomeOwners = (state: StateIF): MhrRegistrationHom
   return owners
 }
 
-export const getMhrRegistrationDocumentId = (state: StateIF): any => {
+export const getMhrRegistrationDocumentId = (state: StateIF): string => {
   return state.stateModel.mhrRegistration.documentId
 }
 
--- a/ppr-ui/src/store/mutations/mutations-model.ts
+++ b/ppr-ui/src/store/mutations/mutations-model.ts
@@ -372,6 +372,10 @@ export const mutateMhrHomeOwners = (state: StateIF, owners: Array<MhrRegistratio
   state.stateModel.mhrRegistration.owners = owners
 }
 
+export const mutateMhrRegistrationDocumentId = (state: StateIF, value) => {
+  state.stateModel.mhrRegistration.documentId = value
+}
+
 export const mutateMhrAttentionReferenceNum = (state: StateIF, value) => {
   state.stateModel.mhrRegistration.attentionReferenceNum = value
 }
--- a/ppr-ui/src/store/mutations/mutations-model.ts
+++ b/ppr-ui/src/store/mutations/mutations-model.ts
@@ -372,7 +372,7 @@ export const mutateMhrHomeOwners = (state: StateIF, owners: Array<MhrRegistratio
   state.stateModel.mhrRegistration.owners = owners
 }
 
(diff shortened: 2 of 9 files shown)


doug-lovett (approved):
Looks good.

garyfan1 (approved):

ppr-ui/src/store/getters/state-getters.ts
```diff
@@ -662,6 +662,10 @@ export const getMhrRegistrationHomeOwners = (state: StateIF): MhrRegistrationHom
   return owners
 }
 
+export const getMhrRegistrationDocumentId = (state: StateIF): any => {
```
cameron-eyds: We could probably declare a `string` type return, since we know that for sure here

ppr-ui/src/store/mutations/mutations-model.ts
```diff
@@ -372,6 +372,10 @@ export const mutateMhrHomeOwners = (state: StateIF, owners: Array<MhrRegistratio
   state.stateModel.mhrRegistration.owners = owners
 }
 
+export const mutateMhrRegistrationDocumentId = (state: StateIF, value) => {
+  state.stateModel.mhrRegistration.documentId = value
+}
+
```
cameron-eyds: Likewise here, we know `value` is of type `string` here i think

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
```diff
@@ -96,18 +132,53 @@ export default defineComponent({
...
+        await context.refs.documentIdForm.validate()
+      }
+    )
+
+    watch(
+      () => localState.attentionReferenceNum,
+      (val: string) => {
+        setMhrAttentionReferenceNum(val)
+      }
+    )
```
cameron-eyds: please revert the format here

cameron-eyds (approved):
Looks good, just some minor types and formatting comments to address, thanks!

ChasBoucher (approved):
Other than Cameron's comments, it looks good.

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
dimak1: Manually updated and saved without formatting. It might cause formatting discrepancies again in the future. I'm using prettier settings `/ppr/ppr-ui/prettier.config.js`. Which formatter/settings you are using? I tried multiple and none could format the way yours does...

ppr-ui/src/store/mutations/mutations-model.ts
dimak1: Added return type, thanks!

ppr-ui/src/store/getters/state-getters.ts
dimak1: Added return type, thanks!

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
cameron-eyds: I don't use any.  
Webstorm has it's own configurations that help with imports and whatnot, as well as following the defined linting rules but otherwise i'm not a fan of relying on something to format the code.
Prettier is opinionated code formatting but i'm not sure i agree with the configuration. I'm all for it if it helps with readability/clean code but in this case, my opinion is that it isn't helping with readability.

cameron-eyds:
Thanks for the updates! Merge away whenever you are ready 💯

ppr-ui/src/views/newMhrRegistration/SubmittingParty.vue
dimak1: We should consolidate and use the same formatter/settings, so we don't have these issues. Otherwise, everyone on the team will have formatting changes and they commit the code.

cameron-eyds: For sure, although the only discrepancies to date has been with prettier. It's not something we've relied on in other projects but if you feel strongly about this, we can get all get onboard with it. 

To do that though, we will need to adjust and agree on the format output by prettier, as of yet i'm not convinced on its formatting.

cameron-eyds: To add to this, doesn't the linting already provide enough to enable consistency between devs?

dimak1: The `ppr-ui` comes with Prettier and a config `/ppr/ppr-ui/prettier.config.js` when cloned. So during the initial setup, I figured we are using these to format the code across the team. I don't mind the format/style or tool, as long we are consistent across. Let's have a chat sometime.

PR #916, opened 2022-08-24
13171 fix email phone validations

ChasBoucher:
*Issue #:* /bcgov/entity13171

*Description of changes:*
Fix/Add validations for Email/Phone

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
@@ -139,13 +139,12 @@
             <v-col cols="6">
               <v-text-field
                 id="phone-number"
-                v-mask="'(###) ###-####'"
+                v-mask="'(NNN) NNN-NNNN'"
                 v-model="owner.phoneNumber"
                 filled
                 :rules="phoneNumberRules"
-                label="Phone Number"
+                label="Phone Number (Optional)"
                 data-test-id="phone-number"
-                validate-on-blur
               />
             </v-col>
             <v-col cols="6">
@@ -153,7 +152,7 @@
                 id="phone-ext"
                 v-model="owner.phoneExtension"
                 filled
-                :rules="maxLength(5, true)"
+                :rules="phoneExtensionRules"
                 label="Extension (Optional)"
                 data-test-id="phone-ext"
               />
@@ -276,7 +275,7 @@ export default defineComponent({
     }
   },
   setup (props, context) {
-    const { required, customRules, maxLength, minLength } = useInputRules()
+    const { required, customRules, maxLength, minLength, isPhone, isNumber, invalidSpaces } = useInputRules()
 
     const {
       getSideTitle,
@@ -363,9 +362,9 @@ export default defineComponent({
         maxLength(70)
       ),
       phoneNumberRules: customRules(
-        required('Enter a phone number'),
-        maxLength(15)
-      )
+        isPhone(14, null)
+      ),
+      phoneExtensionRules: customRules(isNumber(), invalidSpaces(), maxLength(5))
     })
 
     const done = (): void => {
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
@@ -140,7 +140,7 @@
               <v-text-field
                 id="phone-number"
                 v-mask="'(NNN) NNN-NNNN'"
-                v-model="owner.phoneNumber"
+                v-model="displayPhone"
                 filled
                 :rules="phoneNumberRules"
                 label="Phone Number (Optional)"
@@ -236,7 +236,7 @@ import { useHomeOwners } from '@/composables/mhrRegistration'
 import { AutoComplete } from '@/components/search'
 import { BaseAddress } from '@/composables/address'
 import { PartyAddressSchema } from '@/schemas'
-import { focusOnFirstError } from '@/utils'
+import { focusOnFirstError, fromDisplayPhone } from '@/utils'
 import { VueMaskDirective } from 'v-mask'
 
 /* eslint-disable no-unused-vars */
@@ -355,6 +355,7 @@ export default defineComponent({
       isAddressFormValid: false,
       triggerAddressErrors: false,
       isHelpPanelOpen: false,
+      displayPhone: props.editHomeOwner !== null ? props.editHomeOwner.phoneNumber : '',
       firsNameRules: customRules(required('Enter a first name'), maxLength(15)),
       lastNameRules: customRules(required('Enter a last name'), maxLength(25)),
       orgNameRules: customRules(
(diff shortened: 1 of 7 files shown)


ppr-ui/src/composables/useInputRules.ts
```diff
@@ -55,13 +55,23 @@ export const useInputRules = () => {
...
 
+  const isPhone = (
+    maxDigits: number = null,
+    customMsg: string = null
+  ): Array<Function> => {
+    return [
+      v => ((v && maxDigits) ? v.length >= maxDigits : true) || 'Minimum 10 characters',
+      v => (v ? /^\d+$/g.test(v.replace('(', '').replace(') ', '').replace('-', '')) : true) || `${customMsg || 'Must contain numbers only'}`
+    ]
+  }
```
cameron-eyds: Arn't phone numbers strings to our specs that can also contain characters like `-` `()` and whatnot?

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
```diff
@@ -363,9 +362,9 @@ export default defineComponent({
         maxLength(70)
       ),
       phoneNumberRules: customRules(
-        required('Enter a phone number'),
-        maxLength(15)
-      )
+        isPhone(14, null)
```
cameron-eyds: In your rule you've defined `customMsg` with a default, so you don't need to pass in `null` here.

ppr-ui/src/composables/useInputRules.ts
```diff
@@ -55,13 +55,23 @@ export const useInputRules = () => {
...
     }
   ]
 
+  const isPhone = (
+    maxDigits: number = null,
+    customMsg: string = null
+  ): Array<Function> => {
+    return [
+      v => ((v && maxDigits) ? v.length >= maxDigits : true) || 'Minimum 10 characters',
+      v => (v ? /^\d+$/g.test(v.replace('(', '').replace(') ', '').replace('-', '')) : true) || `${customMsg || 'Must contain numbers only'}`
```
cameron-eyds: Could you elaborate on this regex? 
By looking at the message, i would assume it's checking for numbers only but it seems like it's doing a lot more.

ppr-ui/tests/unit/MhrSubmittingParty.spec.ts
```diff
@@ -139,13 +139,13 @@ describe('mhr submitting party', () => {
     expect(submittingPartySection.findAll(ERROR_MSG).length).toBe(0)
 
     const phoneNum = submittingPartySection.find('#submitting-party-phone')
-    phoneNum.setValue('(778) 123-456')
+    phoneNum.setValue('(778) 123-4563')
```
cameron-eyds: This doesn't exactly adhere to the rule you've defined above, `must contain numbers only` so i'm a little confused.

ppr-ui/src/composables/useInputRules.ts
```diff
@@ -55,13 +55,23 @@ export const useInputRules = () => {
   }
 
   const isEmail = (): Array<Function> => [
-    (v: string) => !!v || 'Email address is required',
+    (v: string) => !!v || 'Enter an email address',
```
dimak1, line 58: Could you confirm that we are updating this error message for all email fields across the app?

ppr-ui/src/composables/useInputRules.ts
```diff
@@ -55,13 +55,23 @@ export const useInputRules = () => {
...
+      return pattern.test(v) || 'Enter a valid email address'
     }
   ]
 
+  const isPhone = (
+    maxDigits: number = null,
+    customMsg: string = null
+  ): Array<Function> => {
+    return [
+      v => ((v && maxDigits) ? v.length >= maxDigits : true) || 'Minimum 10 characters',
```
dimak1: We probably don't need this check, as we can use `minLength` instead.

ppr-ui/src/composables/useInputRules.ts
dimak1: Just my feedback, if we enter `(123) A11-2233`, the message 'Must contain numbers' is confusing. Maybe more of a question to UX team...

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
dimak1: No need for a second argument, as we already have a default value of null.

dimak1 (changes requested):

doug-lovett (commented):
Just a comment that the DB2 database phone number max length is 10, so it should be just digits and min length = max length = 10.

ppr-ui/src/composables/useInputRules.ts
ChasBoucher: I will check with UX.

I did a search across ppr for occurances of isEmail and it does not seem to be called anywhere else.

ppr-ui/src/composables/useInputRules.ts
ChasBoucher: The reason I did that is because of v-mask.  It treats the masked characters as as character.   Those characters are not entered by the user, but are included in the field... The mask characters will be stripped when writing to store.  Doug commented on this.. I am going to make that change now.

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
ChasBoucher: Good point... Changed.

ppr-ui/src/composables/useInputRules.ts
ChasBoucher: Very good question... Phone numbers can be only numbers.  The mask is only for user validation and display on review form.  Doug made a comment regarding the phone field in db2 is only 10 digits... I am going to remove the extra characters before writing to store...

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
ChasBoucher: Dima caught that one as welll.  I have made the change..

ppr-ui/tests/unit/MhrSubmittingParty.spec.ts
ChasBoucher: Good point... I will change that test.  The test is for numbers so I should test for that message...

ChasBoucher:
@doug-lovett I am going to ensure that the phone written to store has only numbers and is max 10 digits... I think it is best to do that on this ticket, so it will not conflict with the code for writing to db2.

ppr-ui/src/composables/useInputRules.ts
ChasBoucher: Very good question... v-mask includes the masked characters in the field.  We require only 10 digits.. The validation ensures that we are validating  for numbers not the extra characters.  The extra characters will not be stored in db...

ppr-ui/src/composables/useInputRules.ts
cameron-eyds: https://bcregistry-demo.apigee.io/docs/mhrproxy/1/types/owner

Screen Shot 2022-08-24 at 3 03 27 PM


The spec states that phone numbers are strings so i'm confused.
Was there a request for the spec to change as well?

Where does the requirement for number only come from?

cameron-eyds: 🤔 
Sorry, it's a bit misleading, the spec is saying it is a string but formatted only containing digits so i suppose that is where the confusion might be coming from. 
Can you confirm that the end result of this fields INPUT is as the example is above? 
If so, than all good 👍

ChasBoucher: The end result is a string..

cameron-eyds (approved):
Thanks for the updates! 👍 
@ChasBoucher Looks like some minor linting issues to fix up. 

Screen Shot 2022-08-25 at 12 24 52 PM

ppr-ui/src/composables/useInputRules.ts
ChasBoucher: I have made the change to: "Enter numbers only".  This is base on UX Design comments on ticket

ChasBoucher: Changed to "Enter numbers only".  This is based on UX review

dimak1 (approved):

PR #921, opened 2022-08-26
update code to surpass some bug

garyfan1:
*Issue #:* /bcgov/entity#13215

*Description of changes:*
- Api bug where it doesn't accept `personName` in `submittingParty`, replacing with `businessName` for now.
- Street Number exceeding 6 characters causes error, confirm with design team how to tackle this.


cameron-eyds:
@garyfan1 Are you sure we need this in? 
If the intention is to fix this bug on the api side, i believe we shouldn't commit this and wait for the bug to be fixed. For QA in the meantime, there is the option of going with a Business Submitting Party as a work around. 
Thoughts on this?

garyfan1:
Closing this PR because this fix is discussed unnecessary.

PR #922, opened 2022-08-29
13222 Check search result if MHR Number Search

ChasBoucher:
*Issue #:* /bcgov/13222

*Description of changes:*

1.  When user does an MHR Number Search only, there will be 1 result and it will be selected.. 

### DIFF
--- a/ppr-ui/src/components/tables/mhr/SearchedResults.vue
+++ b/ppr-ui/src/components/tables/mhr/SearchedResults.vue
@@ -342,9 +342,8 @@ export default defineComponent({
       })
       localState.totalResultsLength = resp.totalResultsSize
       if (localState.searchType === UISearchTypes.MHR_NUMBER && localState.totalResultsLength === 1) {
-        const val = true
         // Select search result if an MHR Number Search and search results equals 1.
-        localState.results = localState.results.map(result => ({ ...result, selected: val }))
+        localState.results = localState.results.map(result => ({ ...result, selected: true }))
       }
       const date = new Date(resp.searchDateTime)
       localState.searchTime = pacificDate(date)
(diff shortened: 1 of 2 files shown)


ppr-ui/src/components/tables/mhr/SearchedResults.vue
```diff
@@ -341,6 +341,11 @@ export default defineComponent({
         return result.includeLienInfo !== true ? { ...result, includeLienInfo: false } : result
       })
       localState.totalResultsLength = resp.totalResultsSize
+      if (localState.searchType === UISearchTypes.MHR_NUMBER && localState.totalResultsLength === 1) {
+        const val = true
+        // Select search result if an MHR Number Search and search results equals 1.
+        localState.results = localState.results.map(result => ({ ...result, selected: val }))
```
cameron-eyds: if `val` isn't use anywhere else in its scope, does it make sense to apply the value directly to the mapped property?
ie `selected: true`

cameron-eyds (approved):
Looks good just one minor comment 👍

ppr-ui/src/components/tables/mhr/SearchedResults.vue
ChasBoucher: Good idea... I will make the change...

ppr-ui/src/components/tables/mhr/SearchedResults.vue
```diff
@@ -341,6 +341,10 @@ export default defineComponent({
         return result.includeLienInfo !== true ? { ...result, includeLienInfo: false } : result
       })
       localState.totalResultsLength = resp.totalResultsSize
+      if (localState.searchType === UISearchTypes.MHR_NUMBER && localState.totalResultsLength === 1) {
+        // Select search result if an MHR Number Search and search results equals 1.
+        localState.results = localState.results.map(result => ({ ...result, selected: true }))
```
dimak1, line 346: If there is only one result, do we even need to use map() ?

dimak1 (approved):
Approved with one comment.

doug-lovett (approved):

ppr-ui/src/components/tables/mhr/SearchedResults.vue
ChasBoucher: I did this to keep with existing logic... It may work but I have not tested..

PR #927, opened 2022-08-30
"Include lien info for all" state kept when navigating back from Summary page

garyfan1:
*Issue #:* /bcgov/entity#13383

*Description of changes:*
update checkbox behavior, add state to state model

### DIFF
--- a/ppr-ui/src/components/tables/mhr/SearchedResults.vue
+++ b/ppr-ui/src/components/tables/mhr/SearchedResults.vue
@@ -220,12 +220,23 @@ export default defineComponent({
       getManufacturedHomeSearchResults,
       getFolioOrReferenceNumber,
       getSearchedType,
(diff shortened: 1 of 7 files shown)


dimak1 (approved):

cameron-eyds (approved):
Code looks good!

If a user selects all and sets that to the app state, will it persist or does it reset on a new search? Was my only thought!

garyfan1:
I thought it would retain, turns out it resets for some reason.

cameron-eyds:
> I thought it would retain, turns out it resets for some reason.

Well i don't think you want it to persist to a new search criteria, however you do want it to persist between Select and Review pages. Is that correct?

garyfan1:
Yes that is correct, and that's its behavior rn.

doug-lovett (approved):
Very nice Gary.

PR #930, opened 2022-09-01
LTSA: Pid Summary Request

cameron-eyds:
*Issue #:* /bcgov/entity#12966

*Description of changes:*
* LTSA Configurations and Helper
* Updated PID functionality to validate our GET request
* Updated Validations on OTHER options, including pid
* Updated PID formatting to adhere to schema and for display format.
* Pid error handling and Dialog.

### DIFF
--- a/ppr-ui/src/components/common/PidNumber.vue
+++ b/ppr-ui/src/components/common/PidNumber.vue
@@ -72,6 +72,7 @@ import { useInputRules } from '@/composables'
 import { ltsaSummary } from '@/utils/ltsa-api-helper'
 import { BaseDialog } from '@/components/dialogs'
 import { pidNotFoundDialog } from '@/resources/dialogOptions'
+import { TitleSummariesIF } from '@/interfaces/ltsa-api-interfaces'
 /* eslint-enable no-unused-vars */
 
 export default defineComponent({
@@ -122,20 +123,15 @@ export default defineComponent({
     const dialogRetry = async (retry: boolean): Promise<void> => {
       localState.showNotFoundDialog = false
       if (retry) {
-        await validatePid()
+        await validatePid() ? emitPid() : localState.showNotFoundDialog = true
       } else clearPid()
     }
 
-    const validatePid = async (): Promise<void> => {
+    const validatePid = async (): Promise<boolean> => {
       localState.enablePidLoader = true
-      const ltsaResponse = await ltsaSummary(localState.pidNumber)
+      const { titleSummaries } = await ltsaSummary(localState.pidNumber) as TitleSummariesIF
       localState.enablePidLoader = false
-
-      if (ltsaResponse?.error) {
-        localState.showNotFoundDialog = true
-      } else {
-        emitPid()
-      }
+      return titleSummaries?.length > 0
     }
 
     const clearPid = (): void => {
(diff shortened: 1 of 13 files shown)


ppr-ui/src/components/common/PidNumber.vue
```diff
@@ -106,6 +119,31 @@ export default defineComponent({
...
     }
 
+    const dialogRetry = async (retry: boolean): Promise<void> => {
+      localState.showNotFoundDialog = false
+      if (retry) {
+        await validatePid()
+      } else clearPid()
+    }
+
+    const validatePid = async (): Promise<void> => {
```
dimak1: Since it does more than just validation, should probably be validateAndEmitPid()

ppr-ui/src/utils/ltsa-api-helper.ts
```diff
@@ -0,0 +1,32 @@
...
+      const data = response?.data
+      if (!data) {
+        throw new Error('Invalid API response')
+      }
+      return data
+    })
+    .catch(error => {
+      return {
+        error: {
+          category: ErrorCategories.LTSA_REQUEST,
```
dimak1, line 27: Are we actually doing anything with error details? Do we have some built-in handling (i.e. log entry, show on screen, console etc.)?

dimak1 (approved):
Approved with some comments/questions. Thanks!

ppr-ui/src/components/common/PidNumber.vue
cameron-eyds: Hmm not sure i agree, we've already pulled out the `emitPid()` into its own functionality. 
If you feel strongly about this, i could return a value and subsequently call `emit()` depending on the result?

dimak1: Yea, that was something I would suggest as well. At first, I thought `validatePid()` would just do one action, return true/false, but then I saw it also calls emit. So hence the comment.

ppr-ui/src/utils/ltsa-api-helper.ts
cameron-eyds: Nope, not with these values directly, in this context.
It could be useful if we wanted to pass the statusCode and/or message to an error dialog. 

In this case, we are simply prompting a generic dialog, design doesn't call for specific messaging. 
I have to imagine this IS happening in other scenarios however, since this is the approach that is more or less consistent with other request helpers.

ppr-ui/src/components/common/PidNumber.vue
cameron-eyds: Sure, will update.

ChasBoucher (approved):
Looks Good.  Nice to see how errors are handled..

cameron-eyds:
@dimak1 @ChasBoucher 
Thanks for the review, updated as per Dima's recommendations, also created/updated LTSA Title Summaries types to help define the response from this verification request. 

I imagine we will be doing more with LTSA down the road so this is a start 👍

ppr-ui/src/components/common/PidNumber.vue
cameron-eyds: Updated!

doug-lovett (approved):
LGTM.

dimak1:
> @dimak1 @ChasBoucher Thanks for the review, updated as per Dima's recommendations, also created/updated LTSA Title Summaries types to help define the response from this verification request.
> 
> I imagine we will be doing more with LTSA down the road so this is a start 👍

Thanks, looks good!

PR #933, opened 2022-09-06
Staff Payment, validation, submission

dimak1:
*Issue #:* /bcgov/entity#13260

*Description of changes:*

- Use Staff Payment component on the MHR Review page.
- Validation and Formatting
- Submission of the payment info as a request query parameter string

#### Please Note:

Some of the logic does not make sense to me (maybe there is a business rule that explains it), but when clicking the **Priority checkbox**, the payment data is not being emitted (via `emitStaffPaymentData`) and therefore store model is missing Payment Option and Slip Number. Screenshot of the logic below - `@bcrs-shared-components/staff-payment/StaffPayment.vue`, line 68.

Screen Shot 2022-09-05 at 7 30 43 PM

### DIFF
--- a/ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
+++ b/ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
@@ -1,16 +1,11 @@
 import { useGetters } from 'vuex-composition-helpers'
 import {
-  getMhrAttentionReferenceNum,
-  getMhrRegistrationDocumentId,
-  getMhrRegistrationHomeDescription, getMhrRegistrationHomeOwnerGroups, getMhrRegistrationLocation,
-  getMhrRegistrationSubmittingParty
-} from '@/store/getters'
-import {
   MhrRegistrationDescriptionIF,
   MhrRegistrationHomeLocationIF,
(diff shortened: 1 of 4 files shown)


ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -121,6 +135,7 @@ export const useNewMhrRegistration = () => {
   }
 
   return {
-    buildApiData
+    buildApiData,
+    staffPayment: parseStaffPayment
```
cameron-eyds: This is a little confusing to me, is this the property or the function expression? 
The type indicates the expression, so why not return `parseStaffPayment` ?

cameron-eyds (approved):
Looks good, just the one question!

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
dimak1: We are returning a function, that will be called later. Just used it for naming. Will update.

dimak1:
@cameron-freshworks agree, the naming wasn't clear. Updated!

PR #937, opened 2022-09-07
Fix exempt flag setting incorrect reg status.

doug-lovett:

*Issue #:* /bcgov/entity#13421

*Description of changes:*
Ignore exempt flag to correct search results returning incorrect registration status: EXEMPT instead of HISTORICAL.


cameron-eyds (approved):
Looks good 👍

PR #938, opened 2022-09-07
New reg add legacy db limitation validation rules.

doug-lovett:

*Issue #:* /bcgov/entity#13476

*Description of changes:*
- Handle legacy db table street number length limitation.
- Add new reg character set validation on names (same validation as PPR).


ChasBoucher (approved):
Looks good.  Really good to see how the location address is being handled in API.

PR #939, opened 2022-09-07
13362 Fix no OwnerName showing if last name only search

ChasBoucher:
*Issue #:* /bcgov/entity#13362

*Description of changes:*

- Fix no OwnerName showing if Owner last name only search.
- Remove unneeded code.  Caused validation error.
- 

### DIFF
--- a/ppr-ui/src/components/search/SearchBar.vue
+++ b/ppr-ui/src/components/search/SearchBar.vue
@@ -91,7 +91,6 @@
               :hint="searchHintFirst"
               persistent-hint
               :placeholder="optionFirst"
-              v-validate=" isStaffBcolReg ? '' : 'required' "
               v-model="searchValueFirst"
               @keypress.enter="searchCheck()"
           />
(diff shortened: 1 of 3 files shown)


doug-lovett (approved):

ppr-ui/src/components/search/SearchBar.vue
```diff
@@ -91,7 +91,6 @@
               :hint="searchHintFirst"
               persistent-hint
               :placeholder="optionFirst"
-              v-validate=" isStaffBcolReg ? '' : 'required' "
```
cameron-eyds, line 94: Does this make it so it's not `required` for PPR as well?

ChasBoucher: I don't believe so.  It was some code I put in to test..

cameron-eyds: Would be good to verify as this component is used for both search types (ppr/mhr) i believe, thanks!

cameron-eyds (approved):

PR #955, opened 2022-09-20
Made search result table headers adaptive

owillborn:
*Issue #:* /bcgov/entity###

*Description of changes:*
Added function to check if the result of an MHR search should have "Organization" or "Owner" headers. Made the table header slots adaptive

### DIFF
--- a/ppr-ui/src/components/tables/mhr/SearchedResults.vue
+++ b/ppr-ui/src/components/tables/mhr/SearchedResults.vue
@@ -100,7 +100,7 @@
 
           <template  v-else v-slot:[headerSearchTypeSlot]>
             <span v-if="isOwnerOrOrgSearch" class="pl-8">
-              {{ personOrOrgLabel }} Name
+              {{ ownerOrOrgHeader }} Name
             </span>
             <span v-else class="pl-8">{{ headerSlotLabel }}</span>
           </template>
@@ -231,7 +231,7 @@ import {
 import { BaseHeaderIF, ManufacturedHomeSearchResultIF } from '@/interfaces' // eslint-disable-line no-unused-vars
 import { FolioNumber } from '@/components/common'
 import { pacificDate } from '@/utils'
-import { APIMHRMapSearchTypes, RouteNames, UIMHRSearchTypes, UIMHRSearchTypeValues } from '@/enums'
+import { RouteNames, UIMHRSearchTypes, UIMHRSearchTypeValues } from '@/enums'
 import { cloneDeep } from 'lodash'
 
 export default defineComponent({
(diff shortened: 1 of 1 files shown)


ppr-ui/src/components/tables/mhr/SearchedResults.vue
```diff
@@ -321,6 +325,12 @@ export default defineComponent({
           ? 'Organization'
           : 'Owner'
       }),
+      ownerOrOrgHeader: computed((): string => {
+        const found = getManufacturedHomeSearchResults.value.results
+        if (found) {
+          return found[0]?.organizationName ? 'Organization' : 'Owner'
+        } else return ''
+      }),
```
cameron-eyds: This does seem very similar to `personOrOrgLabel`. 
Do we NEED both computed values?
Can we get away with one?

owillborn: I found that personOrOrgLabel never returned 'Organization' and was planning to remove it, I guess I forgot whoops!

cameron-eyds: Sounds good, thanks!

cameron-eyds (approved):
Looks good!

doug-lovett (approved):
Very nice.

ppr-ui/src/components/tables/mhr/SearchedResults.vue
```diff
@@ -316,10 +320,11 @@ export default defineComponent({
...
       }),
-      personOrOrgLabel: computed((): string => {
-        return getManufacturedHomeSearchResults.value?.searchQuery.type === APIMHRMapSearchTypes.MHRORGANIZATION_NAME
-          ? 'Organization'
-          : 'Owner'
+      ownerOrOrgHeader: computed((): string => {
+        const found = getManufacturedHomeSearchResults.value.results
+        if (found) {
+          return found[0]?.organizationName ? 'Organization' : 'Owner'
+        } else return ''
```
dimak1, line 327: So if no results are found the header will be just 'Name' or 'Status'?

ChasBoucher (approved):
Other than Dima's Comments. Looks Good

dimak1:
@owillborn looks good, just one comment

ppr-ui/src/components/tables/mhr/SearchedResults.vue
owillborn: @dimak1 If there are no results the table will not be displayed, therefore no headers to display... would it be better to return undefined instead of an empty string in that case?

dimak1: In this case, maybe we can simplify to
`return getManufacturedHomeSearchResults.value?.results[0]?.organizationName ? 'Organization' : 'Owner'` since it won't matter what is showing.

dimak1 (approved):
Approved with a comment.

ppr-ui/src/components/tables/mhr/SearchedResults.vue
owillborn: I tried to write it in shorter hand but I can't seem to get it to work when there are no results with shorthand...

PR #958, opened 2022-09-20
changed validation to be false upon submission

owillborn:
*Issue #:* /bcgov/entity###

*Description of changes:*
- Set stepper validation to false upon MHR submission

### DIFF
--- a/ppr-ui/src/views/newMhrRegistration/MhrRegistration.vue
+++ b/ppr-ui/src/views/newMhrRegistration/MhrRegistration.vue
@@ -193,6 +193,10 @@ export default defineComponent({
         // Submit Filing
         localState.submitting = true
         const mhrSubmission = await submitMhrRegistration(buildApiData(), parseStaffPayment())
+        setValidation(MhrSectVal.YOUR_HOME_VALID, MhrCompVal.VALIDATE_APP, false)
+        setValidation(MhrSectVal.SUBMITTING_PARTY_VALID, MhrCompVal.VALIDATE_APP, false) 
+        setValidation(MhrSectVal.HOME_OWNERS_VALID, MhrCompVal.VALIDATE_APP, false) 
+        setValidation(MhrSectVal.LOCATION_VALID , MhrCompVal.VALIDATE_APP, false) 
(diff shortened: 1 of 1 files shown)


ppr-ui/src/views/newMhrRegistration/MhrRegistration.vue
```diff
@@ -194,6 +194,10 @@ export default defineComponent({
         // Submit Filing
         localState.submitting = true
         const mhrSubmission = await submitMhrRegistration(buildApiData(), parseStaffPayment())
+        setValidation(MhrSectVal.YOUR_HOME_VALID, MhrCompVal.VALIDATE_APP, false)
+        setValidation(MhrSectVal.SUBMITTING_PARTY_VALID, MhrCompVal.VALIDATE_APP, false) 
+        setValidation(MhrSectVal.HOME_OWNERS_VALID, MhrCompVal.VALIDATE_APP, false) 
+        setValidation(MhrSectVal.LOCATION_VALID , MhrCompVal.VALIDATE_APP, false)
```
cameron-eyds: `MhrCompVal.VALIDATE_APP` is a single flag in the `reviewConfirmValid` validation block. 
Why this might appear to work is you are creating a new property that is set to `false` in each steps validation block.
The issue is that this new property created persists and each step will be unable to be `valid` in subsequent Registrations. 

You are on the right track though! 
Why this ticket is estimated a bit higher i think..
1) Ensure our `initNewMhr` is actually setting the mhr store values to a default state
2A) If so, look into Step 2 and 3 components as to potentially why they are not setting the validations to `false` when there is no data present.
OR
2B) Re-initialize our Validation State using an approach similar to `initNewMhr()`.  ie an action/mutation that resets it to default values.

cameron-eyds: I would personally go with 2B i think, so you are not relying on any components (now or future state).

owillborn: I'll take another look at it, thanks Cameron!

PR #961, opened 2022-09-23
Manually add MHR to the registration table

dimak1:
*Issue #:* /bcgov/entity#13136

*Description of changes:*
- Update validation for MHR number retrieval
- GET and POST requests to add MHR to the registration table
- Based on the PPR flow

### Please Note:

This functionality needs to be tested in DEV because of the following:
- Searching for a registration number that's _already in the table_ returns a flag `inUserList: false` as if the record does not exists in the table
- `CURRENT_ACCOUNT` is not present in Session Storage, but it is needed for the POST request call to manually add MHR to the table
- CC: @doug-lovett 

### Screenshot of Registration Found dialog with MHR related content
 
Screen Shot 2022-09-22 at 4 56 20 PM

### DIFF
--- a/ppr-ui/package-lock.json
+++ b/ppr-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ppr-ui",
-  "version": "0.3.29",
+  "version": "0.3.30",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "ppr-ui",
-      "version": "0.3.29",
+      "version": "0.3.30",
       "dependencies": {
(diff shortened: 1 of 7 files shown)


ChasBoucher (approved):
Code looks good.
I realize that there will be more dev required once we get some feedback..

cameron-eyds (approved):
Looks good, could you include a version bump!

doug-lovett:
For the comment:
Searching for a registration number that's already in the table returns a flag inUserList: false as if the record does not exists in the table
This functionality is working for mhr numbers that belong to another account and are legacy registrations. A change so new registrations that belong the current account are also checked will be included in the next api PR.

PR #963, opened 2022-09-26
Fix clearing of Submitting Party address fields

dimak1:
*Issue #:* /bcgov/entity#13637

*Description of changes:*

Fixes the issue when look-up address component clears the fields of base address component:

- Any time when Country is changed, the address fields are cleared. So when we manually enter the address and then do a lookup and select a business from dropdown, the change of the Country clears the address fields.
- It does not happen when manually entered Country is the same as the looked up business Country. Because the Country does not change, the address is pre-filled correctly.

### DIFF
--- a/ppr-ui/src/components/parties/party/PartyAutocomplete.vue
+++ b/ppr-ui/src/components/parties/party/PartyAutocomplete.vue
@@ -156,6 +156,8 @@ export default defineComponent({
         setRegisteringParty(newParty)
       } else if (props.isMhrPartySearch) {
         localState.selectedCode = newParty.code
+        // pre-set country code to prevent clearing of base-address component fields (bug 13637)
+        setMhrSubmittingParty({ key: 'address.country', value: newParty.address.country })
         // Set submitting party data to store
         for (const [key, value] of Object.entries(newParty)) {
           setMhrSubmittingParty({ key, value })
--- a/ppr-ui/src/store/mutations/mutations-model.ts
+++ b/ppr-ui/src/store/mutations/mutations-model.ts
@@ -32,6 +32,7 @@ import {
   MhRegistrationSummaryIF
 } from '@/interfaces'
 import { StaffPaymentIF } from '@bcrs-shared-components/interfaces'
+import { set } from 'lodash'
 
 export const mutateAccountProductSubscribtion = (
   state: StateIF, productSubscriptions: AccountProductSubscriptionIF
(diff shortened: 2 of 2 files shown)


ppr-ui/src/store/mutations/mutations-model.ts
```diff
@@ -369,7 +370,7 @@ export const mutateMhrBaseInformation = (state: StateIF, { key, value }) => {
 }
 
 export const mutateMhrSubmittingParty = (state: StateIF, { key, value }) => {
-  state.stateModel.mhrRegistration.submittingParty[key] = value
+  set(state.stateModel.mhrRegistration.submittingParty, key, value)
```
cameron-eyds, line 373: What makes this mutation unique from the rest?

ChasBoucher (approved):
Looks Good.

ppr-ui/src/components/parties/party/PartyAutocomplete.vue
```diff
@@ -156,6 +156,8 @@ export default defineComponent({
         setRegisteringParty(newParty)
       } else if (props.isMhrPartySearch) {
         localState.selectedCode = newParty.code
+        // pre-set country code to prevent clearing of base-address component fields (bug 13637)
+        setMhrSubmittingParty({ key: 'address.country', value: newParty.address.country })
```
cameron-eyds, line 160: Cool, seems straight forward?

ChasBoucher: I agree...

ppr-ui/src/store/mutations/mutations-model.ts
dimak1: Using lodash set() we could set nested properties for `submittingParty`, e.g. `address.country`

ppr-ui/src/components/parties/party/PartyAutocomplete.vue
dimak1: Yes, we only want to set the `address.county` here, so the fields won't get cleared when we set the whole `address`.

ppr-ui/src/store/mutations/mutations-model.ts
cameron-eyds: So this is required to set `address.country` ? 
Fair enough if so! 

I guess there are pros and cons to trying to minimize the actions and mutations, so it seems like this is a good solution/alternative 👍

cameron-eyds (approved):
Looks good!

ppr-ui/src/store/mutations/mutations-model.ts
dimak1: Yea this is needed, otherwise we won't be able to set nested props, like `country` only.

PR #965, opened 2022-09-27
Transfer Components + Api Integration

cameron-eyds:
*Issue #:* /bcgov/entity#13609

*Description of changes:*
* Updates HomeOwners components to be compatible with Transfer Flow
* Implements basic API integration with new Transfer Endpoint
* Feedback updates with regards to Breadcrumb and header listed in ticket

*Notes:* - Some data (such as Submitting Party address) is not collected by the UI and subsequently mocked in the request to satisfy the Schema

### DIFF
--- a/ppr-ui/src/utils/mhr-api-helper.ts
+++ b/ppr-ui/src/utils/mhr-api-helper.ts
@@ -282,3 +282,24 @@ export async function validateDocumentID (documentId: string) {
     }
   }
 }
+
+export async function submitMhrTransfer (payloadData, mhrNumber) {
+  const url = sessionStorage.getItem('MHR_API_URL')
+  const config = { baseURL: url, headers: { Accept: 'application/json' } }
+
+  try {
+    const result = await axios.post(`transfers/${mhrNumber}`, payloadData, config)
+    if (!result?.data) {
+      throw new Error('Invalid API response')
+    }
+    return result.data
+  } catch (error) {
+    return {
+      error: {
+        category: ErrorCategories.REGISTRATION_TRANSFER,
+        statusCode: error?.response?.status || StatusCodes.NOT_FOUND,
+        msg: error?.response?.data?.errorMesage || 'Unknown Error'
+      }
+    }
+  }
+}
--- a/ppr-ui/package-lock.json
+++ b/ppr-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ppr-ui",
-  "version": "0.3.29",
+  "version": "0.3.30",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "ppr-ui",
-      "version": "0.3.29",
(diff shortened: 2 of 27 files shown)


ChasBoucher (approved):
Looks Good.

ppr-ui/src/utils/mhr-api-helper.ts
```diff
@@ -282,3 +282,24 @@ export async function validateDocumentID (documentId: string) {
     }
   }
 }
+
+export async function submitMhrTransfer (payloadData, mhrNumber) {
+  const url = sessionStorage.getItem('MHR_API_URL')
+  const config = { baseURL: url, headers: { Accept: 'application/json' } }
```
dimak1, line 288: For the POST `/mhr/api/v1/transfers/{mhrNumber}` api call, **Account-Id** is required in the headers.

dimak1 (approved):
Nice and clean implementation! Just one comment for the POST call.

ppr-ui/src/utils/mhr-api-helper.ts
cameron-eyds: Oh! I didn't see, do you have a reference?

dimak1: Sent you via RocketChat.

doug-lovett: Following the ppr api I think there is an axios request setup that tries to add the account id in the header for all requests?

doug-lovett (approved):
Looks good, just a note that using a group id of 1 only works if the MH registration has no previous transfers. The group id increments as changes are made. I suggest testing  in this sprint with a new sole owner registration.

ppr-ui/src/utils/mhr-api-helper.ts
cameron-eyds: @dimak1 Sorry, to follow up on this, the `account-id` IS included. 
It's handled by the axios interceptor `axios-ppr.ts`. It is conditionally included in header config when there is a `currentAccount` present. ie any non-staff, for the most part. 

Good eye though, thanks for noting this!

cameron-eyds:
> Looks good, just a note that using a group id of 1 only works if the MH registration has no previous transfers. The group id increments as changes are made. I suggest testing in this sprint with a new sole owner registration.



> Looks good, just a note that using a group id of 1 only works if the MH registration has no previous transfers. The group id increments as changes are made. I suggest testing in this sprint with a new sole owner registration.

@doug-lovett Good call, thanks will do!
 I hadn't done extensive testing with the groups aspect of Home Owners yet, so i'm sure we can flesh this out further in the next iteration.

ppr-ui/src/utils/mhr-api-helper.ts
cameron-eyds: @doug-lovett Oh, you beat me to it 😆 
Thank you!

PR #975, opened 2022-10-03
13670 mhr UI update location types

ChasBoucher:
*Issue #:* /bcgov/entity#13670

*Description of changes:*

- Update MHR location types enum to match api requirements.

### DIFF
--- a/ppr-ui/src/enums/homeLocationTypes.ts
+++ b/ppr-ui/src/enums/homeLocationTypes.ts
@@ -1,8 +1,8 @@
 export enum HomeLocationTypes {
-  LOT = 'lot',
-  HOME_PARK = 'homePark',
-  OTHER_LAND = 'otherLand',
-  OTHER_RESERVE = 'reserve',
-  OTHER_STRATA = 'strata',
-  OTHER_TYPE = 'otherType'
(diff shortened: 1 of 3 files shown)


cameron-eyds (commented):
Do we need to update the watchers in `HomeLocationType` as well?

*Edit - After a quick look, the watchers do use the values vs the enum itself. So will need to update there as well i believe.

dimak1 (approved):

owillborn:
Aside from Cameron's comment, looks good to me!

doug-lovett (approved):
looks good

owillborn (approved):
Looks solid

ChasBoucher:
The watchers seem fine.  They evaluate to the correct values and the store is being updated correctly.

cameron-eyds:
> The watchers seem fine. They evaluate to the correct values and the store is being updated correctly.

@ChasBoucher please have another look at the watcher on line 261 in HomeLocationType.
It references the string value of the enum `otherType` but you've changed the value in the enum here to `other`. 
Maybe i'm just being over cautious, it is most likely just due to a property name being similar name as the value :)
But if you've verified than all good!

cameron-eyds (approved):

PR #988, opened 2022-10-12
Part one: Transfer Group handling + Bug Fixes

cameron-eyds:
*Issue #:* /bcgov/entity#13820

*Description of changes:*
* Fix Breadcrumb MHR Number Label bug
* Updated group handling/parsing logic on existing groups.
* Added Confirm Remove dialog for MHR

*Note:*
- Pending work includes finessing the existing group specific data into the components, ie interest & numerator. 
- This will be in a part 2 commit

### DIFF
--- a/ppr-ui/src/views/mhrInformation/MhrInformation.vue
+++ b/ppr-ui/src/views/mhrInformation/MhrInformation.vue
@@ -32,6 +32,7 @@
                   isMhrTransfer
                   isReadonlyTable
                   :homeOwners="getMhrTransferHomeOwners"
+                  :currentHomeOwners="getMhrTransferCurrentHomeOwners"
                 />
               </template>
 
@@ -95,9 +96,9 @@ export default defineComponent({
   },
   setup (props, context) {
     const {
-      getMhrTransferHomeOwners, getMhrInformation
+      getMhrTransferHomeOwners, getMhrInformation, getMhrTransferCurrentHomeOwners
     } = useGetters<any>([
-      'getMhrTransferHomeOwners', 'getMhrInformation'
+      'getMhrTransferHomeOwners', 'getMhrInformation', 'getMhrTransferCurrentHomeOwners'
     ])
 
     const {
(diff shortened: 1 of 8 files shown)


ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -162,10 +164,22 @@ export default defineComponent({
...
     const parseMhrInformation = async (): Promise<void> => {
-      // Future state to parse all relevant MHR Information
+      await parseCurrentOwnerGroups()
+    }
+
+    const parseCurrentOwnerGroups = async (): Promise<void> => {
       const { data } = await fetchMhRegistration(getMhrInformation.value.mhrNumber)
       const currentOwnerGroups = data?.ownerGroups || [] // Safety check. Should always have ownerGroups
+      // Create an ID to each individual owner for UI Tracking
+      currentOwnerGroups.forEach(ownerGroup => {
```
dimak1, line 176: So this is purely for UI? `MhrRegistrationHomeOwnerIF` has id but it's optional, maybe we should make it required, so we don't have to set it here?

dimak1 (approved):
Approved with a comment.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
cameron-eyds: Required as in filed with an ID, preserved in the DB and returned?
Via MHRegistrations?

cameron-eyds: I'm not opposed to that but it might not be something the API would require/care about 🤔

cameron-eyds: But yes, it does seem that it SHOULD be required, for UI at least. Since we do require it for functionality.

doug-lovett (approved):
Looks good again.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
dimak1: Resolving after discussion.

dimak1:
@cameron-freshworks is this for #13820 right? It's not mentioned in the description.

cameron-eyds:
> @cameron-freshworks is this for #13820 right? It's not mentioned in the description.

Sorry, updated comment to include ticket link 👍

PR #990, opened 2022-10-14
13707 mhr transfer unsaved changes prompt

ChasBoucher:
*Issue #:* /bcgov/entity#13707

*Description of changes:*
- Add Unsaved changes prompt for canceling without saving changes.


ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
```diff
@@ -425,7 +427,7 @@ export default defineComponent({
           delete localState.group.interestTotal
           delete localState.group.tenancySpecified
         }
-
+        if (oldOwner !== localState.owner) setUnsavedChanges(true)
```
cameron-eyds: What about just setting this in the mutation?

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -192,14 +206,19 @@ export default defineComponent({
       setMhrTransferCurrentHomeOwnerGroups(currentOwnerGroups)
     }
 
+    const cancel = async (): Promise<void> => {
+      if (hasUnsavedChanges.value === true) localState.showCancelDialog = true
+      else goToDash()
+    }
```
cameron-eyds: Could this check be included in goToDash vs a separate method?

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -124,6 +132,9 @@ export default defineComponent({
       setShowGroups
     } = useHomeOwners()
 
+    const options = unsavedChangesDialog
+    var showCancelDialog = false
```
cameron-eyds: Should probably use const.
That being said, doesn't this belong in the localState?

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -148,7 +159,9 @@ export default defineComponent({
       }),
       reviewConfirmText: computed((): string => {
         return localState.isReviewMode ? 'Register Changes and Pay' : 'Review and Confirm'
-      })
+      }),
+      options,
+      showCancelDialog
```
cameron-eyds: Oh, you did define in localState. Why not the values as well?

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
cameron-eyds: Just concerned this might apply to registrations as well. Also at this point, the user hasn't submit any change right?

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: I checked onMounted() and also goToReview() where goToDash is being called and if hasUnsavedChanges.value is true than the dialog will display.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: Now initialized in localState..

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: good point...

dimak1:
@ChasBoucher there are some unresolved comments from Cameron...

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
ChasBoucher: I have added a  check to ensure that check is only on Mhr transfer,

cameron-eyds: Seems a little inconsistent with the other approaches, where it is just set in the mutation when a change is made. 
Is that not possible here?

ppr-ui/src/views/mhrInformation/MhrInformation.vue
cameron-eyds: I understand the condition.
Referring specifically here to the line `if (hasUnsavedChanges.value === true) localState.showCancelDialog = true` 
Couldn't this just go in the `goToDash` function?

cameron-eyds: You are creating a new function that does the very same thing the existing one does but with a condition that could just intercept in our `goToDash` funtion.

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
cameron-eyds: Also, there is most likely a root emit required to catch the users who use the browser back button vs just hitting cancel.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: Made the change and put the owner logic in goToDash.. Works...

cameron-eyds: Sounds good, thanks!

cameron-eyds (approved):
Looks good

cameron-eyds:
@ChasBoucher Just some minor linting to fix up

PR #995, opened 2022-10-18
Home-Owners-Removal

cameron-eyds:
*Issue #:* /bcgov/entity#13877

*Description of changes:*
* Implements tracking for Owner changes
* Functionality to remove Existing owners and apply them to payload. 
* Misc updates and clean up. 
* Minor testing updates.

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
@@ -164,7 +164,7 @@
             </template>
           </td>
         </tr>
-        <tr v-else-if="!isMhrTransfer">
+        <tr>
           <td :colspan="4" class="py-1">
             <div
               v-if="showGroups"
@@ -173,7 +173,7 @@
             >
               Group must contain at least one owner
             </div>
-            <div v-else class="my-6 text-center" data-test-id="no-owners-mgs">
+            <div v-else-if="!isMhrTransfer" class="my-6 text-center" data-test-id="no-owners-mgs">
               No owners added yet
             </div>
           </td>
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
@@ -222,7 +222,7 @@ export default defineComponent({
       hasEmptyGroup,
       hasMinimumGroups,
       editHomeOwner
-    } = useHomeOwners()
+    } = useHomeOwners(props.isMhrTransfer)
 
     const localState = reactive({
       currentlyEditingHomeOwnerId: -1,
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
@@ -164,7 +164,7 @@
             </template>
           </td>
         </tr>
-        <tr>
+        <tr v-else-if="!isMhrTransfer">
           <td :colspan="4" class="py-1">
             <div
               v-if="showGroups"
@@ -173,7 +173,7 @@
             >
               Group must contain at least one owner
             </div>
-            <div v-else-if="!isMhrTransfer" class="my-6 text-center" data-test-id="no-owners-mgs">
+            <div v-else class="my-6 text-center" data-test-id="no-owners-mgs">
               No owners added yet
             </div>
           </td>
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
@@ -40,90 +40,137 @@
 
         <tr v-else-if="row.item.id" :key="row.item.id" class="owner-info" :data-test-id="`owner-info-${row.item.id}`">
           <td class="owner-name">
-            <div v-if="row.item.individualName" class="owner-icon-name">
-              <v-icon class="mr-2">mdi-account</v-icon>
-              <div class="font-weight-bold">
-                {{ row.item.individualName.first }}
-                {{ row.item.individualName.middle }}
-                {{ row.item.individualName.last }}
+            <div :class="{'removed-owner': isRemovedHomeOwner(row.item)}">
+              <div v-if="row.item.individualName" class="owner-icon-name">
+                <v-icon class="mr-2">mdi-account</v-icon>
+                <div class="font-weight-bold">
+                  {{ row.item.individualName.first }}
+                  {{ row.item.individualName.middle }}
+                  {{ row.item.individualName.last }}
+                </div>
+              </div>
+              <div v-else class="owner-icon-name">
+                <v-icon class="mr-2">mdi-domain</v-icon>
+                <div class="font-weight-bold">
+                  {{ row.item.organizationName }}
+                </div>
+              </div>
+              <div v-if="row.item.suffix" class="suffix">
+                {{ row.item.suffix }}
               </div>
             </div>
-            <div v-else class="owner-icon-name">
-              <v-icon class="mr-2">mdi-domain</v-icon>
-              <div class="font-weight-bold">
-                {{ row.item.organizationName }}
-              </div>
-            </div>
-            <div v-if="row.item.suffix" class="suffix">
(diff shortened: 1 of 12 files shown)


ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -40,91 +40,138 @@
 
         <tr v-else-if="row.item.id" :key="row.item.id" class="owner-info" :data-test-id="`owner-info-${row.item.id}`">
           <td class="owner-name">
-            <div v-if="row.item.individualName" class="owner-icon-name">
-              <v-icon class="mr-2">mdi-account</v-icon>
-              <div class="font-weight-bold">
-                {{ row.item.individualName.first }}
-                {{ row.item.individualName.middle }}
-                {{ row.item.individualName.last }}
+            <div :class="{'removed-owner': isRemovedHomeOwner(row.item)}">
```
dimak1, line 43: Can we apply `removed-owner` class to the whole row instead? Is it because of `Delete` badge that needs to be in full colour?

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -40,91 +40,138 @@
...
-          <td :colspan="4" class="py-1">
+          <td v-if="showGroups" :colspan="4" class="py-1">
             <div
-              v-if="showGroups"
               class="error-text my-6 text-center"
               :data-test-id="`no-owners-msg-group-${homeOwners.indexOf(row.item)}`"
             >
               Group must contain at least one owner
             </div>
-            <div v-else class="my-6 text-center" data-test-id="no-owners-mgs">
```
dimak1: What's the reason for this removal?

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
```diff
@@ -53,13 +54,28 @@ export const useMhrInformation = () => {
...
+
+  const parseRemovedOwnerGroups = (): any => {
+    const ownerGroups = []
+
+    getMhrTransferHomeOwners.value.forEach(ownerGroup => {
+      const { groupId, ...owners } = ownerGroup
+      ownerGroups.push({
+        owners: [owners].filter(owner => owner.action === ActionTypes.REMOVED),
+        groupId: parseInt(groupId),
+        type: 'SO'
```
dimak1: We mark all deleted owners as Sole Owners?

dimak1 (commented):
Nice implementation. I have a few comments. Also, `Hide Deleted Owners` toggle button is not part of it?

cameron-eyds:
> Nice implementation. I have a few comments. Also, `Hide Deleted Owners` toggle button is not part of it?

I did not see that in the ticket, no.

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
cameron-eyds: We could but then have to override on the Badges and Actions explicitly as well. 
So might as well just apply the styling explicitly to the text.

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
cameron-eyds: There was a bug here. This cell would always render, so after removing owner it would default to 'no owners added yet' msg, in the event we were not dealing with groups. 
Down below there is messaging already for that no?

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
cameron-eyds: This was copy pasta originally, probably from our MVP approach as SO?
This should be dynamic yes, for both Added and Removed. Will update this.

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
dimak1: Those are different scenarios for Registrations (from what I remember). One would render initially when there is no data, and the other one would render after owners were removed.

doug-lovett (approved):
Very nice.

ChasBoucher (approved):
Looks Good.

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
cameron-eyds: After pulling in your commit, there is a bit more work to do here to figure out the type based on what is being submitted in the payload vs what is in Store.

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
cameron-eyds: I tested this removal in Registrations and i couldn't find a scenario where there was no msg displayed in the absence of an Owner.

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
cameron-eyds: Updated. Bit more work than anticipated.

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -152,7 +152,7 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
     }
   }
 
-  const getGroupForOwner = (ownerId: string, isTransfer = false): MhrRegistrationHomeOwnerGroupIF => {
```
cameron-eyds, line 155: Argument was not used

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
dimak1: If you add a solo owner and then remove it - this condition is triggered and a message shows up.

cameron-eyds: Ah fair enough, it's a bug with the data being cleared when the owner is removed. This additional conditional template was just masking it.

Screen Shot 2022-10-18 at 3 23 01 PM

cameron-eyds: It should resolve to no data but the groupId is persisting after clearing the new added owner in Registrations. 
Can fix.

dimak1: Yea, not ideal, but there was a reason for this. We need to preserve the Groups with no Owners, I think this was the fix for it.

cameron-eyds: Hmm ok, will take another look. 
This work around was causing issues with Transfers.

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
dimak1: There might be a few other places we could remove this argument, since we set it up at composable level.

cameron-eyds: You bet, will have another look. 
The refactor came in with the rebase!

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
cameron-eyds: Restored your original code to preserve the work around but also added a condition to disable for Transfer Flow.

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
cameron-eyds: Refactored to utilize the Composable argument and removed individual arguments.

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
cameron-eyds: Just a note, we may need to revisit this as we explore the removal of Groups in Transfer Flow. 
Coming soon *TM

dimak1 (approved):
Looks good!

PR #1001, opened 2022-10-20
13959 UI authorization component

owillborn:
*Issue #:* /bcgov/entity#13959
reviewConfirm

*Description of changes:*
- Added authorization component to transfer review section
- Made minor UI changes for what is visible when user is/isn't in review mode

### DIFF
--- a/ppr-ui/src/views/mhrInformation/MhrInformation.vue
+++ b/ppr-ui/src/views/mhrInformation/MhrInformation.vue
@@ -27,6 +27,8 @@
 
               <!-- MHR Information Review Section -->
               <template v-if="isReviewMode">
+                <!-- TODO: Add some form of transferDetails review, either review flag in existing component or
+                new component. To be added in ticket 13905 -->
                 <section>
                   <HomeOwnersTable
                     class="px-7"
@@ -36,10 +38,10 @@
                     :currentHomeOwners="getMhrTransferCurrentHomeOwners"
                   />
                 </section>
-                <section id="transfer-certify-section" class="mt-10 pt-4 pb-60px">
+                <section id="transfer-certify-section" class="mt-10 py-4">
                   <CertifyInformation
-                  :setShowErrors="validateAuthorizationError"
-                  @certifyValid="authorizationValid = $event"
+                    :setShowErrors="validateAuthorizationError"
+                    @certifyValid="authorizationValid = $event"
                   />
                 </section>
               </template>
@@ -280,7 +282,4 @@ export default defineComponent({
 
 <style lang="scss" scoped>
 @import '@/assets/styles/theme.scss';
-.pb-60px {
-  padding-bottom: 60px;
-}
 </style>
--- a/ppr-ui/src/views/mhrInformation/MhrInformation.vue
+++ b/ppr-ui/src/views/mhrInformation/MhrInformation.vue
@@ -11,7 +11,7 @@
           <v-col cols="9">
             <v-row no-gutters id="mhr-information-header" class="pt-3 pb-3 soft-corners-top">
               <v-col cols="auto">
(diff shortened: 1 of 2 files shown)


dimak1:
@owillborn please update the Description with a ticket/issue number. Also would be nice to see a screenshot of it, would be easier to review the code.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -27,22 +27,29 @@
...
+                    isMhrTransfer
+                    isReadonlyTable
+                    :homeOwners="reviewOwners"
+                    :currentHomeOwners="getMhrTransferCurrentHomeOwners"
+                  />
+                </section>
+                <section id="transfer-certify-section" class="mt-10 pt-4 pb-60px">
+                  <CertifyInformation
+                  :setShowErrors="validateAuthorizationError"
+                  @certifyValid="authorizationValid = $event"
```
cameron-eyds: missing an indent here on the props/events

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -263,4 +280,7 @@ export default defineComponent({
 
 <style lang="scss" scoped>
 @import '@/assets/styles/theme.scss';
+.pb-60px {
+  padding-bottom: 60px;
+}
```
cameron-eyds: Maybe a bit much, try and avoid excessive bottom padding if you can. 
Ideally lean on the padding top so when you come in to add new components your not battling bottom padding from previous components. 

Could we remove this and perhaps change the `pt-4` to `py-4` on the `section` tag where this class was used. 
It might seem narrow for now but all good, we have several more components to slot in here :)

cameron-eyds (approved):
Nice job! Just a few minor comments.

doug-lovett (approved):
LGTM

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -27,22 +27,29 @@
...
               </template>
 
               <!-- MHR Information Section -->
               <template v-else>
                 <HomeOwners isMhrTransfer class="mt-n2" />
+                <TransferDetails :validateTransferDetails="validateTransferDetails" />
               </template>
             </section>
-
-            <TransferDetails :validateTransferDetails="validateTransferDetails" />
```
dimak1, line 45: Besides the Review, TransferDetails should also be showing here, but be dynamic.

dimak1 (approved):

ppr-ui/src/views/mhrInformation/MhrInformation.vue
owillborn: Will do!

cameron-eyds:
@owillborn Looks great, thanks for the updates!

dimak1:
@owillborn just a minor UI update is needed. In the Authorization table, the user's name has too much spacing from the icon.

owillborn:
> @owillborn just a minor UI update is needed. In the Authorization table, the user's name has too much spacing from the icon.

@dimak1 Looks like the same spacing as in MHR registration review and confirm, should I make the change to the CertifyInformation component?

dimak1:
@owillborn `CertifyInformation` is used on other pages, so we don't want to break those by making changes to it. So the answer is no. The way to fix it is to use css on the Review page, and overwrite the styles for that specific class.

owillborn:
@dimak1 gotcha, thanks!

owillborn:
@dimak1 I'm going to leave the UI bug for now as I got stuck trying to figure it out. If its a quick fix for you I would be interested to see how its done :)

PR #1021, opened 2022-10-31
13746 add draft transfers to mhr registration table

ChasBoucher:
*Issue #:* /bcgov/entity#13746

*Description of changes:*

- Add Drafts to MHR Registrations Tab
- 
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the PPR license (Apache 2.0).

PR #1023, opened 2022-10-31
13746 Add Drats to MHR Dashboard

ChasBoucher:
*Issue #:* /bcgov/entity#13746

*Description of changes:*
[] Add drafts to to transfers in MHR Registration table.

### DIFF
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -66,12 +66,12 @@
         <label style="cursor: pointer;">
           <span v-if="!isExpanded">View </span>
           <span v-else>Hide </span>
-          Amendments
+          {{ isPpr ? 'Amendments' : 'History' }}
         </label>
       </v-btn>
     </td>
     <td v-if="inSelectedHeaders('createDateTime')" :class="isChild || item.expanded ? $style['border-left']: ''">
-      <span v-if="!isDraft(item)">
+      <span v-if="!isDraft(item) || !isPpr">
         {{ getFormattedDate(item.createDateTime) }}
       </span>
       <span v-else>
@@ -82,15 +82,15 @@
       v-if="inSelectedHeaders('statusType')"
       :class="isChild || item.expanded ? $style['border-left']: ''"
     >
-      <div v-if="!isChild || isDraft(item)">
+      <div v-if="!isChild || isDraft(item) || !isPpr">
         {{ getStatusDescription(item.statusType) }}
         <p v-if="!isChild && item.hasDraft" class="ma-0">
-          <i>* Draft Amendment</i>
+          <i>{{ isPpr ? '* Draft Amendment' : '* Draft Changes' }}</i>
         </p>
       </div>
     </td>
     <td
-      v-if="inSelectedHeaders('registeringName')"
+      v-if="inSelectedHeaders('registeringName') && isPpr"
       :class="isChild || item.expanded ? $style['border-left']: ''"
     >
       <span v-if="item.registeringName">{{ getRegisteringName(item.registeringName) }}</span>
@@ -103,10 +103,10 @@
       {{ item.registeringParty || item.submittingParty || '' }}
     </td>
     <td
-      v-if="inSelectedHeaders('ownerNames')"
+      v-if="inSelectedHeaders('ownerNames') && isPpr"
       :class="isChild || item.expanded ? $style['border-left']: ''"
     >
-      {{ item.ownerNames }}
+      {{ item.ownerNames && isPpr}}
     </td>
     <td
       v-if="inSelectedHeaders('securedParties') && isPpr"
@@ -115,7 +115,7 @@
       {{ item.securedParties || '' }}
     </td>
     <td
-      v-if="inSelectedHeaders('clientReferenceId')"
+      v-if="inSelectedHeaders('clientReferenceId') && isPpr"
       :class="isChild || item.expanded ? $style['border-left']: ''"
     >
       {{ item.clientReferenceId }}
@@ -597,7 +597,8 @@ export default defineComponent({
 
     const isDraft = (item: any): boolean => {
       // RegistrationSummaryIF | DraftResultIF | MhrDraftTransferApiIF
-      return props.isPpr ? item.type !== undefined : item.statusType === undefined
+      return props.isPpr
+        ? item.type !== undefined : (item.statusType === APIStatusTypes.DRAFT || item.statusType === undefined)
     }
 
     const isExpired = (item: RegistrationSummaryIF): boolean => {
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -106,7 +106,7 @@
       v-if="inSelectedHeaders('ownerNames') && isPpr"
       :class="isChild || item.expanded ? $style['border-left']: ''"
     >
-      {{ item.ownerNames && isPpr}}
+      {{ item.ownerNames}}
     </td>
     <td
       v-if="inSelectedHeaders('securedParties') && isPpr"
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -90,7 +90,7 @@
       </div>
     </td>
     <td
-      v-if="inSelectedHeaders('registeringName') && isPpr"
+      v-if="inSelectedHeaders('registeringName')"
       :class="isChild || item.expanded ? $style['border-left']: ''"
     >
       <span v-if="item.registeringName">{{ getRegisteringName(item.registeringName) }}</span>
@@ -103,7 +103,7 @@
       {{ item.registeringParty || item.submittingParty || '' }}
     </td>
     <td
-      v-if="inSelectedHeaders('ownerNames') && isPpr"
+      v-if="inSelectedHeaders('ownerNames')"
       :class="isChild || item.expanded ? $style['border-left']: ''"
     >
       {{ item.ownerNames}}
@@ -115,7 +115,7 @@
       {{ item.securedParties || '' }}
     </td>
     <td
-      v-if="inSelectedHeaders('clientReferenceId') && isPpr"
+      v-if="inSelectedHeaders('clientReferenceId')"
       :class="isChild || item.expanded ? $style['border-left']: ''"
     >
       {{ item.clientReferenceId }}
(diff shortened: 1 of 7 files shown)


ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -198,9 +200,44 @@ export const useNewMhrRegistration = () => {
...
     const draftFilings = await getMhrDrafts()
     const myMhrHistory = await mhrRegistrationHistory()
-    setMhrTableHistory([...draftFilings, ...myMhrHistory])
+    const filteredMhrHistory = addHistoryDraftsToMhr(myMhrHistory, draftFilings)
+    setMhrTableHistory([...filteredMhrHistory])
   }
 
+  function addHistoryDraftsToMhr (mhrHistory: MhRegistrationSummaryIF[], mhrDrafts: MhrDraftTransferApiIF[]):
+    MhRegistrationSummaryIF[] {
+    const sorter = require('lodash')
```
dimak1: We can use import lodash as usual, and even better if we just import what we need, instead of the whole library. 

`import { orderBy } from 'lodash'`

dimak1 (approved):
Approved with a comment. Also, could you please add a test?

dimak1:
@ChasBoucher you have another PR #1021 that does the same? Why two PRs? Can we close one of them?

dimak1 (changes requested):
This PR is similar/same as another PR #1021.

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
ChasBoucher: Changed.

ChasBoucher:
Closed 1021

ppr-ui/src/resources/tableHeaders.ts
```diff
@@ -467,34 +467,13 @@ export const mhRegistrationTableHeaders: Array<BaseHeaderIF> = [
...
-    value: 'ownerNames',
-    display: true
-  },
-  {
-    class: 'min-column-mds',
-    sortable: false,
-    text: 'Folio/Reference Number',
-    value: 'clientReferenceId',
-    display: true
-  },
```
cameron-eyds: Did Designs call for these columns to be removed?

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -198,9 +200,44 @@ export const useNewMhrRegistration = () => {
...
+            mhrNumber: transfer.mhrNumber,
+            baseRegistrationNumber: transfer.mhrNumber,
+            submittingParty: draft.submittingParty,
+            clientReferenceId: transfer.clientReferenceId,
+            createDateTime: draft.createDateTime,
+            error: draft.error,
+            registrationDescription: draft.registrationDescription,
+            hasDraft: true,
+            ownerNames: '',
+            path: draft.path,
```
cameron-eyds, line 233: Would `path` for the draft serve the same purpose as the `hasDraft` flag?
What does the api return in the event when there is no draft?

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -198,9 +200,44 @@ export const useNewMhrRegistration = () => {
...
+    // add drafts to Registrations..
+    const filteredMhrHistory = sorter
+      .orderBy(mhrHistory.filter(t => t.registrationDescription === 'REGISTER NEW UNIT'), ['createdDateTime'], ['desc'])
+    filteredMhrHistory.forEach(transfer => {
+      transfer.baseRegistrationNumber = transfer.mhrNumber
+      var mhrDrafts = sortedDraftFilings.filter(s => s.mhrNumber === transfer.mhrNumber)
+      if (mhrDrafts?.length > 0) {
+        transfer.hasDraft = true
+        transfer.changes = []
+        mhrDrafts.forEach(draft => {
```
cameron-eyds: Some spacing might help for readability. 

Don't use `var` please, `let` if `mhrDrafts` needs reassigning but `const` otherwise. 
So in this case, `const` 

please use more accurate names in your filter expressions, `t` and `s` is not very intuitive to any other developer who might work here. 

Also, please check if there is an ENUM value for `REGISTER NEW UNIT`.

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
cameron-eyds, line 233: Would it be easier here to spread the `draft` and include any additional information? Or is there specific information in the draft you don't want?
ie `...draft, newProperties: newValues` etc

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -198,9 +200,44 @@ export const useNewMhrRegistration = () => {
...
 
+  function addHistoryDraftsToMhr (mhrHistory: MhRegistrationSummaryIF[], mhrDrafts: MhrDraftTransferApiIF[]):
+    MhRegistrationSummaryIF[] {
+    const sorter = require('lodash')
+    const sortedDraftFilings = sorter.orderBy(mhrDrafts, ['createDateTime'], ['desc'])
+    // add drafts to Registrations..
+    const filteredMhrHistory = sorter
+      .orderBy(mhrHistory.filter(t => t.registrationDescription === 'REGISTER NEW UNIT'), ['createdDateTime'], ['desc'])
+    filteredMhrHistory.forEach(transfer => {
+      transfer.baseRegistrationNumber = transfer.mhrNumber
```
cameron-eyds, line 217: Why the alias?

ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
```diff
@@ -97,6 +97,9 @@ export interface MhRegistrationSummaryIF {
   statusType: string
   submittingParty: string
   username: string
+  baseRegistrationNumber?: string
```
cameron-eyds, line 100: How does this differ from `mhrNumber`?

ppr-ui/src/components/tables/common/TableRow.vue
```diff
@@ -103,10 +103,10 @@
       {{ item.registeringParty || item.submittingParty || '' }}
     </td>
     <td
-      v-if="inSelectedHeaders('ownerNames')"
+      v-if="inSelectedHeaders('ownerNames') && isPpr"
       :class="isChild || item.expanded ? $style['border-left']: ''"
     >
-      {{ item.ownerNames }}
+      {{ item.ownerNames && isPpr}}
```
cameron-eyds: I'm not sure i understand the `isPpr` in the template. We've got a computed value in `item.ownerNames` which i'm assuming is a string or something followed by a boolean in `isPpr`. 
Was the intention to short circuit `item.ownerNames`?
If so, there is no fallback value either

ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
ChasBoucher: It is required as a reference in the child.. I thought the same as you at first, but found it did not work, until I included the baseRegistrationNumber.... It is based on the RegistrationSummaryIF..

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
ChasBoucher: I have made a change and removed the alias... I refer to order by directly now..

ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
cameron-eyds: ah i see, the table itself requires this property name.

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
cameron-eyds: I guess its a healthy mix of both transfer/draft. 
No action required here

ChasBoucher: Ok..

ppr-ui/src/resources/tableHeaders.ts
ChasBoucher: I followed the design in the ticket.. See https://projects.invisionapp.com/share/QW130SEIE7JB#/screens/469824642

cameron-eyds: I believe the design gives the assumption that these columns are 'filtered', not removed entirely.
Please verify with Tracey.  @ChasBoucher

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
ChasBoucher: Good Catch ...t,s  I put those there when I first... forgot to change..
Made some changes to clarify..

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
ChasBoucher: Very Good Question... 

Q1: I am not sure to be honest... I am using hasDraft to ensure that I don't break any existing logic... 

Q2: The api has two options... If you include "collapse=true&pageNumber=1&fromUI=true"  in querystring than the "changes" element is included which contains history, but not drafts.  That is why we add drafts to the original query.. 

The Registrations table uses the changes element to determine if the item is collapsible.

ppr-ui/src/components/tables/common/TableRow.vue
ChasBoucher: Good catch... Not necessary,

ChasBoucher:
Adding tests.

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
cameron-eyds: Ok sounds fine, we can give it a try 👍

ppr-ui/src/resources/tableHeaders.ts
ChasBoucher: Confirmed with Tracey.  Fields were not put in design,  I have made the change... need to test when BCRGEG.. is working...

cameron-eyds: @ChasBoucher Confirmed they are to remain and are filtered in the design? Or confirmed they are to be removed?

cameron-eyds (approved):
Thank you for the updates 👍

ChasBoucher:
> This PR is similar/same as another PR #1021.

I closed the request...

dimak1:
@ChasBoucher thanks for resolving all the comments. Approved.

dimak1 (approved):

ChasBoucher:
Approved for merge, but we are waiting for Registration history to be available to run one more test.

PR #1025, opened 2022-11-03
Validation for Transfer Details & Confirm Completion components

dimak1:
*Issue #:* /bcgov/entity#13878
*Issue #:* /bcgov/entity#13960

*Description of changes:*
- Validation for Transfer Details and Confirm Completion components
- Unit tests

### Screenshots:

Screen Shot 2022-11-04 at 2 48 17 PM

### DIFF
--- a/ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
+++ b/ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
@@ -24,7 +24,7 @@
     <!-- Authorization -->
     <section id="mh-certify-section" class="mt-10 pt-4">
       <CertifyInformation
-        isMhr
+        :sectionNumber=1
         :setShowErrors="validateAuthorization"
         @certifyValid="authorizationValid = $event"
       />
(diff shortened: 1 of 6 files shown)


ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
```diff
@@ -98,7 +98,6 @@ export default defineComponent({
     } = useMhrValidations(toRefs(getMhrRegistrationValidationModel.value))
 
     const localState = reactive({
-      isMhr: true,
```
cameron-eyds, line 101: 😆 Where did this property come from. 
Glad you removed!

cameron-eyds (approved):
Looks good!
Future state we might want to refactor the validations as the amount of Components that need validating grows. 

We could also modify our more global scroll to work everywhere. I do like the idea of Scrolling to the error border. Does it ever scroll past the section labels / headers etc?

ChasBoucher (approved):
Looks Good.

dimak1:
> Looks good!
> 
> Future state we might want to refactor the validations as the amount of Components that need validating grows. 
> 
> 
> 
> We could also modify our more global scroll to work everywhere. I do like the idea of Scrolling to the error border. Does it ever scroll past the section labels / headers etc?

@cameron-eyds thanks, yea we might want to refactor and create global validations/scrolls/utils. I agree, we have so many implementations that we need to consolidate.

Scroll to error scrolls to first class, and stops at the top, it doesn't scrolls past it.

PR #1026, opened 2022-11-03
UI cleanup and homeOwners validation edge case

owillborn:
*Issue #:* /bcgov/entity#13727

*Description of changes:*
-UI fixes (see original ticket for specifics)
-homeOwners section now validates when review and confirm screen is reached. Will display error bar on table if no owners added and review screen was reached.

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-card flat class="py-6 px-8 mb-20 rounded">
+  <v-card flat class="py-6 px-8 mb-5 rounded">
     <v-row id="mhr-home-add-person">
       <v-col cols="3">
         <label class="generic-label"> {{ getSidebarTitle }} </label>
@@ -500,8 +500,5 @@ export default defineComponent({
     color: $gray7;
     line-height: 24px;
   }
-}
-.mb-20{
-  margin-bottom: 20px !important;
 }
 </style>
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-card flat class="py-6 px-8 rounded">
+  <v-card flat class="py-6 px-8 mb-20 rounded">
     <v-row id="mhr-home-add-person">
       <v-col cols="3">
         <label class="generic-label"> {{ getSidebarTitle }} </label>
@@ -500,5 +500,8 @@ export default defineComponent({
     color: $gray7;
     line-height: 24px;
   }
+}
+.mb-20{
+  margin-bottom: 20px !important;
 }
 </style>
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-card flat rounded :class="{ 'border-error-left': showTableError }">
+  <v-card flat rounded :class="{ 'border-error-left': showTableError || reviewedNoOwners}">
     <v-data-table
       id="mh-home-owners-table"
       class="home-owners-table"
@@ -219,7 +219,7 @@ import { AddEditHomeOwner } from '@/components/mhrRegistration/HomeOwners'
 import TableGroupHeader from '@/components/mhrRegistration/HomeOwners/TableGroupHeader.vue'
 /* eslint-disable no-unused-vars */
 import { MhrRegistrationHomeOwnerIF } from '@/interfaces'
-import { ActionTypes } from '@/enums'
+import { ActionTypes, RouteNames } from '@/enums'
 /* eslint-enable no-unused-vars */
 import { useActions } from 'vuex-composition-helpers'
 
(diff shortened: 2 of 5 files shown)


ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-card flat class="py-6 px-8 rounded">
+  <v-card flat class="py-6 px-8 mb-20 rounded">
```
cameron-eyds: `mb-20` is out of range of the helper classes. I think the max helper class is 10.

cameron-eyds: To add, each increment here is worth about 4px by Vuetify definition. 
ie `mb-1` === 4px margin bottom and so on

owillborn: Thanks for letting me know, was unaware!

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -343,6 +345,10 @@ export default defineComponent({
       }
     )
 
+    watch(() => context.root.$route.name, (route: string) => {
+      if (route === RouteNames.MHR_REVIEW_CONFIRM) localState.reviewed = true
+    })
+
```
cameron-eyds: I believe we already have a validation flag that is set when a user navigates to MHR_REVIEW_CONFIRM.

This line in the Review Confirm View, 
`setValidation(MhrSectVal.REVIEW_CONFIRM_VALID, MhrCompVal.VALIDATE_STEPS, true)` is already setting this global flag, so i don't believe you need localState `reviewed` here.

ppr-ui/src/components/mhrRegistration/ReviewConfirm/HomeOwnersReview.vue
```diff
@@ -6,22 +6,25 @@
...
-      <section class="px-6 my-2" v-if="hasHomeOwners || (hasGroups && showGroups)">
+      <section v-show="!getStepValidation(MhrSectVal.HOME_OWNERS_VALID)"
+      :class="hasHomeOwners ? 'pt-30px px-6' : 'px-6 py-8'">
+        <span>
+          <v-icon color="error">mdi-information-outline</v-icon>
+          <span class="error-text mx-1">This step is unfinished.</span>
+          <router-link :to="{ path: `/${RouteNames.MHR_REGISTRATION}/${RouteNames.HOME_OWNERS}` }"
+          >Return to this step to complete it.</router-link>
+        </span>
+      </section>
```
cameron-eyds: Could you explain the change to `section` tag vs div? Is it based on validation scroll to effect?

Also refer to this for helper classes: https://vuetifyjs.com/en/styles/spacing/#how-it-works
Is `pt-30px` from vuetify or custom? 

Also, need an indent on the class binding `:class`

ppr-ui/src/components/mhrRegistration/ReviewConfirm/HomeOwnersReview.vue
```diff
@@ -6,22 +6,25 @@
...
+
+      <section class="px-6 my-2" v-if="hasHomeOwners">
```
cameron-eyds, line 19: What displays here if this is `false` 
I guess previously there was fallback values below for missing information.

owillborn: If this is false only the error message will display, as per requirements in the ticket.
image

cameron-eyds: ok fair enough, requirements maybe updated?

That being said, are the fallback values still utilized/required below?

ppr-ui/src/components/mhrRegistration/ReviewConfirm/HomeOwnersReview.vue
owillborn: Switched to section simply for consistency with the other section tags within the div. Didn't see any visual differences, is one preferred over the other in this case?

Thanks for the helper class link, pt-30px is a custom class in base.scss

Good catch

ppr-ui/src/components/mhrRegistration/ReviewConfirm/HomeOwnersReview.vue
owillborn: Not sure on what you mean by fallback values...

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
owillborn: You're right, I looked for a flag for a while but couldn't find it, guess its this one haha!

ppr-ui/src/components/mhrRegistration/ReviewConfirm/HomeOwnersReview.vue
cameron-eyds: Basically a fallback value is an `||` scenario. Ie render `someProperty` || 'N/A'

owillborn: If you mean the template within this v-if, then yes it is used to display the table on R/C when owners are present

owillborn: ohhhh gotcha, then I guess there's no case in which tenancy type can be N/A but still have an owner present. I'll remove the fallbacks

dimak1 (approved):
Other than Cameron's comments, looks good to me.

cameron-eyds (approved):
Thanks for the updates!

PR #1028, opened 2022-11-07
Update Submission Payload

dimak1:
*Description of changes:*
- Update payload data for submission
- Confirm if Submitting Party is a Business Account info (as in UI/Requirements) or a Current User's info

#### Submitting Party on the Review Screen:

Screen Shot 2022-11-07 at 1 44 21 PM

### DIFF
--- a/ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
+++ b/ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
@@ -1,7 +1,7 @@
-import { MhrHomeOwnerGroupIF, MhrTransferApiIF, MhrTransferIF } from '@/interfaces'
+import { MhrTransferApiIF, MhrTransferIF, SubmittingPartyIF } from '@/interfaces'
 import { useGetters } from 'vuex-composition-helpers'
 import { readonly, ref } from '@vue/composition-api'
-import { ActionTypes, ApiHomeTenancyTypes, HomeTenancyTypes } from '@/enums'
+import { ActionTypes, ApiHomeTenancyTypes, HomeTenancyTypes, UIRegistrationTypes } from '@/enums'
 
 // Validation flag for Transfer Details
 const transferDetailsValid = ref(false)
@@ -18,6 +18,7 @@ export const useMhrInformation = () => {
     getMhrTransferConsideration,
     getMhrTransferDate,
     getMhrTransferOwnLand,
+    getMhrTransferSubmittingParty,
     getMhrTransferAttentionReference,
     getMhrTransferHomeOwnerGroups
   } = useGetters<any>([
@@ -29,6 +30,7 @@ export const useMhrInformation = () => {
     'getMhrTransferConsideration',
     'getMhrTransferDate',
     'getMhrTransferOwnLand',
+    'getMhrTransferSubmittingParty',
     'getMhrTransferAttentionReference',
     'getMhrTransferHomeOwnerGroups'
   ])
@@ -45,7 +47,10 @@ export const useMhrInformation = () => {
     return {
       mhrNumber: '',
       ownerGroups: [],
-      submittingParty: {},
+      submittingParty: {
+        emailAddress: '',
+        phoneNumber: ''
+      },
       declaredValue: null,
       consideration: '',
       transferDate: '',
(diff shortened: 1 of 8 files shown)


ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
```diff
@@ -85,22 +91,23 @@ export const useMhrInformation = () => {
       transferDate: getMhrTransferDate.value,
       ownLand: getMhrTransferOwnLand.value,
       attentionReference: getMhrTransferAttentionReference.value,
-      documentDescription: 'SALE OR GIFT',
+      documentDescription: UIRegistrationTypes.TRANSFER_OF_SALE,
       submittingParty: {
+        // Need to confirm if this needs to be a Business Name instead
+        // because in UI, Submitting Party is a business not a person
+        // but the API spec requires firstName, lastName
```
cameron-eyds, line 98: Submitting party can be a person OR a business afaik

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
```diff
@@ -85,22 +91,23 @@ export const useMhrInformation = () => {
...
-        emailAddress: getCurrentUser.value.contacts[0].email,
-        phoneNumber: getCurrentUser.value.contacts[0].phone.replace(/[^0-9.]+/g, '') // Remove special chars
-      },
+        address: getMhrTransferSubmittingParty.value.address,
+        emailAddress: getMhrTransferSubmittingParty.value.emailAddress,
+        phoneNumber: getMhrTransferSubmittingParty.value.phoneNumber,
+        phoneExtension: getMhrTransferSubmittingParty.value.phoneExtension
+        // Cleanup below after confirming that Submitting Party info is Account's Admin info
+        // emailAddress: getCurrentUser.value.contacts[0].email,
+        // phoneNumber: getCurrentUser.value.contacts[0].phone.replace(/[^0-9.]+/g, '') // Remove special chars
```
cameron-eyds, line 109: I think you can confirm with Tracey and remove these comments 👍 
That being said, there is no submitting party information completed in a Transfer filing so how are the root properties of `getMhrTransferSubmittingParty` populated? 
Is it current account information?

cameron-eyds: Ah i looked below and see. 
So isn't saving the results of `getAccountInfoFromAuth` ie currentAccount to both current account info and submitting party info a bit of a decoy? ie storing the same information twice under two different alias.

*Sorry another edit, not saying anything is wrong here, just that we have a lot of these decoys/patterns to be aware of :(

cameron-eyds: Or is that different from current account?

dimak1:
@cameron-eyds updated the Description with a screenshot from UI requirements. We do display the Account name. In the database, the name is just one field. So currently (in our submit call) we pass the Current User's first name and last name, which I think should be an Account name. I also believe the current API spec is outdated, and Thor will send me the latest specs.

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
dimak1: Current account is the logged-in user (e.g BCREG2 Della FORTYTHREE). The Account Info contains the account name (e.g. PPR 2). So they are not the same. We are making a separate call to `/orgs` to get name and address info. We are doing some parsing to align the prop names of the interfaces. We could probably optimize it still.

cameron-eyds: For sure, the `/orgs` request is part of `getAccountInfoFromAuth` right?
That is called in `MhrInformation` via `getSubmittingPartyInformation`, saved to localState and subsequently submitting party info. So what do we designate that? 
It's misleading as the CONST used in that request is `CurrentAccount`

cameron-eyds: So to be plain, Submitting Party is or is not the Current Account?

dimak1: Just to reply to your comment. Based on **currentAccount** id we make a calls to `/orgs` to get Account Info submitting party.

dimak1:
This is an old PR and has many conflicts, and rebasing would take a long time. Opened a new PR #1131 instead. Will close this soon.

PR #1029, opened 2022-11-07
13746-Add missing transfers to child, add documentid

ChasBoucher:
*Issue #:* /bcgov/entity#13746

*Description of changes:*

- Add missing Transfers 
- Add "Completed" ux status for complted transfers
- Add documentid to provide uniqueness for table - follows existing ppr logic.
- Fix deletion/editing of draft

### DIFF
--- a/ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
+++ b/ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
@@ -9,7 +9,7 @@ import {
   MhrDraftTransferApiIF
 } from '@/interfaces'
 import { StaffPaymentIF } from '@bcrs-shared-components/interfaces'
-import { APIMhrDescriptionTypes, APIStatusTypes, HomeTenancyTypes } from '@/enums'
+import { APIStatusTypes, HomeTenancyTypes } from '@/enums'
 import { getMhrDrafts, mhrRegistrationHistory } from '@/utils'
 import { orderBy } from 'lodash'
 export const useNewMhrRegistration = () => {
@@ -199,7 +199,7 @@ export const useNewMhrRegistration = () => {
 
   const fetchMhRegistrations = async (): Promise<void> => {
     const draftFilings = await getMhrDrafts()
-    const myMhrHistory = await mhrRegistrationHistory()
+    const myMhrHistory = await mhrRegistrationHistory(true)
     const filteredMhrHistory = addHistoryDraftsToMhr(myMhrHistory, draftFilings)
     setMhrTableHistory([...filteredMhrHistory])
   }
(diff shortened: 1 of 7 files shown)


cameron-eyds (approved):
Looks fine to me!
Have you tested against existing Registrations Table (PPR)?
Also, what is the documentId used for specifically?

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -208,21 +208,34 @@ export const useNewMhrRegistration = () => {
     MhRegistrationSummaryIF[] {
     const sortedDraftFilings = orderBy(mhrDrafts, ['createDateTime'], ['desc'])
 
-    const registerNewUnit = mhrHistory.filter(registrations =>
```
dimak1, line 211: Considering this composable is used in a few places, is it safe to update the logic here? Won't it break anything pre-existing?

dimak1 (commented):
@ChasBoucher so far looks good, just have one comment.

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
ChasBoucher: Glad you made me check. We're good. This is the only place that MhrHistory is being used at present.  I checked the api call and although I added the optional parameter, it is only being used here as well.

ChasBoucher:
DocumentId is the key used in the table.  It needs to be unique or a warning is thrown. 
Ppr Registrations are ok, but I just notice a few issues with the Staff role that I want to fix...

cameron-eyds:
> DocumentId is the key used in the table. It needs to be unique or a warning is thrown. Ppr Registrations are ok, but I just notice a few issues with the Staff role that I want to fix...

Sounds good thanks, so `documentId` is generated and returned by the backend i'm assuming. 
There is probably a business rule around that! Helpful in providing a unique ID for us then :) 
Thanks for the looking into it.

cameron-eyds:
@ChasBoucher Sorry Charles, 1 more item noticed after the DRAFT refactors. 
When you resume a DRAFT using the EDIT button, i no longer see it pulling on the DRAFT itself but the original Registration again. 
Does this PR also resolve that?

cameron-eyds:
@ChasBoucher Sorry Charles, 1 more item noticed after the DRAFT refactors. 
When you resume a DRAFT using the EDIT button, i no longer see it pulling on the DRAFT itself but the original Registration again. 
Does this PR also resolve that?

ChasBoucher:
> @ChasBoucher Sorry Charles, 1 more item noticed after the DRAFT refactors. When you resume a DRAFT using the EDIT button, i no longer see it pulling on the DRAFT itself but the original Registration again. Does this PR also resolve that?

Just tested and this pr does address the issue with editing of drafts as well.

cameron-eyds:
> Collaborator

Excellent, thank you 👍

cameron-eyds (approved):

PR #1033, opened 2022-11-08
Fix Attention Ref Num - QA feedback

dimak1:
*Issue #:* /bcgov/entity#13878

*Description of changes:*

- Fix Attention Ref Num from showing up for different Transfers
- Update error message on Review screen

### DIFF
--- a/ppr-ui/src/views/mhrInformation/MhrInformation.vue
+++ b/ppr-ui/src/views/mhrInformation/MhrInformation.vue
@@ -277,16 +277,19 @@ export default defineComponent({
         return !isGlobalEditingMode.value && isTransferDetailsValid.value && true
       }),
       isValidTransferReview: computed((): boolean => { // is valid on review step
-        return localState.isReviewMode &&
-        isRefNumValid.value &&
-        localState.isCompletionConfirmed &&
-        !localState.validateAuthorizationError
+        return (
+          localState.isReviewMode &&
+          isRefNumValid.value &&
+          localState.isCompletionConfirmed &&
+          !localState.validateAuthorizationError
+        )
       }),
       transferErrorMsg: computed((): string => {
         const isValidReview = localState.isReviewMode ? !localState.isValidTransferReview : !localState.isValidTransfer
-        return localState.validate && isValidReview && localState.isReviewMode
-          ? '< Please make any required changes'
-          : ''
+        const errorMsg = localState.isReviewMode
+          ? '< Please complete required information'
+          : '< Please make any required changes'
+        return localState.validate && isValidReview ? errorMsg : ''
       }),
       reviewConfirmText: computed((): string => {
         return localState.isReviewMode ? 'Register Changes and Pay' : 'Review and Confirm'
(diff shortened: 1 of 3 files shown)


owillborn:
Too many changes... Looks good to me

owillborn (approved):

dimak1:
Converting to Draft as I need to make another update...

ChasBoucher (approved):
Looks Good

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -279,24 +277,27 @@ export default defineComponent({
...
+        return localState.validate && isValidReview ? errorMsg : ''
       }),
       reviewConfirmText: computed((): string => {
         return localState.isReviewMode ? 'Register Changes and Pay' : 'Review and Confirm'
       }),
       reviewOwners: computed(() => {
         return getMhrTransferHomeOwners.value.filter(owner => owner.action !== ActionTypes.REMOVED)
       }),
-      attentionReference: getMhrTransferAttentionReference.value,
+      attentionReference: '',
```
cameron-eyds, line 300: I guessed right.
Is store also preserving data on a cancellation of a Transfer?

cameron-eyds: Using the getter here is not a bad way to do it, the root of the bug is probably that store isn't clear on cancel (Or re initialized during a new tranfers, either or)

dimak1: The store is cleared on transfer initiation, but the value in the field (on the page) was still displaying the old value.(**Edit**: that was the bug when the value was still in the field somehow). I tested this fix and the data is preserved in the store when navigating Back and Next, and also it is properly resetting on the new Transfer.

cameron-eyds: Oh! How was it preserving the store data if it was cleared (properly) on Transfer initiation?

dimak1: It is resetting to empty as expected on Transfer initiation. Also, it is preserved in store (even with the getter removed).

cameron-eyds: I understand that is what it's doing now after de-coupling the local property from the getter.
I'm referring to the original bug. 

Correct me if i'm wrong, wasn't the number preserved in store which was why the component was rendering with a previous entered Number on a new transfer?

dimak1: That's the thing - the store was empty but the field in UI was not.

Screen Shot 2022-11-08 at 12 11 36 PM

cameron-eyds: Very interesting indeed! 
Ok, will resolve this if you are confident that the bug its not a failure to clear the store data when initializing a new Transfer. 
Thanks for the discussion!

cameron-eyds (approved):

ppr-ui/src/views/mhrInformation/MhrInformation.vue
dimak1: During Transfer init, we do reset the data to be empty (all good there). But it is not reflected on the UI. 
Thanks!

PR #1034, opened 2022-11-09
Duplicate secured party fixes

owillborn:

*Description of changes:*

- A few UI fixes in the autocomplete list when searching to add existing secured parties:
           - Added parties cannot be interacted with in the list
           - Parties that were already added now have different coloring and don't change when clicked
           - Changed mouse rollover actions on added parties
           
- Updated contact info phone number and text, used by error dialogs

- When a user manually enters a secured party (individual or business), that party is compared with the current secured parties to ensure no exact matches of first, middle and last names, as well as address.
- If such a match is found, an error dialogue will be displayed to the user and the party will not be added to the table.
image

- SecuredPartyDialog component with isDuplicate prop will be used to display undetected Duplicate Party error in review and confirm once API check is implemented

*** #14177 Similar secured party check for edit: Made a small change to complete this ticket. Similar parties are now detected when a Secured Party is edited.

### DIFF
--- a/ppr-ui/package.json
+++ b/ppr-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ppr-ui",
-  "version": "0.3.41",
+  "version": "0.3.42",
   "private": true,
   "appName": "Assets UI",
   "sbcName": "SBC Common Components",
--- a/ppr-ui/package.json
+++ b/ppr-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ppr-ui",
-  "version": "0.3.42",
+  "version": "0.3.44",
   "private": true,
   "appName": "Assets UI",
   "sbcName": "SBC Common Components",
--- a/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
+++ b/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
@@ -6,7 +6,7 @@
     attach="#app"
     content-class="secured-party-dialog"
   >
-    <v-card id="secured-party-dialog" class="pl-4 pr-1 pt-7 mt-7">
+    <v-card id="secured-party-dialog" :class="!duplicate && !review ? 'pl-4 pr-1 pt-7 mt-7' : 'pl-1 pr-1 pt-7 mt-7'">
       <v-row no-gutters>
         <v-col cols="11">
           <v-row no-gutters>
@@ -40,6 +40,10 @@
           existing {{ partyWord }} Party listed below or use your information to create
           a new {{ partyWord }} Party?
         </p>
+        <p v-else-if="!review" class="text-md-center px-6 pt-3" :class="$style['intro']">
+          Registrations cannot list a Secured Party with the same name and address more than once. This registration
+          already contains this Secured Party:
+        </p>
         <p v-else class="text-md-center px-6 pt-3" :class="$style['intro']">
           Duplicate Secured Parties have been detected in this registration.<br>
           Registrations cannot list a Secured Party with the same name and address more than once.<br>
@@ -52,7 +56,7 @@
         </div>
 
         <v-container class="currentParty">
-          <v-row :class="[$style['companyRow'], { 'primaryRow': showSelected }]">
+          <v-row :class="[$style[duplicate ? 'companyRowDuplicate' : 'companyRow'], { 'primaryRow': showSelected }]">
             <v-col cols="auto" :class="$style['iconColumn']">
               <v-icon :class="$style['companyIcon']">
                 {{party.businessName ? 'mdi-domain' : 'mdi-account'}}
@@ -127,7 +131,7 @@
             >
           </v-row>
         </v-container>
-        <div v-else >
+        <div v-else-if="review" >
           <p class="text-md-center px-6 pt-3" :class="$style['intro']">
             Cancelling and re-starting you registration may resolve this issue.<br>
             If you do not wish to proceed, contact BC registries staff:
@@ -136,13 +140,20 @@
         </div>
       </div>
       <v-card-actions class="pt-6 pb-8">
-        <v-btn
+        <v-btn v-if="!duplicate && !review"
           :class="$style['dialogButton']"
           id="dialog-cancel-button"
           color="primary"
           outlined
           @click="exit()"
           >Exit</v-btn
+        >
+        <v-btn v-else
+          :class="$style['dialogButton']"
+          class="primary dialog-btn"
+          color="primary"
+          @click="exit()"
+          >OK</v-btn
         >
       </v-card-actions>
     </v-card>
@@ -180,6 +191,10 @@ export default defineComponent({
       type: Boolean,
       default: false
     },
+    isReview: {
+      type: Boolean,
+      default: false
+    },
     defaultIsRegisteringParty: Boolean
   },
   setup (props, context) {
@@ -187,6 +202,9 @@ export default defineComponent({
       showSelected: true,
       duplicate: computed((): boolean => {
         return props.isDuplicate
+      }),
+      review: computed((): boolean => {
+        return props.isReview
       }),
       party: computed((): PartyIF => {
         return props.defaultParty
@@ -270,6 +288,16 @@ export default defineComponent({
 }
 .companyText {
   font-weight: 700;
+}
+.companyRowDuplicate {
+  background-color: #f1f1f1;
+  border-radius: 4px 4px 4px 4px;
+  margin-bottom: 10px;
+  padding-top: 10px;
+  padding-bottom: 10px;
+  margin-right: 10px;
+  margin-left: 10px;
+  border: 1px solid white;
 }
 
 .companyRow {
--- a/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
+++ b/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
@@ -56,7 +56,8 @@
         </div>
 
         <v-container class="currentParty">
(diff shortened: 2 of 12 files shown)


owillborn:
Converting to draft found some visual issues

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -308,4 +340,8 @@ export default defineComponent({
   margin-left: 330px;
   width: 75px;
 }
+.padLeft {
```
dimak1: Should be kebab case: `pad-left`

ppr-ui/src/components/parties/composables/useSecuredParty.ts
```diff
@@ -95,6 +95,22 @@ export const useSecuredParty = (props, context) => {
     return idx !== -1
   }
 
+  const hasMatchingSecuredParty = (addedParty: PartyIF): boolean => {
+    // store state without newly added party
+    const parties = cloneDeep(getAddSecuredPartiesAndDebtors.value.securedParties)
+    if (localState.partyBusiness === 'I') {
+      return parties.some(party =>
+        JSON.stringify(party.personName) === JSON.stringify(addedParty.personName) &&
```
dimak1: Why use JSON.stringify here?

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
owillborn: good catch, fixed

ppr-ui/src/components/parties/composables/useSecuredParty.ts
owillborn: This is the way I know to compare entire objects for equivalence, is there a better way to do so? I thought this also looked cleaner than comparing each object attribute

owillborn: party.personName and party.address are both objects

owillborn: @dimak1 Wrote it a different way let me know what you think

dimak1 (approved):
Looks good to me.

ppr-ui/src/components/parties/composables/useSecuredParty.ts
dimak1: Looks good, thanks.

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -151,11 +188,25 @@ export default defineComponent({
...
   },
   setup (props, context) {
     const localState = reactive({
       showSelected: true,
+      duplicate: computed((): boolean => {
+        return props.isDuplicate
+      }),
+      review: computed((): boolean => {
+        return props.isReview
+      }),
```
cameron-eyds: Not really necessary. I would remove the local computed values here and use the props. 
It also helps improve the naming convention, `isDuplicate` and `isReview` is a little more intuitive/representative of a boolean. 

I've seen this type of implementation in a few places where the props are essentially aliased through a computed property and it's just not required. The prop can be used directly in the template and will be reactive. In some instances i've even seen the alias'ing the cause of some properties becoming Non Reactive.

ppr-ui/src/components/parties/composables/useSecuredParty.ts
```diff
@@ -95,6 +95,22 @@ export const useSecuredParty = (props, context) => {
     return idx !== -1
   }
 
+  const hasMatchingSecuredParty = (addedParty: PartyIF): boolean => {
+    // store state without newly added party
+    const parties = cloneDeep(getAddSecuredPartiesAndDebtors.value.securedParties)
+    if (localState.partyBusiness === 'I') {
```
cameron-eyds: This doesn't make a lot of sense to me, from an outside perspective. What does `I` mean? 

Looking in the code i see `partyBusiness.value ===  I` or `partyBusiness.value === B` but what do these values represent? 

If you know, it might be a good idea to create an enum representing what these values truly mean so it's more readable & understandable in the code.

ppr-ui/src/components/parties/party/EditParty.vue
```diff
@@ -363,21 +373,20 @@ export default defineComponent({
...
-              currentSecuredParty.value.businessName, false
-            )
-            // check if any results
-            if (response?.length > 0) {
-              // show secured party selection popup
-              showDialog()
-              localState.dialogResults = response?.slice(0, 50)
-              return
-            }
+        if ((currentSecuredParty.value.businessName) && (partyBusiness.value === 'B')) {
```
cameron-eyds: Only 1 set of parenthesis required here. 
Also, perhaps we could update our `B` value into an enum with is more representative of what the value means.

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -30,39 +35,53 @@
...
               <div :class="$style['addressText']">
                 {{ party.address.street }}, {{ party.address.city }}
                 {{ party.address.region }} , {{ party.address.postalCode }},
                 {{ getCountryName(party.address.country) }}
               </div>
               <div>
-                <v-chip x-small label color="primary" text-color="white"
+                <v-chip  x-small v-if="!duplicate" label color="primary" text-color="white"
                   >NEW</v-chip
                 >
```
cameron-eyds: I know not your code, but perhaps keep the tags on the same line. 
The exception is typically a self closing component. ie `<SomeComponent  />` 
But this is a closing tag in itself, so best together. There is a few instances of this as well, try to catch them 👍

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -113,16 +132,30 @@
...
       <v-card-actions class="pt-6 pb-8">
-        <v-btn
+        <v-btn v-if="!duplicate && !review"
           :class="$style['dialogButton']"
           id="dialog-cancel-button"
           color="primary"
           outlined
           @click="exit()"
           >Exit</v-btn
         >
```
cameron-eyds: As above

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -113,16 +132,30 @@
...
+        <v-btn v-else
+          :class="$style['dialogButton']"
+          class="primary dialog-btn"
+          color="primary"
+          @click="exit()"
+          >OK</v-btn
+        >
```
cameron-eyds: As above

cameron-eyds (commented):
Great job! Just a few comments, nice work on this 👍

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
owillborn: Done, good catch. Getting more familiar with how/when to use props vs localState in the template

ppr-ui/src/components/parties/party/EditParty.vue
owillborn: Thanks, changed to enum values

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
owillborn: Quick fixes

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
cameron-eyds: You are not alone! 
The script syntax changed a between Vue versions and this little bit was just part of the learning curve from that version jump. 
Thank you for the clean up 👍

ppr-ui/src/components/parties/composables/useSecuredParty.ts
owillborn: Changed to enums, you're right about it not making much sense. New implementation is much easier to read/follow

ppr-ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ppr-ui",
-  "version": "0.3.41",
+  "version": "0.3.42",
```
cameron-eyds: You might need to bump version to `44`, just a little behind!

ppr-ui/src/components/parties/composables/useSecuredParty.ts
cameron-eyds: So much better, thank you 👍

cameron-eyds (approved):
Looks great, nice clean up and less code 👍

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -6,19 +6,24 @@
     attach="#app"
     content-class="secured-party-dialog"
   >
-    <v-card id="secured-party-dialog" class="pl-4 pr-1 pt-7 mt-7">
+    <v-card id="secured-party-dialog" :class="!isDuplicate && !isReview ? 'pl-4 pr-1 pt-7 mt-7' : 'pl-1 pr-1 pt-7 mt-7'">
```
dimak1: You can also combine static and dynamic class attributes since only one css class is dynamic. eg:
`class="pr-1 pt-7 mt-7" :class="!isDuplicate && !isReview ? 'pl-4' : 'pl-1'"`

owillborn: will do, I like that approach

PR #1036, opened 2022-11-12
Transfer Details validations

dimak1:
*Issue #:* /bcgov/entity#14091

*Description of changes:*

- Refactor Transfer Details validation to meet UX requirements
- Update unit tests

### Transfer Details validations

- Different states of the validations and label highlight 

Screen Shot 2022-11-11 at 10 42 55 PM

Screen Shot 2022-11-11 at 10 43 01 PM

Screen Shot 2022-11-11 at 10 43 10 PM

Screen Shot 2022-11-11 at 10 43 32 PM

### DIFF
--- a/ppr-ui/src/components/mhrTransfers/TransferDetails.vue
+++ b/ppr-ui/src/components/mhrTransfers/TransferDetails.vue
@@ -8,10 +8,14 @@
     </p>
 
     <v-card flat class="py-6 px-8 rounded" :class="{ 'border-error-left': showFormError }">
-      <v-form ref="transferDetailsForm" v-model="transferDetailsValid">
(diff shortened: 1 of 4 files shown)


cameron-eyds (approved):
Looks fine to me! 
Could probably have done the same handling from the parent (MhrInformation in this case) with a prop and event vs accessing the child methods through a ref.
This is just a different way of different things.

dimak1:
@cameron-eyds yea I was doing that initially, using a boolean trigger in parent and watcher in child components. But I had to send validation state back to parent, so went ahead with this approach using a ref.

cameron-eyds:
> 

@dimak1 Sounds good, thanks for the response! Sounds like you thought through the process and went with what works best for you, so well done!

PR #1039, opened 2022-11-14
Clear Transfer Details data on Undo

dimak1:
*Description of changes:*

- The requirement is to erase Transfer Details data/fields when the component gets hidden (for instance via Undo button click).

### DIFF
--- a/ppr-ui/src/views/mhrInformation/MhrInformation.vue
+++ b/ppr-ui/src/views/mhrInformation/MhrInformation.vue
@@ -492,7 +492,7 @@ export default defineComponent({
     watch(
       () => hasUnsavedChanges.value,
       (val: boolean) => {
-        if (val === false && context.refs.transferDetailsComponent) {
+        if (!val && context.refs.transferDetailsComponent) {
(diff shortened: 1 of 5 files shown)


ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -488,6 +489,16 @@ export default defineComponent({
       }
     )
 
+    watch(
+      () => hasUnsavedChanges.value,
+      (val: boolean) => {
+        if (val === false && context.refs.transferDetailsComponent) {
```
cameron-eyds: Could probably use `if (!val...)` here to simplify.

cameron-eyds (approved):
LGTM, just a minor comment.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
dimak1: Updated.

ChasBoucher (commented):
Looks good.

PR #1041, opened 2022-11-15
fix for comparing manual add to autocomplete

owillborn:

*Description of changes:*
- Fix for when a user manually enters the same secured party as one added from the autocomplete search/dropdown
- The address objects from the search do not have the optional fields, so this fix removes the optional fields (if empty) from the manually added party to allow them to be compared.

### DIFF
--- a/ppr-ui/package.json
+++ b/ppr-ui/package.json
@@ -29,6 +29,7 @@
     "core-js": "^3.6.5",
     "flush-promises": "^1.0.2",
     "http-status-codes": "^2.1.4",
+    "lodash-clean": "^2.2.3",
     "material-design-icons": "^3.0.1",
     "moment": "^2.29.1",
     "moment-timezone": "^0.5.34",
--- a/ppr-ui/src/components/parties/composables/useSecuredParty.ts
+++ b/ppr-ui/src/components/parties/composables/useSecuredParty.ts
@@ -4,7 +4,7 @@ import { useGetters, useActions } from 'vuex-composition-helpers'
 import { PartyAddressSchema } from '@/schemas'
 import { ActionTypes, APIRegistrationTypes, RegistrationFlowType, SecuredPartyTypes } from '@/enums'
 import { checkAddress, formatAddress } from '@/composables/address/factories/address-factory'
-import { cloneDeep, isEqual } from 'lodash'
+import { cloneDeep, isEqual, omitBy, isNil, _ } from 'lodash'
 import { useParty } from '@/composables/useParty'
 
 const initPerson = { first: '', middle: '', last: '' }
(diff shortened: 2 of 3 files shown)


cameron-eyds (approved):
LGTM! 
The cleaning method looks like it could be a generic utility method, for reuse where required.
If you feel like refactoring later to make it more global, don't hesitate to do so :)

ppr-ui/src/components/parties/composables/useSecuredParty.ts
```diff
@@ -95,6 +95,15 @@ export const useSecuredParty = (props, context) => {
     return idx !== -1
   }
 
+  const removeEmptyAttributes = (address: AddressIF): AddressIF => {
+    // removes any undefined/empty attributes from party clone so it can be compared to autocomplete objects
+    const keys = Object.keys(address)
+    keys.forEach(key => {
+      if (!address[key]) delete address[key]
```
dimak1: Is the idea here to mutate the address object? Is it intentional?

owillborn:
> 

Oooo good idea!

ppr-ui/src/components/parties/composables/useSecuredParty.ts
owillborn: Yes I want to change the clone of the address object. The addresses from the autocomplete search don't have the optional fields, but they still need to be compared for equality in a duplicate Secured Party. So removing any undefined/empty attributes allows them to be compared properly

dimak1:
@owillborn you could also explore [lodash](https://lodash.com/) methods, maybe you can find something to help with the removal of empty values.

owillborn:
> @owillborn you could also explore [lodash](https://lodash.com/) methods, maybe you can find something to help with the removal of empty values.

Ill take a look, thanks!

ppr-ui/src/components/parties/composables/useSecuredParty.ts
dimak1: ok thanks for clarification

dimak1 (approved):

owillborn:
> @owillborn you could also explore [lodash](https://lodash.com/) methods, maybe you can find something to help with the removal of empty values.

@dimak1 found something much nicer, thanks!

ppr-ui/package.json
```diff
@@ -29,6 +29,7 @@
     "core-js": "^3.6.5",
     "flush-promises": "^1.0.2",
     "http-status-codes": "^2.1.4",
+    "lodash-clean": "^2.2.3",
```
dimak1, line 32: hmm not sure about importing a library for just one function

owillborn:
@dimak1 @cameron-eyds Commit history got really messy, gonna close this and create a new one, sorry

dimak1:
@owillborn just in case, you should be doing a **rebase** not **merge** when syncing with `main` branch. I'll send you a link.

PR #1042, opened 2022-11-15
use compact to trim address objects

owillborn:

*Description of changes:*
- Fix for when a user manually enters the same secured party as one added from the autocomplete search/dropdown
- The address objects from the search do not have the optional fields, so this fix removes the optional fields (if empty) from the manually added party to allow them to be compared.


dimak1 (approved):
Nice, you found a clean way 👍🏻  yea, lodash could be very useful and have many helper methods.

cameron-eyds (approved):
GJ! 
Less is always more

PR #1047, opened 2022-11-18
14175-14177 Add code to fix issue with duplicate Secured Parties

ChasBoucher:
*Description of changes:*

- Added code to normalize address and name data to check for duplicates
- Fixed some broken tests based on changes... ie.  adding toUpperCase broke a few test.
- Fixed bugs reported in tickets #14175

### DIFF
--- a/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
+++ b/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
@@ -120,7 +120,7 @@
               </div>
             </v-col>
             <v-col cols="2" class="pt-5"
-              ><v-btn
+              ><v-btn v-if="!isExistingSecuredParty(result.code)"
                 class="ml-auto float-right"
                 color="primary"
                 :class="$style['partyButton']"
@@ -128,6 +128,9 @@
               >
                 Select
               </v-btn>
+              <span class="auto-complete-added" v-else>
+                <v-icon :class="[$style['icon-bump'], 'auto-complete-added']">mdi-check</v-icon>Added
+              </span>
             </v-col>
           </v-row>
         </v-container>
@@ -193,6 +196,10 @@ export default defineComponent({
       type: Boolean,
       default: false
     },
+    activeIndex: {
+      type: Number,
+      default: -1
+    },
     defaultIsRegisteringParty: Boolean
   },
   setup (props, context) {
@@ -223,7 +230,7 @@ export default defineComponent({
 
     const countryProvincesHelpers = useCountriesProvinces()
 
-    const { addSecuredParty, setRegisteringParty } = useSecuredParty(props, context)
+    const { addSecuredParty, setRegisteringParty, isExistingSecuredParty } = useSecuredParty(props, context)
 
     const selectParty = (idx: number) => {
       const selectedResult = localState.results[idx]
@@ -237,7 +244,7 @@ export default defineComponent({
         newParty.action = ActionTypes.EDITED
         setRegisteringParty(newParty)
       } else {
-        addSecuredParty(newParty)
+        addSecuredParty(newParty, props.activeIndex)
       }
       context.emit('emitResetClose')
     }
@@ -247,7 +254,7 @@ export default defineComponent({
         localState.party.action = ActionTypes.EDITED
         setRegisteringParty(localState.party)
       } else {
-        addSecuredParty(localState.party)
+        addSecuredParty(localState.party, props.activeIndex)
       }
       context.emit('emitResetClose')
     }
@@ -265,6 +272,7 @@ export default defineComponent({
       createParty,
       exit,
       onHover,
+      isExistingSecuredParty,
       ...countryProvincesHelpers,
       ...toRefs(localState)
     }
--- a/ppr-ui/src/components/parties/composables/useSecuredParty.ts
+++ b/ppr-ui/src/components/parties/composables/useSecuredParty.ts
@@ -4,10 +4,9 @@ import { useGetters, useActions } from 'vuex-composition-helpers'
 import { PartyAddressSchema } from '@/schemas'
 import { ActionTypes, APIRegistrationTypes, RegistrationFlowType, SecuredPartyTypes } from '@/enums'
 import { checkAddress, formatAddress } from '@/composables/address/factories/address-factory'
-import { cloneDeep, isEqual, omitBy, overSome, isNaN, isNil, isEmpty } from 'lodash'
+import { cloneDeep, isEqual } from 'lodash'
 import { useParty } from '@/composables/useParty'
-import { addRegistrationSummary } from '@/utils'
-
+import { normalizeObject } from '@/utils/validation-helper'
 const initPerson = { first: '', middle: '', last: '' }
 const initAddress = {
   street: '',
@@ -117,14 +116,6 @@ export const useSecuredParty = (props, context) => {
     const workObject2 = Object.create(object2)
     const addressEqual = isEqual(normalizeObject(workObject1), normalizeObject(workObject2))
     return addressEqual
-  }
-
-  const normalizeObject = (convertObject: any): any => {
-    convertObject = omitBy(convertObject, overSome([isNil, isNaN, isEmpty]))
-    Object.entries(convertObject).forEach(([key, value]) => {
-      if (typeof value === 'string') convertObject[key] = value?.toUpperCase().replaceAll(' ', '').trim()
-    })
-    return convertObject
   }
 
   const removeSecuredParty = (): void => {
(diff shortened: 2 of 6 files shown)


dimak1:
@ChasBoucher there's lots going on here... i've seen that @owillborn had tried to use JSON.stringify to compare objects. Maybe you go back to that and see if it's cleaner that way?

ppr-ui/src/components/parties/composables/useSecuredParty.ts
```diff
@@ -96,21 +97,36 @@ export const useSecuredParty = (props, context) => {
...
   const hasMatchingSecuredParty = (addedParty: PartyIF): boolean => {
-    // store state without newly added party
+    // store state without newly added party.
     const parties = cloneDeep(getAddSecuredPartiesAndDebtors.value.securedParties)
     if (localState.partyType === SecuredPartyTypes.INDIVIDUAL) {
       return parties.some(party =>
-        isEqual(party.personName, addedParty.personName) &&
-        isEqual(party.address, addedParty.address)
+        isObjectEqual(party.personName, addedParty.personName) &&
+        isObjectEqual(party.address, addedParty.address)
```
cameron-eyds, line 105: Odd!
`isEqual` docs state it does a proper deep comparison on objects, does it not?

https://lodash.com/docs/4.17.15#isEqual

dimak1: Yea, I would expect it to compare objects as well...

ppr-ui/src/components/parties/composables/useSecuredParty.ts
```diff
@@ -96,21 +97,36 @@ export const useSecuredParty = (props, context) => {
...
+    return addressEqual
+  }
+
+  const normalizeObject = (convertObject: any): any => {
+    convertObject = omitBy(convertObject, overSome([isNil, isNaN, isEmpty]))
+    Object.entries(convertObject).forEach(([key, value]) => {
+      if (typeof value === 'string') convertObject[key] = value?.toUpperCase().replaceAll(' ', '').trim()
+    })
+    return convertObject
+  }
```
cameron-eyds: We have a helper method called `cleanEmpty()`, this could potentially solve the same issue? 
(Although not the case conversion) 

The only concern here, this is adding in complex custom code, that isn't typed and challenging to follow without looking documentation on all the various helper imports. 

If you feel this is absolutely required, than please put it somewhere more global(utils) and with proper doc bloc comments so others can follow and use again 👍 

* The comment is for both methods, sorry!

cameron-eyds (commented):
Just a few comments!

ppr-ui/src/components/parties/composables/useSecuredParty.ts
ChasBoucher: Ill have a look at cleanEmpty...If it does'nt work I will move function to utills..

ppr-ui/src/components/parties/composables/useSecuredParty.ts
ChasBoucher: Not the way we were using it... The objects being compared were not of the proper type..  I am using isEqual in the isObjectEqual after I normalize...

I kept the normalize logic separate because it can be used for any type of simple object...

ppr-ui/src/components/parties/composables/useSecuredParty.ts
cameron-eyds: Sounds good, thanks! 

I think if you could type this and add some intuitive doc bloc comments, it would go a long way to make this more intuitive for other developers.

ppr-ui/src/components/parties/composables/useSecuredParty.ts
cameron-eyds: Fair enough, can't claim that JS is simple when it comes to types and it's (often) type coercion.

ppr-ui/src/components/parties/composables/useSecuredParty.ts
ChasBoucher: cleanEmpty gives the same functionality as omit which is a lodash function..
I do like the idea of moving to utils.

Which import would you suggest... I think this function will be used elsewhere..

cameron-eyds: `validation-helper.ts` seems like a fit?

ChasBoucher: Funny.  That's where I put it... I have checked in the code.

ChasBoucher:
> @ChasBoucher there's lots going on here... i've seen that @owillborn had tried to use JSON.stringify to compare objects. Maybe you go back to that and see if it's cleaner that way?

@dimak1 I tried JSON.stringify as well.   Issue is, it only returns equal if all the properties are in the exact order... Makes sense.  A Json object is text..

ppr-ui/src/components/parties/composables/useSecuredParty.ts
```diff
@@ -96,21 +96,28 @@ export const useSecuredParty = (props, context) => {
...
     }
   }
 
+  const isObjectEqual = (object1: any, object2: any): boolean => {
+    const workObject1 = Object.create(object1)
+    const workObject2 = Object.create(object2)
+    const addressEqual = isEqual(normalizeObject(workObject1), normalizeObject(workObject2))
+    return addressEqual
+  }
+
```
cameron-eyds: As mentioned in the first comment, thinking this is also a utility function. 
Besides the address const name it could be used for anything, though not a big deal, can be relocated in a future commit as well, whatever you prefer.

cameron-eyds (approved):
Thanks for the updates, had another small comment but otherwise looks good 👍

ppr-ui/src/components/parties/composables/useSecuredParty.ts
dimak1: Agree with Cameron, this should be moved out to a util file.

dimak1 (approved):

ppr-ui/src/components/parties/composables/useSecuredParty.ts
ChasBoucher: Ok

ChasBoucher:
The code for duplication, similar is complete.
I am working with Otto to ensure all the requirements from both 14175 and 14177 are in this pr...

ChasBoucher:
Moved isObjectEqual to utils
With Otto fixed issues in #14177

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -120,14 +120,17 @@
               </div>
             </v-col>
             <v-col cols="2" class="pt-5"
-              ><v-btn
+              ><v-btn v-if="!isExistingSecuredParty(result.code)"
```
cameron-eyds, line 123: Please keep the `>` with its tag unless its self closing (ie `/>`

ppr-ui/src/components/parties/composables/useSecuredParty.ts
```diff
@@ -164,11 +158,18 @@ export const useSecuredParty = (props, context) => {
     setAddSecuredPartiesAndDebtors(parties)
   }
 
-  const addSecuredParty = (newParty: PartyIF) => {
-    let parties = getAddSecuredPartiesAndDebtors.value // eslint-disable-line
-    let newList: PartyIF[] = parties.securedParties // eslint-disable-line
-    newParty.action = ActionTypes.ADDED
-    newList.push(newParty)
+  const addSecuredParty = (newParty: PartyIF, activeIndex: number = -1) => {
+    const parties = getAddSecuredPartiesAndDebtors.value // eslint-disable-line
```
cameron-eyds, line 162: Is there a `Parties` interface? 
If so, you could use and remove the lint suppression maybe?

cameron-eyds (commented):
Looks fine just a few minor comments 👍
My comments are not blocking a merge here.

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
cameron-eyds: *Apologies, i know this formatting probably wasn't your code to begin with but none the less 😁

cameron-eyds: Just keep this in mind, if you do see this, to fix the formatting :)

ChasBoucher: Makes sense.

ppr-ui/src/components/parties/composables/useSecuredParty.ts
ChasBoucher: Not at a base interface level...

PR #1050, opened 2022-11-22
14244 Add Document Id for drafts

ChasBoucher:
*Issue #:* /bcgov/entity###14244

*Description of changes:*

- Add document id to child rows if not a draft

### DIFF
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -29,12 +29,13 @@
           </v-btn>
         </v-col>
         <v-col style="padding-top: 2px;">
-          <p v-if="isDraft(item)" :class="{ 'ma-0': true, 'pl-9': isChild }">Pending</p>
+          <p v-if="isDraft(item)" :class="{ 'ma-0': true, 'pl-9': isChild}">Pending</p>
           <!-- child drafts will sometimes show outside their base reg during the sort -->
           <div v-if="isChild || (isDraft(item) && item.baseRegistrationNumber)" :class="isChild ? 'pl-9' : ''">
-            <p class="ma-0">{{ item.registrationNumber }}</p>
+            <p v-if="!isPpr" style="font-size: 0.875rem !important;" :class="{ 'ma-0': true }">{{ item.documentId }}</p>
+            <p v-if="isPpr" class="ma-0">{{ item.registrationNumber }}</p>
             <p class="ma-0" style="font-size: 0.75rem !important;">
-              <b>{{ (isPpr ? 'Base Registration: ' : 'MHR Number: ') + item.baseRegistrationNumber}}</b>
+              <b>{{ (isPpr ? 'Base Registration: ' : 'MHR Number: ')}}<br>{{ item.baseRegistrationNumber }}</b>
             </p>
           </div>
           <p v-else class="ma-0">
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -29,11 +29,11 @@
           </v-btn>
         </v-col>
         <v-col style="padding-top: 2px;">
-          <p v-if="isDraft(item)" :class="{ 'ma-0': true, 'pl-9': isChild}">Pending</p>
+          <p v-if="isDraft(item)" :class="{ 'ma-0': true, 'pl-9': isChild }">Pending</p>
           <!-- child drafts will sometimes show outside their base reg during the sort -->
           <div v-if="isChild || (isDraft(item) && item.baseRegistrationNumber)" :class="isChild ? 'pl-9' : ''">
-            <p v-if="!isPpr" style="font-size: 0.875rem !important;" :class="{ 'ma-0': true }">{{ item.documentId }}</p>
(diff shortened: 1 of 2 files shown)


dimak1 (approved):
Could you bump up the version please.

ppr-ui/src/components/tables/common/TableRow.vue
```diff
@@ -29,12 +29,13 @@
           </v-btn>
         </v-col>
         <v-col style="padding-top: 2px;">
-          <p v-if="isDraft(item)" :class="{ 'ma-0': true, 'pl-9': isChild }">Pending</p>
+          <p v-if="isDraft(item)" :class="{ 'ma-0': true, 'pl-9': isChild}">Pending</p>
```
cameron-eyds: Not a deal breaker but the space can stay, probably a misclick?

ppr-ui/src/components/tables/common/TableRow.vue
```diff
@@ -29,12 +29,13 @@
...
           <!-- child drafts will sometimes show outside their base reg during the sort -->
           <div v-if="isChild || (isDraft(item) && item.baseRegistrationNumber)" :class="isChild ? 'pl-9' : ''">
-            <p class="ma-0">{{ item.registrationNumber }}</p>
+            <p v-if="!isPpr" style="font-size: 0.875rem !important;" :class="{ 'ma-0': true }">{{ item.documentId }}</p>
+            <p v-if="isPpr" class="ma-0">{{ item.registrationNumber }}</p>
```
cameron-eyds: Maybe a good chance to use `v-if` and `v-else` 
v-if isPpr / v-else you doc id. 

Also, your styling in the `!isPpr` is a little off. Your bound class is always `true`? 
Also, is the `important` required? `p` default should be .875rem so if something is overriding, perhaps make a class to keep the template neat.

ppr-ui/tests/unit/TableRow.spec.ts
```diff
@@ -208,7 +208,7 @@ describe('TableRow tests', () => {
         expect(rowData.exists()).toBe(true)
         expect(rowData.length).toBe(11)
         // reg num
-        if (isChild) expect(rowData.at(0).text()).toBe(`Pending  Base Registration: ${baseReg.baseRegistrationNumber}`)
+        if (isChild) expect(rowData.at(0).text()).toBe(`Pending   Base Registration: ${baseReg.baseRegistrationNumber}`)
```
cameron-eyds: Additional spacing intentional?

cameron-eyds (commented):
Just a few comments!

ppr-ui/src/components/tables/common/TableRow.vue
ChasBoucher: Made the change...

ppr-ui/src/components/tables/common/TableRow.vue
ChasBoucher: Good point..

ppr-ui/tests/unit/TableRow.spec.ts
ChasBoucher: Yes... It takes care of the line break I added. 
The line break <br>was added to ensure there is a line break for both Ppr and registration number wraps.

cameron-eyds: Sounds good, thanks!

ppr-ui/src/components/tables/common/TableRow.vue
```diff
@@ -29,11 +29,11 @@
           </v-btn>
         </v-col>
         <v-col style="padding-top: 2px;">
-          <p v-if="isDraft(item)" :class="{ 'ma-0': true, 'pl-9': isChild}">Pending</p>
+          <p v-if="isDraft(item)" :class="{ 'ma-0': true, 'pl-9': isChild }">Pending</p>
           <!-- child drafts will sometimes show outside their base reg during the sort -->
           <div v-if="isChild || (isDraft(item) && item.baseRegistrationNumber)" :class="isChild ? 'pl-9' : ''">
-            <p v-if="!isPpr" style="font-size: 0.875rem !important;" :class="{ 'ma-0': true }">{{ item.documentId }}</p>
             <p v-if="isPpr" class="ma-0">{{ item.registrationNumber }}</p>
+            <p v-else style="font-size: 0.875rem;" :class="{ 'ma-0': true }">{{ item.documentId }}</p>
```
cameron-eyds: This class binding is still set to always `true`. 

Using class binding, ie `:class={}` is a way to conditionally add a class. In this case it's always `true` so there is no condition. 
Was the intention here to apply a condition? 
If not, you can simply add the class to the element, ie `class="ma-0"`

cameron-eyds: @ChasBoucher

ChasBoucher: I am not sure why it is there... I did'nt change the existing code other than the condition... but I will make the change.

cameron-eyds: Fair enough, all good! 
Thank you 👍

cameron-eyds (approved):

owillborn (approved):
Looks solid

PR #1055, opened 2022-11-24
Bug fixes for duplicate SP's

owillborn:

*Description of changes:*

Bug Fixes : 
- 14311 Disabled rollover on added rows in Similar SP dialog
- 14177 Refined edit mode to check duplicate before similar SP, fix small flow bugs
- 14175 Display correct dialog when comparing SP's added from autocomplete to manually added SP's

Added SP's in similar SP dialog
image

### DIFF
--- a/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
+++ b/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
@@ -98,7 +98,7 @@
         <v-container v-if="!isDuplicate">
           <v-row
             class="searchResponse"
-            :class="$style['companyRow']"
+            :class="$style[isExistingSecuredParty(result.code) ? 'companyRowNoHover' : 'companyRow']"
             v-for="(result, i) in results"
             :key="i"
             @mouseover="onHover"
@@ -139,7 +139,7 @@
             Cancelling and re-starting you registration may resolve this issue.<br>
             If you do not wish to proceed, contact BC registries staff:
           </p>
-          <error-contact :class="$style['padLeft']"/>
+          <error-contact :class="$style['pad-left']"/>
         </div>
       </div>
       <v-card-actions class="pt-6 pb-8">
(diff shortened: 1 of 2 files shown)


ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -311,6 +311,17 @@ export default defineComponent({
   border: 1px solid white;
 }
 
+.companyRowNoHover {
```
cameron-eyds: snake case as the other classes. 
for `companyrow` as well 👍

cameron-eyds (approved):
Looks good, minor comment.

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
owillborn: Kinda confused on when to use which case. The majority of classes in this file are camel case. hmmmmm

cameron-eyds: It's a valid question. The formatting is a matter of preference and consistency.
We didn't build this app originally and there is many instances of a few different ways to do things. 
This is one of those instances. 

This type of thing, we sort of try and fix as we go, so it's not on you to catch these things but the sort of thing we can spot and help with during code review.

cameron-eyds: If you feel the fix is large and quite out of scope, than by all means you can leave it but just commented so that you are aware of snake case being the preference for element identifiers (id's, classes, test-ids' etc)

owillborn: Gotcha, thanks for clarifying! I'll make that change

dimak1 (approved):
Other than outstanding comments, looks good.

ChasBoucher (approved):
Looks Good

PR #1058, opened 2022-11-25
14426 Duplicate Secured Party modal instead of error message under 'Business Legal Name'

owillborn:
*Issue #:* /bcgov/entity#14426

*Description of changes:*
- Moved duplicate secured party check after validation check so that invalid parties cannot be compared.
- Fixed bug Divya found in 14175 regarding amendments being blocked. Made changes to IsObjectEqual to handle undefined inputs.

### DIFF
--- a/ppr-ui/src/utils/validation-helper.ts
+++ b/ppr-ui/src/utils/validation-helper.ts
@@ -343,13 +343,10 @@ export function normalizeObject (convertObject: any): any {
   return convertObject
 }
 
-export function isObjectEqual (object1: any, object2: any): boolean {
+export function isObjectEqual (object1: any = null, object2: any = null): boolean {
   //
   // Does an equality between 2 objects
   //
-  if (object1 === undefined) object1 = null
-  if (object2 === undefined) object2 = null
-  // these ifs as Object.create() needs null or defined object
   const workObject1 = Object.create(object1)
   const workObject2 = Object.create(object2)
   const objectEqual = isEqual(normalizeObject(workObject1), normalizeObject(workObject2))
--- a/ppr-ui/src/utils/validation-helper.ts
+++ b/ppr-ui/src/utils/validation-helper.ts
@@ -347,6 +347,9 @@ export function isObjectEqual (object1: any, object2: any): boolean {
   //
   // Does an equality between 2 objects
   //
+  if (object1 === undefined) object1 = null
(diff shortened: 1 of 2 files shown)


ppr-ui/src/utils/validation-helper.ts
```diff
@@ -347,6 +347,9 @@ export function isObjectEqual (object1: any, object2: any): boolean {
   //
   // Does an equality between 2 objects
   //
+  if (object1 === undefined) object1 = null
+  if (object2 === undefined) object2 = null
+  // these ifs as Object.create() needs null or defined object
```
ChasBoucher: Looks good..

ChasBoucher (approved):
Looks Good

ppr-ui/src/utils/validation-helper.ts
dimak1: Not really. There are other cleaner ways to do it. We can either have default values as arguments or when creating an Object: `isObjectEqual (object1: any = null, object2: any = null)` or `Object.create(object1 || null)`

I'm just unsure if we even need to check if objects are equal if one is undefined/null. We can simply do `if (!object1 || !object2) return false`.

dimak1 (changes requested):

owillborn:
@dimak1 I think you're right about returning false if either is undefined. Only cases that one of the objects could be undefined is when comparing a business to an individual, or there are no parties in the table. Both of these cases have no outcomes where a duplicate should be detected, I'll make that change. Thank you!

owillborn:
> @dimak1 I think you're right about returning false if either is undefined. Only cases that one of the objects could be undefined is when comparing a business to an individual, or there are no parties in the table. Both of these cases have no outcomes where a duplicate should be detected, I'll make that change. Thank you!

@dimak1 looking on it now, since it is a util function and some comparisons may be made with undefined values in the future, I'm gonna go with the default values. More of a templated approach imo

cameron-eyds (approved):
Good clean up, thank you 👍 
Default values are much cleaner.

dimak1:
@owillborn I see what you are saying... that there could be a case when we need to compare both undefined objects? Hmm ok then. Thanks for the clarification.

dimak1 (approved):

doug-lovett (approved):

PR #1059, opened 2022-11-25
13848 fix mhr table child ux issues

ChasBoucher:
*Issue #:* /bcgov/entity#13848

*Description of changes:*

- Add Toast message, auto scroll when submitting(Review and Confirm)
- Includes #14398 - a very small fix to allow for Submitting Drafts

### DIFF
--- a/ppr-ui/src/components/mhrTransfers/TransferDetails.vue
+++ b/ppr-ui/src/components/mhrTransfers/TransferDetails.vue
@@ -180,7 +180,7 @@ export default defineComponent({
       declaredValue: getMhrTransferDeclaredValue.value?.toString(),
       consideration: getMhrTransferConsideration.value,
       transferDate: getMhrTransferDate.value,
-      isOwnLand: getMhrTransferOwnLand.value === null ? false : getMhrTransferOwnLand.value,
+      isOwnLand: getMhrTransferOwnLand.value || false,
       enableWarningMsg: false,
       showFormError: computed(() => localState.validateTransferDetails && !localState.isTransferDetailsFormValid)
     })
--- a/ppr-ui/src/components/mhrTransfers/TransferDetails.vue
+++ b/ppr-ui/src/components/mhrTransfers/TransferDetails.vue
@@ -180,7 +180,7 @@ export default defineComponent({
       declaredValue: getMhrTransferDeclaredValue.value?.toString(),
       consideration: getMhrTransferConsideration.value,
       transferDate: getMhrTransferDate.value,
-      isOwnLand: getMhrTransferOwnLand.value,
+      isOwnLand: getMhrTransferOwnLand.value === null ? false : getMhrTransferOwnLand.value,
       enableWarningMsg: false,
       showFormError: computed(() => localState.validateTransferDetails && !localState.isTransferDetailsFormValid)
     })
--- a/ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
+++ b/ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
@@ -95,7 +95,7 @@ export const useMhrInformation = () => {
       declaredValue: getMhrTransferDeclaredValue.value,
       consideration: getMhrTransferConsideration.value,
       transferDate: getMhrTransferDate.value,
-      ownLand: getMhrTransferOwnLand.value,
+      ownLand: getMhrTransferOwnLand.value === null ? false : getMhrTransferOwnLand.value,
       attentionReference: getMhrTransferAttentionReference.value,
       documentDescription: 'SALE OR GIFT',
       submittingParty: {
--- a/ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
+++ b/ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
@@ -95,7 +95,7 @@ export const useMhrInformation = () => {
       declaredValue: getMhrTransferDeclaredValue.value,
       consideration: getMhrTransferConsideration.value,
       transferDate: getMhrTransferDate.value,
-      ownLand: getMhrTransferOwnLand.value === null ? false : getMhrTransferOwnLand.value,
+      ownLand: getMhrTransferOwnLand.value || false,
       attentionReference: getMhrTransferAttentionReference.value,
       documentDescription: 'SALE OR GIFT',
       submittingParty: {
--- a/ppr-ui/src/utils/feature-flags.ts
+++ b/ppr-ui/src/utils/feature-flags.ts
@@ -9,8 +9,8 @@ export const defaultFlagSet: LDFlagSet = {
   'bcregistry-ui-mhr-enabled': false,
   'search-registration-number': false,
   'search-serial-number': false,
-  'mhr-ui-enabled': false,
-  'mhr-registration-enabled': false
+  'mhr-ui-enabled': true,
+  'mhr-registration-enabled': true
 }
 /**
  * The Launch Darkly client instance.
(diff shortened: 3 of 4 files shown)


ppr-ui/src/utils/feature-flags.ts
```diff
@@ -9,8 +9,8 @@ export const defaultFlagSet: LDFlagSet = {
   'bcregistry-ui-mhr-enabled': false,
   'search-registration-number': false,
   'search-serial-number': false,
-  'mhr-ui-enabled': false,
-  'mhr-registration-enabled': false
+  'mhr-ui-enabled': true,
```
dimak1: Undo these.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -408,6 +408,13 @@ export default defineComponent({
           setUnsavedChanges(false)
           // Delete the draft on successful submission
           if (getMhrInformation.value.draftNumber) await deleteMhrDraft(getMhrInformation.value.draftNumber)
+          const newItem: RegTableNewItemI = {
+            addedReg: String(getMhrInformation.value.mhrNumber),
```
dimak1: What about using `.toString()` ?

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -408,6 +408,13 @@ export default defineComponent({
...
+            addedRegParent: '',
+            addedRegSummary: null,
+            prevDraft: String(getMhrInformation.value.documentId) || ''
```
dimak1: Same as above.

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
```diff
@@ -95,7 +95,7 @@ export const useMhrInformation = () => {
       declaredValue: getMhrTransferDeclaredValue.value,
       consideration: getMhrTransferConsideration.value,
       transferDate: getMhrTransferDate.value,
-      ownLand: getMhrTransferOwnLand.value,
+      ownLand: getMhrTransferOwnLand.value === null ? false : getMhrTransferOwnLand.value,
```
dimak1: `ownLand: getMhrTransferOwnLand.value || false`

ppr-ui/src/components/mhrTransfers/TransferDetails.vue
```diff
@@ -180,7 +180,7 @@ export default defineComponent({
       declaredValue: getMhrTransferDeclaredValue.value?.toString(),
       consideration: getMhrTransferConsideration.value,
       transferDate: getMhrTransferDate.value,
-      isOwnLand: getMhrTransferOwnLand.value,
+      isOwnLand: getMhrTransferOwnLand.value === null ? false : getMhrTransferOwnLand.value,
```
dimak1: `isOwnLand: getMhrTransferOwnLand.value || false`

dimak1 (changes requested):

ppr-ui/src/components/mhrTransfers/TransferDetails.vue
ChasBoucher: ok

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
ChasBoucher: ok

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: good point...

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: done..

ppr-ui/src/utils/feature-flags.ts
ChasBoucher: oops

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: I did some testing with .toString() and it fails... 

Looked up the reason why and String(x) ensures the value of the object converted.
So I tried value.toString() and get a type error... I think String(x) is the most reliable...

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: Same as above

ppr-ui/src/components/mhrTransfers/TransferDetails.vue
dimak1: Missed this change.

dimak1 (changes requested):
One more change that was missed.

ppr-ui/src/components/mhrTransfers/TransferDetails.vue
ChasBoucher: not sure how that happened... Anyways... I have fixed the issue.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
dimak1: Hmm interesting why is it failing on type error, as `addedReg: string` is a string type.

cameron-eyds: Why hard type cast it at all? 
isn't `mhrNumber` typed and initialized as a string?

cameron-eyds: `MhRegistrationSummaryIF`

ChasBoucher: do you want me to change it?

cameron-eyds: Is the code required/doing something or this is a decoy?

If the latter, then yes i'd just go with `addedReg:  getMhrInformation.value.mhrNumber`. Why cast it to a string if it already is, perhaps i'm missing something?

ChasBoucher: Did some testing without String and ran into an issue...  I think we can get away without it, but there is another issue...I am going to fix..

ChasBoucher: To answer your question..  This code is required for locating the current record saved.  I is used to ensure that the scrolling on the table locates the correct record...

cameron-eyds: The code I was referring to was the casting of the value, not assigning the property value.

You are saying the `String()` is required? 
Seems like an underlying issue if we need to hard cast a string to a string..

ChasBoucher: Made change..

cameron-eyds: Resolved 👍

cameron-eyds (commented):
LGTM!

cameron-eyds (approved):
LGTM

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: Made change to no conversion..  Both values are strings... Not sure why I did that in the first place.

doug-lovett (approved):

dimak1 (approved):

PR #1060, opened 2022-11-28
Transfer UX Sweep

cameron-eyds:
*Issue #:* /bcgov/entity#13993

*Description of changes:*
* Lengthy UX updates as described in ticket

### DIFF
--- a/ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
+++ b/ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
@@ -7,7 +7,7 @@ import {
 } from '@/interfaces'
 import '@/utils/use-composition-api'
 
-import { isRaw, readonly, ref, toRefs, watch } from '@vue/composition-api'
+import { readonly, ref, toRefs, watch } from '@vue/composition-api'
 import { useActions, useGetters } from 'vuex-composition-helpers'
 import { ActionTypes, HomeTenancyTypes } from '@/enums'
 import { MhrCompVal, MhrSectVal } from '@/composables/mhrRegistration/enums'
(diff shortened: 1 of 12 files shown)


ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -414,7 +410,7 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
     // Update all fractional amounts for groups that already exist
     const updatedGroups = currentHomeOwnerGroups.map(group => {
       const newNumerator = (LCM / group.interestDenominator) * group.interestNumerator
-      group.interestNumerator = newNumerator
+      group.interestNumerator = Math.round(newNumerator)
```
dimak1, line 413: Why do ween to round the numerator here?

cameron-eyds: After i removed the code above, this was required to prevent fractions and retain whole numbers after edits. 

The jist, the code above prevented editing down of the `interestDenominator`. It also prevented a new denominator to be defined. 
ie a user wants to delete all groups, define a new one with a lower denominator. Same instance of editing. 

The side of effect of that codes removal was some scenarios where the number would be calculated using a denominator that was not cleanly divisible. 

So this just prevents an error state in the UI (fractional ownership).

dimak1: Ok, thank for clarification. Hopefully, it simplifies fractional ownership, as it is a complex one.

dimak1 (approved):

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
cameron-eyds: It is for sure and to be honest i'm 100% that this is the absolute best solution, since it will change numerator values if a user edits there common denominator and so on. 

That being said, it does solve some issues but there seems to be an outlandish amount of edge cases, so won't be surprised if there is more adjustment here 😓

doug-lovett (approved):
Looks good.

PR #1070, opened 2022-11-30
14244-Change-Mhr-Draft-to-display-DocumentRegistrationNumber

ChasBoucher:
*Issue #:* /bcgov/entity14244
*Description of changes:*

- Add RegistrationNumber to registration-interface
- Change DocumentId to DocumentRegistration on Child row 

### DIFF
--- a/ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
+++ b/ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
@@ -101,7 +101,7 @@ export interface MhRegistrationSummaryIF {
   changes?: MhRegistrationSummaryIF []
   hasDraft?: boolean
   documentId?: string
-  documentRegistrationNumber?:string
+  documentRegistrationNumber?: string
   expand?: boolean, // used in UI table to toggle expansion.
   expireDays?: number, // Number of days until expiry
 }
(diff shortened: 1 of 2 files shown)


ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
```diff
@@ -101,6 +101,7 @@ export interface MhRegistrationSummaryIF {
   changes?: MhRegistrationSummaryIF []
   hasDraft?: boolean
   documentId?: string
+  documentRegistrationNumber?:string
```
owillborn: Missing a space?

dimak1: We are just adding new field on top of `documentId`? Both will be used?

dimak1 (approved):

ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
ChasBoucher: They are separate fields(api returns both)... Both are used in various reports and displays and code...  In this case the MHR table child rows.  Honestly,  I am not quite sure of all the business rules around these two ids(numbers).  I know I was displaying the wrong number and it was not included in any of the interfaces...

doug-lovett (approved):
Looks good.

cameron-eyds (approved):

ppr-ui/src/interfaces/ppr-api-interfaces/registration-interfaces.ts
dimak1: Ok thanks for the clarification.

PR #1084, opened 2022-12-07
UX updates to PDF download button

dimak1:
*Issue #:* /bcgov/entity#13980

*Description of changes:*

- Update the way we display PDF and 'i' icons for large results in the Search History table

#### The change
PDF icon will be displayed only once, before the user clicks and tries to download large search report. After that, 'i' will be shown (with a updated tooltip) util report is available

#### The flow
1. PDF icon is displayed for all search results (small and large)
2. Clicking on large search results, will change the icon to a loading/spinner state for 5 seconds
3. If the report is still not available, the icon will switch to 'i' (instead of PDF)
4. Subsequent clicks will still show 'i' icon until PDF is ready for download

#### Screenshots
 
Screen Shot 2022-12-06 at 8 37 03 PM

Screen Shot 2022-12-06 at 8 37 09 PM

Screen Shot 2022-12-06 at 8 37 25 PM

Screen Shot 2022-12-06 at 8 37 34 PM

### DIFF
--- a/ppr-ui/src/components/tables/SearchHistory.vue
+++ b/ppr-ui/src/components/tables/SearchHistory.vue
@@ -95,13 +95,19 @@
                   >
                     <template v-slot:activator="{ on, attrs }">
                       <v-btn
-                        icon
                         v-if="!item.inProgress"
-                        color="primary"
+                        :icon="item.isPdfRequested"
+                        :depressed="!item.isPdfRequested"
+                        :color="item.isPdfRequested ? 'primary' : ''"
+                        :class="{ 'pdf-btn px-0 mt-n3' : !item.isPdfRequested }"
                         :loading="item.loadingPDF"
                         @click="refreshRow(item)"
                       >
-                        <v-icon color="primary" v-bind="attrs" v-on="on">
+                        <div v-if="!item.isPdfRequested" style="display: flex">
+                          <img src="@/assets/svgs/pdf-icon-blue.svg" />
+                          <span class="pl-1">PDF</span>
+                        </div>
+                        <v-icon v-else color="primary" v-bind="attrs" v-on="on">
                           mdi-information-outline
                         </v-icon>
                       </v-btn>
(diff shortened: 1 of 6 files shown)


ppr-ui/src/components/tables/SearchHistory.vue
```diff
@@ -95,13 +95,19 @@
                   >
                     <template v-slot:activator="{ on, attrs }">
                       <v-btn
-                        icon
                         v-if="!item.inProgress"
-                        color="primary"
+                        :icon="item.isPdfRequested"
+                        :depressed="!item.isPdfRequested"
```
cameron-eyds, line 100: Requirements call for a conditional box-shadow? Seems so specific!

ppr-ui/src/components/tables/SearchHistory.vue
```diff
@@ -216,12 +222,7 @@ export default defineComponent({
       }),
       searchHistory: computed(
         (): Array<SearchResponseIF> => {
-          let searchHistory = null
-          searchHistory = getSearchHistory.value
-          if (!searchHistory) {
-            return []
-          }
-          return searchHistory
+          return getSearchHistory.value || []
```
cameron-eyds, line 225: Nice clean up!

cameron-eyds (approved):
Looks good to me, nice work!
Like anything that is quickly to market, this will need some solid regression testing and time in QA. Just important to call that out to Divya and design team etc.

ppr-ui/src/components/tables/SearchHistory.vue
dimak1: Thanks

ppr-ui/src/components/tables/SearchHistory.vue
dimak1: Just trying to keep consistent with other PDF buttons (from above). Since this button switches between icon and icon+text, we need depressed only in certain scenarios.

cameron-eyds: Makes sense, thanks for the explanation!

dimak1:
@cameron-eyds agree, this needs to be tested further by QA. We only updated the icons and added a delay, the core functionality stays the same.

doug-lovett (approved):
Looks good.

owillborn (approved):

ppr-ui/src/store/actions/actions-model.ts
```diff
@@ -169,7 +169,10 @@ export const setSearchDebtorName: ActionIF = ({ commit }, debtorName: Individual
 export const setSearchHistory: ActionIF = ({ commit }, searchHistory: Array<SearchResponseIF>): void => {
   // need to set .loadingPDF so that the loader circle triggers when set
   //  - if it starts as undefined it wont trigger on change
-  for (let i = 0; i < searchHistory?.length || 0; i++) { searchHistory[i].loadingPDF = false }
+  for (let i = 0; i < searchHistory?.length || 0; i++) {
+    searchHistory[i].loadingPDF = false
+    searchHistory[i].isPdfRequested = false
+  }
```
dimak1, line 175: Had to revert to original code, as the following was failing the tests with some strange errors 🤷🏻 :
 ```
 (searchHistory || []).forEach(item => {
    item.loadingPDF = false
    item.isPdfRequested = false
  })
```
@cameron-eyds

PR #1085, opened 2022-12-08
MHR API add reg summary lienRegistrationType

doug-lovett:

*Issue #:* /bcgov/entity#14330

*Description of changes:*
Add lienRegistration to API get account summary registration info for MHR.


cameron-eyds (approved):
Looks good!

dimak1 (approved):

cameron-eyds:
Just a follow up Q: Does the type of `lienRegistrationType` matter with regards to UI or is it just a condition if this property is `truthy`, we know we there is an active lien. 

From the design, it doesn't really indicate types, just that one exists. So i'm assuming the presence of a type here would suffice for the UI.

doug-lovett:
PPR lien registration type may matter for some MHR registrations. A non-empty value is equivalent to an existence check. There may be some future registrations that can complete base on the registration type.

cameron-eyds:
> PPR lien registration type may matter for some MHR registrations. A non-empty value is equivalent to an existence check. There may be some future registrations that can complete base on the registration type.

Excellent, sounds good. Thanks for the info!

PR #1087, opened 2022-12-08
14134 UI review and confirm updates

owillborn:
*Issue #:* /bcgov/entity#14134

*Description of changes:*
- Added cases for new Location Types in review and confirm screen
- Reserve:
image
- PID Number (for both Strata and Other):
image
- Strata and Other, manual option (other will not show "Strata" before Lot and Plan):
image

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/HomeLocation/HomeLocationType.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeLocation/HomeLocationType.vue
@@ -198,7 +198,7 @@ import { useActions, useGetters } from 'vuex-composition-helpers'
 import { HomeLocationTypes } from '@/enums'
 import { HomeLocationInfo, PidNumber } from '@/components/common'
 import HomeLocationDescription from './HomeLocationDescription.vue'
-import { useInputRules, useMhrValidations } from '@/composables'
+import { useInputRules, useMhrValidations, useNewMhrRegistration } from '@/composables'
 import { MhrLocationInfoIF } from '@/interfaces'
 import { PidInfoIF } from '@/interfaces/ltsa-api-interfaces'
 /* eslint-enable no-unused-vars */
@@ -223,11 +223,14 @@ export default defineComponent({
       'getMhrRegistrationValidationModel'
     ])
     const {
-      setMhrLocation
+      setMhrLocation,
+      setIsManualLocation
     } = useActions<any>([
-      'setMhrLocation'
+      'setMhrLocation',
+      'setIsManualLocation'
     ])
 
+    const { resetLocationInfoFields } = useNewMhrRegistration()
     const { customRules, maxLength, required } = useInputRules()
     const {
       MhrCompVal,
@@ -321,7 +324,7 @@ export default defineComponent({
       for (const [key, value] of Object.entries(val)) {
         setMhrLocation({ key: key, value: value })
       }
-    })
+    }, { deep: true })
     watch(() => localState.additionalDescription, () => {
       setMhrLocation({ key: 'additionalDescription', value: localState.additionalDescription })
     })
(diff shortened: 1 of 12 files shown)


ppr-ui/src/components/mhrRegistration/HomeLocation/HomeLocationType.vue
```diff
@@ -346,15 +349,18 @@ export default defineComponent({
       localState.otherTypeOption = null
       localState.dealerManufacturerLot = ''
       localState.toggleInfoForm = false
+      setIsManualLocation(false)
+      localState.locationInfo = resetLocationInfoFields(localState.locationInfo)
```
cameron-eyds, line 353: Is the argument required?
In this scenario you are mutating locationInfo vs creating/returning/assigning a new object. 
This isn't wrong but i'm not entirely sure of the purpose, since you've explicitly defined what properties you are mutating back to empty strings. 

This argument might applicable if `resetLocationInfoFields` was spreading or iterating over each property vs explicitly reseting specific values. 

I could be missing something though, please let me know if so!

ppr-ui/src/components/mhrRegistration/ReviewConfirm/HomeLocationReview.vue
```diff
@@ -62,14 +62,98 @@
...
-          <!-- Other PID Type -->
+          <!-- Reserve -->
+          <template v-if="getMhrRegistrationLocation.otherType === HomeLocationTypes.OTHER_RESERVE">
+            <v-row no-gutters class="px-6 pt-1">
+              <v-col cols="3" class="pt-1">
+                <h3>Legal Land Description</h3>
+              </v-col>
+              <v-col v-if="getMhrRegistrationLocation.bandName || getMhrRegistrationLocation.reserveNumber"
+                cols="9" class="pt-1"
+                >
```
cameron-eyds: Minor formatting here, if it can't all fit on one line, new lines for each condition usually. Also, the `>` chevron to align with the tag.

An example but if you wanted to keep some props inline with the tag, i'm fine with that too, but it might throw the props out of alignment a bit.
Screen Shot 2022-12-08 at 1 00 18 PM

ppr-ui/src/components/mhrRegistration/ReviewConfirm/HomeLocationReview.vue
```diff
@@ -62,14 +62,98 @@
...
+
+            <!-- No PID -->
+            <v-row no-gutters v-else class="px-6 pt-1">
+              <v-col cols="3" class="pt-1">
+                <h3>Legal Land Description</h3>
+              </v-col>
+              <v-col cols="9" class="pt-1" v-if="getMhrRegistrationLocation.lot &&
+                getMhrRegistrationLocation.landDistrict &&
+                getMhrRegistrationLocation.plan"
+                >
```
cameron-eyds: Same as above

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -127,6 +132,24 @@ export const useNewMhrRegistration = () => {
...
+    location.section = ''
+    location.township = ''
+    location.range = ''
+    location.meridian = ''
+    location.parcel = ''
+    location.block = ''
+    location.exceptPlan = ''
+    return location
+  }
+
```
cameron-eyds: Related to above comment. 
You can go two ways with this method i think. 
Iterate over the argument and mutate each property OR do away with the argument and return a new object. 
This is sort of an in-between state.

cameron-eyds (commented):
REALLY nice stuff, good job. Just some minor comments 👍

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
owillborn: An iteration approach is much cleaner, thank you!

ChasBoucher (approved):
LGTM

ppr-ui/src/components/mhrRegistration/HomeLocation/HomeLocationType.vue
dimak1: Hmm could we just initialize a new empty obj `localState.locationInfo = {} as MhrLocationInfoIF`

owillborn: That doesn't work :/ the values wont reset, not totally sure why...

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -127,6 +132,13 @@ export const useNewMhrRegistration = () => {
     }
   }
 
+  const resetLocationInfoFields = (location: MhrLocationInfoIF): MhrLocationInfoIF => {
+    Object.entries(location).forEach(([key]) => {
+      location[key] = ''
+    })
+    return location
+  }
```
cameron-eyds, line 140: Just keep in mind this only works because all the properties are of String Type.
If we were to add types outside of this, would require a bit more finesse 👍

cameron-eyds (approved):
Looks good, thank you!

PR #1088, opened 2022-12-08
 use idp guid to uniquely identify users

bolyachevets:
Issue #, if available:
bcgov/entity#14596

Description of changes:
after keycloak migration to gold, sub from jwt token (i.e. keycloak guid) will be changed for users, as it is tied to a specific IDP instance, and idp guid should be used instead to uniquely identify users


doug-lovett (approved):
LGTM.

bolyachevets:
> LGTM.
@doug-lovett
Hi Doug, there is a an additional db schema update to add index (I checked previous PR by Patrick and it appears that schema changes will need to be done manually by someone on ppr side). I am also not sure why the pipeline is failing here (possibly due to the addition of db upgrade script). In any case, we'd like to merge this to dev today. Are you OK with this?

doug-lovett:
@bolyachevets Hi Andrew, yes I am fine with merging. I have manually applied the database DDL changes to the users table in DEV

PR #1096, opened 2022-12-14
14225 fix mhr parent child for sbc staff

ChasBoucher:
*Issue #:* /bcgov/entity#14225

*Description of changes:*

- Add logic to disable Open/Edit buttons on Mhr Dashboard when not a qualified supplier..  

### DIFF
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -432,8 +432,14 @@ export default defineComponent({
   },
   emits: ['action', 'error', 'freezeScroll', 'toggleExpand'],
   setup (props, { emit }) {
-    const { isRoleQualifiedSupplier, isRoleStaff, isRoleStaffSbc, isRoleStaffBcol, isRoleStaffReg
+    const { 
+      isRoleQualifiedSupplier,
+      isRoleStaff,
+      isRoleStaffSbc,
+      isRoleStaffBcol,
+      isRoleStaffReg
     } = useGetters<any>(['isRoleQualifiedSupplier', 'isRoleStaff', 'isRoleStaffSbc', 'isRoleStaffBcol', 'isRoleStaffReg'])
+    
     const {
       getFormattedDate,
       getRegistrationType,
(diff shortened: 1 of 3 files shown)


dimak1 (approved):

ppr-ui/src/components/tables/common/TableRow.vue
```diff
@@ -557,8 +564,9 @@ export default defineComponent({
     }
 
     const isEnabledMhr = (item: MhRegistrationSummaryIF) => {
-      return item.statusType === APIStatusTypes.MHR_ACTIVE && isRoleQualifiedSupplier.value &&
-        (item.registrationDescription === APIMhrDescriptionTypes.REGISTER_NEW_UNIT ||
+      return item.statusType === APIStatusTypes.MHR_ACTIVE && (isRoleQualifiedSupplier.value &&
```
dimak1: To make it cleaner, maybe we can put all roles into a separate computed property?

cameron-eyds: Also, doesn't `isRoleStaff` sort of cover all roles but `helpdesk`?
It might even cover that one too, not sure what a helpdesk account looks like but they might have staff role as well.

cameron-eyds: If not all good but worth looking into.

cameron-eyds:
Some minor linting fixes are needed 👍 
Screen Shot 2022-12-14 at 7 34 53 AM

doug-lovett (approved):
No additional comments.

ppr-ui/src/components/tables/common/TableRow.vue
ChasBoucher: @dimak1 Refactored

@cameron-eyds I found I had to include each role separately(according to account) or the button appeared... I think further to Dimas comment the other day, that roles need to be cleaned up on Dev(possibly test).. It seems that we should only have to use isRoleQualifiedSupplier...

cameron-eyds: Ok, thank you for checking. 
Hopefully we can revisit/clean up in the future.

cameron-eyds (approved):

PR #1126, opened 2023-01-04
14619 home owners add edit validation issue

owillborn:
*Issue #:* /bcgov/entity#14619

*Description of changes:*
- Fixed step 3 validation bug when user leaves add/edit form open and navigates to review and confirm
- State is no longer valid when add/edit form left open and user is autoscrolled to form with red error message and bar

image

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
@@ -519,8 +519,8 @@ export default defineComponent({
     const done = (): void => {
       // @ts-ignore - function exists
       context.refs.addHomeOwnerForm.validate()
-      setValidation(MhrSectVal.ADD_EDIT_OWNERS_VALID, MhrCompVal.OWNERS_VALID, true)
       if (localState.isHomeOwnerFormValid && localState.isAddressFormValid) {
+        setValidation(MhrSectVal.ADD_EDIT_OWNERS_VALID, MhrCompVal.OWNERS_VALID, true)
         if (props.editHomeOwner) {
           editHomeOwner(
             localState.owner as MhrRegistrationHomeOwnerIF,
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
@@ -1,8 +1,8 @@
 <template>
-  <v-card flat class="py-6 px-8 mb-5 rounded" :class="{ 'border-error-left': showTableError}">
+  <v-card flat class="py-6 px-8 mb-5 rounded" :class="{ 'border-error-left': showTableError || showReviewedError}">
     <v-row id="mhr-home-add-person">
       <v-col cols="3">
-        <label class="generic-label" :class="{ 'error-text' : showTableError}"> {{ getSidebarTitle }} </label>
+        <label class="generic-label" :class="{ 'error-text' : showTableError || showReviewedError}"> {{ getSidebarTitle }} </label>
       </v-col>
       <v-col cols="9">
         <v-form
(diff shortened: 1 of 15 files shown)


dimak1:
@owillborn appreciate the effort, but this solution is not consistent with the overall validation approach. For this particular flow we do not use the state model props for validation. We already have a prop `MhrSectVal.HOME_OWNERS_VALID` that is responsible for keeping the error state of the Home Owners step. You can set this to false (if invalid) instead of `setMhrAddEditOwnerError`, if that makes sense. Happy to discuss further.

dimak1 (changes requested):
Although this solution works, I suggest to refactor as per above comment.

owillborn:
> @owillborn appreciate the effort, but this solution is not consistent with the overall validation approach. For this particular flow we do not use the state model props for validation. We already have a prop `MhrSectVal.HOME_OWNERS_VALID` that is responsible for keeping the error state of the Home Owners step. You can set this to false (if invalid) instead of `setMhrAddEditOwnerError`, if that makes sense. Happy to discuss further.

Thanks for the discussion, looks a lot better now. Please take another look when you're free:) @dimak1

dimak1:
@owillborn you are welcome. At a glance it looks much cleaner. Good stuff!

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
```diff
@@ -499,6 +519,7 @@ export default defineComponent({
     const done = (): void => {
       // @ts-ignore - function exists
       context.refs.addHomeOwnerForm.validate()
+      setValidation(MhrSectVal.ADD_EDIT_OWNERS_VALID, MhrCompVal.OWNERS_VALID, true)
```
dimak1: We should check if form and address are valid before setting the flags to true. So this needs to be inside an if-condition below.

dimak1 (changes requested):
Looks better indeed. Just one comment.

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
owillborn: Good catch

dimak1 (approved):

cameron-eyds (approved):
Looks good to me, have you also verified the changed behaviour in the Transfer flow?

owillborn:
> Looks good to me, have you also verified the changed behaviour in the Transfer flow?

Found some minor issues in transfers, working on that now

ChasBoucher (approved):
Looks Good

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -482,9 +491,11 @@ export default defineComponent({
 
       // If transfer is valid, enter review mode
       if (localState.isValidTransfer) {
+        setValidation(MhrSectVal.ADD_EDIT_OWNERS_VALID, MhrCompVal.OWNERS_VALID, true)
         localState.isReviewMode = true
         localState.validate = false
       } else {
+        setValidation(MhrSectVal.ADD_EDIT_OWNERS_VALID, MhrCompVal.OWNERS_VALID, isGlobalEditingMode.value)
```
cameron-eyds: MHR Information flow and MH Registrations don't share a validation state, so i'm a little curious about this implementation. 
I don't believe we should need to change the MHR Validation state in our MHR Information flow. 

If you don't mind, perhaps look a bit deeper into WHY you needed to do this. I feel like the `isGlobalEditingMode` flag should* be enough to represent the state of the add edit component. Saving this state back to another validation state doesn't seem quite right 🤔

owillborn: Way cleaner now, thanks for the chat:) @cameron-eyds

cameron-eyds: Awesome! No prob at all sir 👍

PR #1127, opened 2023-01-05
14806 Add new account type and auth for SBC Staff

ChasBoucher:
*Issue #:* /bcgov/entity#14806

*Description of changes:*
- Add SBC_STAFF to AccountTypes
- Change method of authenticating SBC_STAFF to use AccountType
- Change timeout on RegistrationTable.spec.ts to 3000. Was timing out.
- Minor linting

### DIFF
--- a/ppr-ui/src/utils/auth-helper.ts
+++ b/ppr-ui/src/utils/auth-helper.ts
@@ -5,7 +5,7 @@ import { StatusCodes } from 'http-status-codes'
 import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
 
 // Interfaces, Enums
-import { AccountProductCodes, AccountProductMemberships, AccountProductRoles } from '@/enums'
+import { AccountProductCodes, AccountProductMemberships, AccountProductRoles, AccountTypes } from '@/enums'
 import {
   AccountProductSubscriptionIF, AddressIF, PartyIF, SearchPartyIF, UserProductSubscriptionIF
 } from '@/interfaces'
@@ -181,8 +181,7 @@ export async function getSbcFromAuth (): Promise<boolean> {
         if (!data) {
           return false
         }
-        const branchName = data?.branchName
-        if (branchName?.includes('Service BC')) {
+        if (data.orgType === AccountTypes.SBC_STAFF) {
           return true
         }
         return false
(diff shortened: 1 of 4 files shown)


ppr-ui/src/utils/auth-helper.ts
```diff
@@ -181,8 +181,7 @@ export async function getSbcFromAuth (): Promise<boolean> {
         if (!data) {
           return false
         }
-        const branchName = data?.branchName
-        if (branchName?.includes('Service BC')) {
+        if (data.orgType === AccountTypes.SBC_STAFF) {
```
cameron-eyds, line 184: Just to verify, this `orgType` has been set in all deployed environments? 
ie when this change goes to PROD, SBC Staff will still have their access?

cameron-eyds (approved):
Code looks fine to me! 
Just had the one question

ppr-ui/src/utils/auth-helper.ts
ChasBoucher: I am assuming so... The change has already been made to the API and I believe it is in production..

ppr-ui/tests/unit/RegistrationTable.spec.ts
```diff
@@ -278,9 +278,9 @@ describe('Test registration table with results', () => {
           startDate: null,
           status: ''
         })
-      }, 1000)
+      }, 3000)
```
dimak1, line 281: I see the comment it was timing out... What about 1.5s or 2s?

dimak1 (approved):

ppr-ui/tests/unit/RegistrationTable.spec.ts
ChasBoucher: I don't mind experimenting with different times, but I think 3 is a happy medium.  Different machines will give different results.  The main area the test is failing is locally... For example, when I changed to 1 second the test failed..  It sometimes fails with 3 seconds as well.

dimak1: hmm i see... testing UI with a 3s timeout goes beyond what unit test suppose to do, especially if it still fails sometimes (as you say). But I guess, let's monitor and adjust in the future if need be.

dimak1 (approved):

PR #1132, opened 2023-01-09
Lien Verification

cameron-eyds:
*Issue #:* /bcgov/entity#14849

*Description of changes:*
* Includes Lien Verification at Submission
* Includes errors and alert messaging on Review / Confirm
* Updated civic address default

### DIFF
--- a/ppr-ui/src/views/mhrInformation/MhrInformation.vue
+++ b/ppr-ui/src/views/mhrInformation/MhrInformation.vue
@@ -368,6 +368,10 @@ export default defineComponent({
       reviewOwners: computed(() => {
         return getMhrTransferHomeOwners.value.filter(owner => owner.action !== ActionTypes.REMOVED)
       }),
+      /** True if Jest is running the code. */
+      isJestRunning: computed((): boolean => {
+        return (process.env.JEST_WORKER_ID !== undefined)
+      }),
       attentionReference: '',
       isCompletionConfirmed: false,
       cancelOptions: unsavedChangesDialog,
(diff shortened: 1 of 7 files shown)


ChasBoucher (approved):
Looks good.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -449,6 +456,17 @@ export default defineComponent({
 
       // If already in review mode, file the transfer
       if (localState.isReviewMode) {
+        // Verify no lien exists prior to submitting filing
+        const regSum = !localState.isJestRunning
```
dimak1, line 474: Could we check for Jest at the same time we check for isReviewMode, line 458?

dimak1 (approved):
Looks good, just one suggestion.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
cameron-eyds: We could* although that would prevent all the interactions within this block, inside our unit tests. 
Example being, triggering the error state for required fields which in turn displays validation messaging etc.

doug-lovett (approved):
LGTM.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
dimak1: Sounds good, thanks for clarifying. Resolving.

cameron-eyds: No problem at all, a very valid question!

PR #1137, opened 2023-01-12
Search Dropdown Bug Fixes and UXA updates.

dimak1:
*Issues:* 
- bcgov/entity#14606
- bcgov/entity#14878

*Description of changes:*
- Search component fixes and UXA feedback updates

### DIFF
--- a/ppr-ui/src/components/search/SearchBar.vue
+++ b/ppr-ui/src/components/search/SearchBar.vue
@@ -124,7 +124,7 @@
         <v-row>
           <v-col class="py-0">
             <div
-              v-if="selectedSearchType && (shouldShowFeeHint || isMHRSearchType(selectedSearchType.searchTypeAPI))"
+              v-if="shouldShowFeeHint || (selectedSearchType && isMHRSearchType(selectedSearchType.searchTypeAPI))"
               class="ppr-mhr-info mt-5 mb-7"
             >
               <v-icon size="20">mdi-information-outline</v-icon>
@@ -305,9 +305,13 @@ export default defineComponent({
       categoryMessage: computed((): string => {
         return localState.validations?.category?.message || ''
       }),
+      isPPROnly: computed((): boolean => hasPprRole.value && !(hasMhrRole.value && getFeatureFlag('mhr-ui-enabled'))),
       shouldShowFeeHint: computed((): boolean => {
-        return (!(isRoleStaffBcol.value || isRoleStaffReg.value) &&
-          (isPPRSearchType(localState.selectedSearchType?.searchTypeAPI))) || (hasPprRole.value && !hasMhrRole.value)
+        return (
+          localState.isPPROnly ||
+          (!(isRoleStaffBcol.value || isRoleStaffReg.value) &&
+            isPPRSearchType(localState.selectedSearchType?.searchTypeAPI))
+        )
       }),
       dialogOptions: computed((): DialogOptionsIF => {
         const options = { ...paymentConfirmaionDialog }
--- a/ppr-ui/src/components/search/SearchBar.vue
+++ b/ppr-ui/src/components/search/SearchBar.vue
@@ -121,20 +121,24 @@
             </v-row>
           </v-col>
         </v-row>
-        <div
-          v-if="selectedSearchType && (shouldShowFeeHint || isMHRSearchType(selectedSearchType.searchTypeAPI))"
-          class="ppr-mhr-info mt-5 mb-7"
-        >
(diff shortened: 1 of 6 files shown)


ppr-ui/src/components/search/SearchBarList.vue
```diff
@@ -130,9 +133,10 @@ export default defineComponent({
         }
         return allSearchTypes
       }),
+      isPPROnly: computed((): boolean => hasPprRole.value && !(hasMhrRole.value && getFeatureFlag('mhr-ui-enabled'))),
       displayItems: [],
       displayGroup: {
-        1: false,
+        1: !(hasMhrRole.value && hasPprRole.value),
```
cameron-eyds: When looking i found neither `hasMhrRole` or `hasPprRole` was set when this component mounted (the timing i was referring to previously). 
Does this not always initialize `displayGroup` as `1: true` due to that?
If it works though, great 👍

cameron-eyds (approved):
LGTM. Just the one comment to verify

ppr-ui/src/components/search/SearchBarList.vue
dimak1: I found a scenario when it doesn't work, so need to re-think this.

cameron-eyds: As we discussed previously, a watcher could help with the timing.

ChasBoucher (approved):
Looks good.

ppr-ui/src/components/search/SearchBarList.vue
```diff
@@ -130,9 +133,10 @@ export default defineComponent({
         }
         return allSearchTypes
       }),
+      isPPROnly: computed((): boolean => hasPprRole.value && !(hasMhrRole.value && getFeatureFlag('mhr-ui-enabled'))),
       displayItems: [],
       displayGroup: {
-        1: false,
+        1: !(hasPprRole.value && (hasMhrRole.value && getFeatureFlag('mhr-ui-enabled'))),
```
cameron-eyds: How does this resolve the timing issue?

Also long term, keep in mind this flag will eventually be `true` always and we will remove this flag. Would this code remain true then? Shouldn't we be able to dictate these default values without a flag?

You seem opposed to my previous ideas of implementation, which is totally fine but i can't agree with use a 'temporary' feature flag to determine multiple values and conditions in this file. We should use the flags sparingly.

cameron-eyds (changes requested):
See comment

ppr-ui/src/components/search/SearchBarList.vue
dimak1: I'm not opposed, just trying to find a solution that works for all scenarios/roles/flags.

dimak1: We need to use `mhr-ui-enabled` flag otherwise the dropdown behaviour won't be consistent.

cameron-eyds: Please elaborate.

dimak1: Messaged you offline.

dimak1:
@cameron-eyds I refactored a bit and created a new getter prop using a feature-flag (for easier refactoring in the future). We need to check for the MHR feature-flag for the dropdown, just like we do for the other MHR-related functionality.
Also, as you mentioned onMounted() the role props are false, but I checked in onUpdated() they are set to true, right before the loading state is disabled, so we are good to use the roles props.

ppr-ui/src/components/search/SearchBar.vue
```diff
@@ -301,9 +305,13 @@ export default defineComponent({
       categoryMessage: computed((): string => {
         return localState.validations?.category?.message || ''
       }),
+      isPPROnly: computed((): boolean => hasPprRole.value && !(hasMhrRole.value && getFeatureFlag('mhr-ui-enabled'))),
```
cameron-eyds: I think* the && condition is the same as the new `hasMhrRoleEnabled` property, might be useful here as well.

cameron-eyds (approved):
Looks good to me!
Providing the functionality is still as expected when this flag is removed 👍

ppr-ui/src/components/search/SearchBar.vue
dimak1: Right, I meant to update it, thanks

dimak1: Updated

PR #1141, opened 2023-01-17
14937 ssp modal fixes

owillborn:
*Issue #:* /bcgov/entity#14937

Refactors from UXA/QA

*Description of changes:*
- Similar Secured Party modal will now only appear on amend if changes have been made to the form.
- Changed the badge displayed for an amended SSP

In the Amendment of a registration:
image
After clicking "Amend" and making changes, then selecting the new Secured Party from the SSP modal:
image

### DIFF
--- a/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
+++ b/ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
@@ -235,7 +235,12 @@ export default defineComponent({
 
     const countryProvincesHelpers = useCountriesProvinces()
 
-    const { addSecuredParty, setRegisteringParty, isExistingSecuredParty } = useSecuredParty(props, context)
+    const {
+      setRegisteringParty,
+      isExistingSecuredParty,
+      addEditSecuredParty,
+      currentSecuredParty
+    } = useSecuredParty(props, context)
 
     const selectParty = (idx: number) => {
       const selectedResult = localState.results[idx]
@@ -249,7 +254,8 @@ export default defineComponent({
         newParty.action = ActionTypes.EDITED
         setRegisteringParty(newParty)
       } else {
-        addSecuredParty(newParty, props.activeIndex)
+        currentSecuredParty.value = newParty
+        addEditSecuredParty()
       }
       context.emit('emitResetClose')
     }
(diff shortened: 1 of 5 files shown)


ppr-ui/src/components/parties/party/EditParty.vue
```diff
@@ -369,16 +371,18 @@ export default defineComponent({
...
-          const response: [SearchPartyIF] = await partyCodeSearch(
-            currentSecuredParty.value.businessName, false
-          )
-          // check if any results
-          if (response?.length > 0) {
-            // show secured party selection popup
-            showDialog()
-            localState.dialogResults = response?.slice(0, 50)
-            return
+          if (!isEqual(currentSecuredParty, originalSecuredParty)) {
```
cameron-eyds, line 374: Does lodash `isEqual` provide a deep comparison on any/all nested objects properties. 
It does state `objects` as supported but perhaps verify it will handle nested object properties, ie `personName`, `contact` and so on

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
```diff
@@ -259,7 +265,8 @@ export default defineComponent({
         localState.party.action = ActionTypes.EDITED
         setRegisteringParty(localState.party)
       } else {
-        addSecuredParty(localState.party, props.activeIndex)
+        currentSecuredParty.value = localState.party
```
cameron-eyds, line 268: Keep in mind, in javascript this is only a reference to the original object (localstate.party), since thats how js works with values that are not primitive. 
If that is your intention, all good, but any changes made to party would also reflect in currentSecuredParty. 

This is a blessing and curse in JS. Just keep in mind, if you wanted to decouple, you would need to make a copy of the object, not reference it 👍

cameron-eyds (approved):
Code looks good to me 👍 Just some minor comments.

ppr-ui/src/components/parties/party/EditParty.vue
owillborn: Will verify, thanks!

ppr-ui/src/components/dialogs/SecuredPartyDialog.vue
owillborn: This is the intention. I needed a reference of the info on the form in useSecuredParty.ts which updates as information is input/removed from the form. Thanks for the info!

ppr-ui/src/components/parties/party/EditParty.vue
owillborn: @cameron-eyds docs say it performs a deep comparison, also did some tests to make sure

dimak1 (approved):
Good questions in the comments. Approved.

ChasBoucher (approved):
Looks good.

PR #1143, opened 2023-01-18
Reset stepper validations on cancel

owillborn:
*Issue #:* /bcgov/entity#14702

*Description of changes:*
- When user proceeds on cancel without saving, validations are reset
- Green checkmark is no longer showing when a registration is cancelled and a new one is initiated

### DIFF
--- a/ppr-ui/src/components/common/ButtonFooter.vue
+++ b/ppr-ui/src/components/common/ButtonFooter.vue
@@ -209,7 +209,10 @@ export default defineComponent({
     }
     const handleDialogResp = (val: boolean) => {
       localState.showCancelDialog = false
-      if (!val) goToDashboard()
+      if (!val) {
+        emit('cancelProceed')
+        goToDashboard()
+      }
     }
     /** Save the draft version from data stored in the state model. */
     const saveDraft = async (): Promise<Boolean> => {
--- a/ppr-ui/package-lock.json
+++ b/ppr-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ppr-ui",
-  "version": "0.4.1",
+  "version": "0.4.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "ppr-ui",
-      "version": "0.4.1",
+      "version": "0.4.2",
       "dependencies": {
         "@bcrs-shared-components/corp-type-module": "^1.0.7",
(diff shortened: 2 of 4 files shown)


ppr-ui/src/components/common/ButtonFooter.vue
```diff
@@ -209,7 +209,10 @@ export default defineComponent({
     }
     const handleDialogResp = (val: boolean) => {
       localState.showCancelDialog = false
-      if (!val) goToDashboard()
+      if (!val) {
+        emit('cancelProceed')
```
dimak1, line 213: So we need to reset validation only when Cancel Dialog is confirmed? What about page refresh, or any other scenarios, would it be reset by itself?

ChasBoucher (approved):
Looks good.

ppr-ui/src/components/common/ButtonFooter.vue
owillborn: Ticket calls for an issue with the cancel button, I will verify the other scenarios

doug-lovett (approved):
Looks good to me.

ppr-ui/src/components/common/ButtonFooter.vue
cameron-eyds: This would reset naturally i believe, yes. 
Since we app would be re-mount the state (and subsequently validation state) would be back to the default states it is initialized with.

cameron-eyds (approved):
Looks good. 
I know it seems like an additional step by having the parent (MhrRegistration) handle this functionality but it's 100% the right approach. 
Opens the door for any future events on Cancel for other instances of ButtonFooter and keeps the functionality exclusive to MhrReg where it should be. 
Simple but good! 👍

owillborn:
> Looks good. I know it seems like an additional step by having the parent (MhrRegistration) handle this functionality but it's 100% the right approach. Opens the door for any future events on Cancel for other instances of ButtonFooter and keeps the functionality exclusive to MhrReg where it should be. Simple but good! 👍

That's how I saw it too 👍

ppr-ui/src/components/common/ButtonFooter.vue
dimak1: Thanks!

owillborn: Confirmed the validations reset in this scenario as Cameron described

PR #1144, opened 2023-01-18
14329 fix mhr table filter bugs

ChasBoucher:
*Issue #:* /bcgov/entity#14329
*Description of changes:*

- Some minor fixes for filtering.

### DIFF
--- a/ppr-ui/src/components/common/DatePicker.vue
+++ b/ppr-ui/src/components/common/DatePicker.vue
@@ -70,8 +70,8 @@ import {
 export default defineComponent({
   name: 'DatePicker',
   props: {
-    setEndDate: { type: String },
-    setStartDate: { type: String }
+    setEndDate: { type: Date },
+    setStartDate: { type: Date }
   },
   emits: ['submit'],
   setup (props, { emit }) {
--- a/ppr-ui/src/components/common/DatePicker.vue
+++ b/ppr-ui/src/components/common/DatePicker.vue
@@ -10,7 +10,7 @@
       </v-col>
       <v-col
         class="picker-title pl-4"
-        :class="{ 'picker-err': endDate === null && datePickerErr && !disableEndDate }"
+        :class="{ 'picker-err': endDate === null && datePickerErr }"
         cols="6"
       >
         Select End Date:
--- a/ppr-ui/src/components/common/DatePicker.vue
+++ b/ppr-ui/src/components/common/DatePicker.vue
@@ -26,7 +26,6 @@
       </v-col>
       <v-col cols="6">
         <v-date-picker
-          :readonly="disableEndDate"
           color="primary"
           :min="startDate ? startDate : null"
           :max="today"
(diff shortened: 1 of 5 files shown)


doug-lovett (approved):
LGTM

ppr-ui/src/components/common/DatePicker.vue
```diff
@@ -10,7 +10,7 @@
       </v-col>
       <v-col
         class="picker-title pl-4"
-        :class="{ 'picker-err': endDate === null && datePickerErr }"
+        :class="{ 'picker-err': endDate === null && datePickerErr && !disableEndDate }"
```
dimak1: Maybe we could rename the new prop to `enableEndDate` for clarity?

ppr-ui/src/components/tables/RegistrationTable.vue
```diff
@@ -3,8 +3,9 @@
     <date-picker
       v-show="showDatePicker"
       ref="datePicker"
-      :setEndDate="submittedEndDate"
+      :setEndDate=null
```
dimak1: Not to introduce more work, but maybe to clean up tech debt while at it: `setEndDate` and `setStartDate` are passing a _Date_ type object, but `endDate` and `startDate` are _String_ type props in `DatePicker`.

dimak1 (approved):
Approved, with few suggestions.

ppr-ui/src/components/tables/RegistrationTable.vue
cameron-eyds: Dima is 100% right here. 
Also, you shouldn't need to set `null` explicitly? Whats the default prop value on the component that it requires null here?

ppr-ui/src/enums/registrationTypes.ts
```diff
@@ -168,7 +168,7 @@ export enum UIRegistrationClassTypes {
 export enum APIMhrDescriptionTypes {
   REGISTER_NEW_UNIT = 'REGISTER NEW UNIT',
   SALE_OR_GIFT = 'SALE / GIFT TRANSFER',
-  CONVERTED = '** CONVERTED **'
+  CONVERTED = 'RECORD CONVERSION'
```
cameron-eyds, line 171: This change happened on the API side?

ppr-ui/src/components/tables/RegistrationTable.vue
cameron-eyds: Also, does this impact PPR registrations? I know this table is a hybrid between the two.

cameron-eyds (commented):
Looks good so far, just some minor questions!

ppr-ui/src/enums/registrationTypes.ts
ChasBoucher: Yes.. I did test.

ppr-ui/src/components/tables/RegistrationTable.vue
ChasBoucher: I made the above change and changed the type in the datepicker. I tested both ppr and mhr and any other components that use the date picker and they all work.

ppr-ui/src/components/common/DatePicker.vue
ChasBoucher: Good catch.. This code is not required after the change to allow for date range..

cameron-eyds (approved):
Looks good 👍

PR #1150, opened 2023-01-23
Tooltip for converted registrations

owillborn:
*Issue #:* /bcgov/entity#15029

*Description of changes:*
- Added tooltip for converted MHR registrations 
image

### DIFF
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -69,7 +69,23 @@
         {{ getRegistrationType(item.registrationType) }}
         <span v-if="isPpr && !isChild"> - Base Registration</span>
       </div>
-      <div v-else class="pr-2">{{ getMhrDescription(item.registrationDescription) }}</div>
+      <div v-else class="pr-2">{{ getMhrDescription(item.registrationDescription) }}
+        <v-tooltip
+        v-if="item.registrationDescription === APIMhrDescriptionTypes.CONVERTED"
+        class="pa-2"
+        content-class="top-tooltip"
+        nudge-right="2"
+        top
+        transition="fade-transition"
+      >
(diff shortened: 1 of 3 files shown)


ppr-ui/src/components/tables/common/TableRow.vue
```diff
@@ -496,6 +512,9 @@ export default defineComponent({
 
     })
 
+    const convertedTooltipText: string = 'The records for this registration were converted from paper to digital ' +
```
dimak1: Since this is a static text, why not put it straight into the template?

dimak1 (approved):
Looks good, just one suggestion.

ppr-ui/src/components/tables/common/TableRow.vue
owillborn: I could, just thought it might look a little cleaner as the template is already pretty crowded. What do you think?

dimak1: A `<template>` is meant to hold HTML structure, content, etc. anything that the user interacts with. `<script>` is for the dynamic stuff, it is used to define logic/behaviour of the component. It does sometimes include dynamic content as it might be cleaner. But for static, it should be in the template.

cameron-eyds: Definitely agree with Dima here. 
There is a place for declaring string values in the script tho, a url for example. 

In this case, the static text can go right in :)

cameron-eyds (approved):
Minor stuff otherwise LGTM

ppr-ui/src/components/tables/common/TableRow.vue
owillborn: Gotcha, thanks guys. Changes made

doug-lovett (approved):
Looks good.

ChasBoucher (approved):
Looks good.

PR #1153, opened 2023-01-24
14825 remove optional first name in search

owillborn:
*Issue #:* /bcgov/entity#14825

*Description of changes:*
-Removed "(Optional)" text for Debtor Name search First name text field for staff accounts
-Fixed icon alignment for ticket #15029
image

### DIFF
--- a/ppr-ui/src/components/search/SearchBar.vue
+++ b/ppr-ui/src/components/search/SearchBar.vue
@@ -87,7 +87,7 @@
                   filled
                   :hint="searchHintFirst"
                   persistent-hint
-                  :label="optionFirst"
+                  label="First Name"
                   v-model="searchValueFirst"
                   @keypress.enter="searchCheck()"
                 />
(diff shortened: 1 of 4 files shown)


ppr-ui/src/components/search/SearchBar.vue
```diff
@@ -87,7 +87,7 @@
                   filled
                   :hint="searchHintFirst"
                   persistent-hint
-                  :label="optionFirst"
+                  label="First Name"
```
dimak1: Just to confirm, the _Optional_ is removed from all roles? I remember it was added recently, and now removed?

owillborn: Ticket called for it to be removed from staff accounts, it was already removed from the other roles. The validation for staff was already treating it as a required field, but the "(Optional)" label was still there...

owillborn: Will double check though, does seem a bit odd

owillborn: @dimak1 Confirmed with Divya, the only case in which it should have the optional first name field is for a staff account when doing an MHR Owner Name search. Will refactor

dimak1: Great, thank you!

ChasBoucher (approved):
Looks Good

cameron-eyds (approved):

dimak1 (approved):

doug-lovett (approved):

ChasBoucher (approved):
Looks Good

PR #1155, opened 2023-01-25
Fix Dashboard Titles via Products getter (for non-staff roles)

dimak1:
*Issue #:* 
- bcgov/entity#15097

*Description of changes:*
- Issue with the headers (for non-staff roles) was due to Subscribed Products not being fetched during Dashboard load
- Products were not loading because Account Id was not set during auth load (it was being set later on by some way that I could not identify)
- To fix the issue I had to move the fetch of the Products later during loading of the application


ChasBoucher (approved):
Looks good..

ppr-ui/src/interfaces/state-interface.ts
```diff
@@ -1,6 +1,7 @@
-import { StateModelIF } from '@/interfaces'
+import { AccountModelIF, StateModelIF } from '@/interfaces'
 
 // State model example
 export interface StateIF {
   stateModel: StateModelIF
+  account: AccountModelIF
```
cameron-eyds, line 6: I don't this merits an entirely separate model, there is account related items already in our state.
You can include this property in our current stateModel, just perhaps a name which describes these properties more concisely.
Could these properties be included in `accountInformation`? How do they differ?

ppr-ui/src/interfaces/account-interface.ts
```diff
@@ -0,0 +1,6 @@
+import { AccountModelIF } from '@/interfaces'
+
+// account model
+export interface AccountIF {
+  account: AccountModelIF
+}
```
cameron-eyds, line 6: Not really required if you use our stateModel. 
Can apply an interface to the property vs a new interface for a total state.

ppr-ui/src/interfaces/store-interfaces/account-model-interface.ts
```diff
@@ -0,0 +1,9 @@
+export interface AccountModelIF {
+  currentAccount: {
+    id: number
+  }
+  currentAccountMembership: any
+  currentUser: Object
+  pendingApprovalCount: number
+  userSettings: Array<Object>
```
cameron-eyds: Are these properties used?

ppr-ui/src/store/getters/state-getters.ts
```diff
@@ -43,6 +45,10 @@ import { toRefs } from '@vue/composition-api'
 import { MhrSectVal } from '@/composables/mhrRegistration/enums'
 import { getFeatureFlag } from '@/utils'
 
+export const getAccountModel = (state: AccountIF): AccountModelIF => {
+  return state.account
+}
+
```
cameron-eyds, line 51: I feel strongly this shouldn't be separate from our stateModel unless there is some major reason why. 

Either way when adding new properties to our stateModel or creating a new model, you need to initialize the state properties in the store in order for inner working of Vue to create those reactivity hooks.

ppr-ui/src/App.vue
```diff
@@ -469,6 +456,23 @@ export default class App extends Mixins(AuthMixin) {
         }
       }
       this.setUserInfo(userInfo)
+
+      const accountId = this.getAccountModel?.currentAccount?.id || this.getAccountId
```
cameron-eyds: How do these Id's differ?

cameron-eyds (commented):
Few questions/comments around the state changes.
The changes with regards to the timing of which items are loading looks good to me 👍

ppr-ui/src/store/getters/state-getters.ts
dimak1: This model already existed in the state store, along with other models (auth, notification). I just accessed it because I needed to get `id` under `account.currentAccount`. So everything is already there and initiated.
Screen Shot 2023-01-25 at 3 30 03 PM

cameron-eyds: O man, i wonder why. 
Did it have any other accessors? 

There is account information in our state as well 😕

ppr-ui/src/App.vue
dimak1: The way account id is set for different users (staff vs non-staff) is not the same, but I could probably just use the `id` under currentAccount.

ppr-ui/src/interfaces/store-interfaces/account-model-interface.ts
dimak1: No, they are not used, but thought of keeping them in the IF. I could remote to make it cleaner.

cameron-eyds: Totally, the interface is a welcome addition, i'm just stumped as to the purpose of the segregated state 🤷

ppr-ui/src/interfaces/account-interface.ts
dimak1: stateModel does not have an account id on the initial load.

Screen Shot 2023-01-25 at 3 37 14 PM

cameron-eyds (approved):
Thanks for clarifying that this 'side' state already existed, i didn't even realize. 
This is some welcome cleanup to that work, although i'm not sure why it was external in the first place, especially if not used.
Your work is good to go though! Thanks for adding the interfaces 👍

ppr-ui/src/interfaces/account-interface.ts
cameron-eyds: Sorry, i don't follow the meaning in the context of this comment specifically.
`null` is still a valid type by which to initialize a property.

dimak1:
@cameron-eyds thanks for the comments. 
Just to further clarify. The way things are set during initial load is not entirely clear to me, especially when using `sbc-common-components`. `accountId` is not set until sometime later (after Dashboard mounted), so I checked the entire app's state and found there is an `account` model that has the `id`. So I just used that, since it's already available. It might not be the ideal solution, but trying to step thru the code (to catch the right timing) was taking too long.

cameron-eyds:
> @cameron-eyds thanks for the comments. Just to further clarify. The way things are set during initial load is not entirely clear to me, especially when using `sbc-common-components`. `accountId` is not set until sometime later (after Dashboard mounted), so I checked the entire app's state and found there is an `account` model that has the `id`. So I just used that, since it's already available. It might not be the ideal solution, but trying to step thru the code (to catch the right timing) was taking too long.

Totally understand, probably a bit to unpack there. Thank you for looking into it 👍

ppr-ui/src/interfaces/state-interface.ts
dimak1: This is a different state model that is set by `SbcSignin` common component (set during sign on). It is similar and some props are the same (within our app's state), but it set earlier and we have access during Dashboard load.

doug-lovett (approved):
No additional comments, looks good.

owillborn (commented):
LGTM

ChasBoucher (approved):
Looks Good

ChasBoucher (approved):
Looks good

ChasBoucher (approved):
Looks Good

PR #1164, opened 2023-01-31
14950 bc reg staff payment updates for transfer

ChasBoucher:
*Issue #:* /bcgov/entity#14950

*Description of changes:*

- Add MHr Transfer options for BC Reg Staff

*** Please note that we are still unable to save a transfer if staff, because of missing document id.   In Standup this morning(2023-01-31), there will be a change made in the api to automatically assign the document Id....

### DIFF
--- a/ppr-ui/src/components/tables/common/TableRow.vue
+++ b/ppr-ui/src/components/tables/common/TableRow.vue
@@ -506,11 +506,9 @@ export default defineComponent({
         return props.setItem
       }),
       enableOpenEdit: computed(() => {
-        return isRoleQualifiedSupplier.value &&
-          !isRoleStaff.value &&
+        return (isRoleQualifiedSupplier.value || isRoleStaffReg.value || isRoleStaff.value) &&
           !isRoleStaffSbc.value &&
-          !isRoleStaffBcol.value &&
-          !isRoleStaffReg.value
+          !isRoleStaffBcol.value
       })
 
     })
--- a/ppr-ui/src/views/mhrInformation/MhrInformation.vue
+++ b/ppr-ui/src/views/mhrInformation/MhrInformation.vue
@@ -412,7 +412,7 @@ export default defineComponent({
 
     // Get Account Info from Auth to be used in Submitting Party section in Review screen
     const getAccountInformation = async (): Promise<AccountInfoIF> => {
-      return isRoleQualifiedSupplier.value === true ? getAccountInfoFromAuth() : {} as AccountInfoIF
+      return isRoleQualifiedSupplier.value ? getAccountInfoFromAuth() : {} as AccountInfoIF
     }
 
     const parseMhrInformation = async (): Promise<void> => {
@@ -445,7 +445,8 @@ export default defineComponent({
     }
 
     const parseSubmittingPartyInfo = (): void => {
-      var submittingParty = {} as SubmittingPartyIF
+      let submittingParty = {} as SubmittingPartyIF
+
       if (isRoleQualifiedSupplier.value) {
         submittingParty.businessName = localState.accountInfo?.name
         submittingParty.address = localState.accountInfo?.mailingAddress
@@ -454,17 +455,19 @@ export default defineComponent({
         submittingParty.phoneExtension = localState.accountInfo?.accountAdmin?.phoneExtension
       } else {
         submittingParty = getMhrTransferSubmittingParty.value
-        const localAccountInfo = {} as AccountInfoIF
-        localAccountInfo.name = getMhrTransferSubmittingParty.value.businessName
-        localAccountInfo.mailingAddress = getMhrTransferSubmittingParty.value.address
-        localAccountInfo.accountAdmin = {
-          firstName: getMhrTransferSubmittingParty.value.personName.firstName,
-          lastName: getMhrTransferSubmittingParty.value.personName.LastName,
+        localState.accountInfo.name = getMhrTransferSubmittingParty.value.businessName
+        localState.accountInfo.mailingAddress = getMhrTransferSubmittingParty.value.address
+        if (getMhrTransferSubmittingParty.value.businessName) {
+          localState.accountInfo.isBusinessAccount = true
+          localState.accountInfo.name = getMhrTransferSubmittingParty.value.businessName
+        }
+        localState.accountInfo.accountAdmin = {
+          firstName: getMhrTransferSubmittingParty.value.personName?.firstName,
+          lastName: getMhrTransferSubmittingParty.value.personName?.LastName,
           email: getMhrTransferSubmittingParty.value.emailAddress,
           phone: getMhrTransferSubmittingParty.value.phoneNumber,
           phoneExtension: getMhrTransferSubmittingParty.value.phoneExtension
         }
-        localState.accountInfo = localAccountInfo
       }
       setMhrTransferSubmittingParty(submittingParty)
     }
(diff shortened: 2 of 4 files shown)


ppr-ui/src/components/tables/common/TableRow.vue
```diff
@@ -506,11 +506,9 @@ export default defineComponent({
         return props.setItem
       }),
       enableOpenEdit: computed(() => {
-        return isRoleQualifiedSupplier.value &&
-          !isRoleStaff.value &&
+        return (isRoleQualifiedSupplier.value || isRoleStaffReg.value || isRoleStaff.value) &&
           !isRoleStaffSbc.value &&
```
cameron-eyds, line 510: No action required here, oddly enough this check is required because `isRoleStaff` includes SBC Staff.
Would be scary to change something like this now but just what i noticed recently in other work.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -399,16 +403,16 @@ export default defineComponent({
...
       localState.loading = false
 
       localState.dataLoaded = true
     })
 
     // Get Account Info from Auth to be used in Submitting Party section in Review screen
-    const getSubmittingPartyInformation = async (): Promise<AccountInfoIF> => {
-      return getAccountInfoFromAuth()
+    const getAccountInformation = async (): Promise<AccountInfoIF> => {
+      return isRoleQualifiedSupplier.value === true ? getAccountInfoFromAuth() : {} as AccountInfoIF
```
cameron-eyds: Condition can be simplified to `!!isRoleQualifiedSupplier.value` or even `isRoleQualifiedSupplier.value` if the return type is already a boolean.

So for staff in the this instance, would they not want to pull in the submitting party from the Reg?
I'm assuming this is a separate (future) ticket?

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -417,7 +421,8 @@ export default defineComponent({
 
       // Store a snapshot of the existing OwnerGroups for baseline of current state
       await setMhrTransferCurrentHomeOwnerGroups(cloneDeep(data.ownerGroups))
-
+      // Store existing submitting party to be used if user is BC Registry Staff
+      setMhrTransferSubmittingParty(data.submittingParty)
```
cameron-eyds, line 425: O i see, i guess this answers my question from above, this is setting the submitting party to store from the existing Reg?

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -440,13 +445,27 @@ export default defineComponent({
...
 
     const parseSubmittingPartyInfo = (): void => {
-      const submittingParty = {
-        businessName: localState.accountInfo.name,
-        address: localState.accountInfo.mailingAddress,
-        emailAddress: localState.accountInfo.accountAdmin.email,
-        phoneNumber: localState.accountInfo.accountAdmin.phone,
-        phoneExtension: localState.accountInfo.accountAdmin.phoneExtension
-      } as SubmittingPartyIF
+      var submittingParty = {} as SubmittingPartyIF
```
cameron-eyds: The main reason we use `let` over `var` is that you have more granular control of the scope of the blocked variable. 
ie, by using var here, it is scoped to the entirety of the function. Although that may not be an issue here, in more complex functions it very well could be. 
It just best practices to use `let` to scope your variable to the wrapping `{}` vs the function block as a whole.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
cameron-eyds: Oh bit confused, you are setting the submitting party down below in the `parseSubmittingParty` 
Why here too?

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: See comment below..

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: Yes..

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -440,13 +445,27 @@ export default defineComponent({
...
+        localAccountInfo.name = getMhrTransferSubmittingParty.value.businessName
+        localAccountInfo.mailingAddress = getMhrTransferSubmittingParty.value.address
+        localAccountInfo.accountAdmin = {
+          firstName: getMhrTransferSubmittingParty.value.personName.firstName,
+          lastName: getMhrTransferSubmittingParty.value.personName.LastName,
+          email: getMhrTransferSubmittingParty.value.emailAddress,
+          phone: getMhrTransferSubmittingParty.value.phoneNumber,
+          phoneExtension: getMhrTransferSubmittingParty.value.phoneExtension
+        }
+        localState.accountInfo = localAccountInfo
```
cameron-eyds: Bit confused whats happening here.
Why are we mapping the submitting party data to `accountInfo`? Appears we are assigning the same data set to both submitting party(state) and accountInfo (local). 

Also, i don't believe you need a const, should just be able to map directly to `localState.accountInfo`
ie `localState.accountInfo = { <mapping here> }`

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: Changed line 415.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: Makes sense... I made the change...

ppr-ui/src/views/mhrInformation/MhrInformation.vue
ChasBoucher: I will enter a comment in code to explain.  

I only added this code because we are unable to save a transfer without a  Submitting Party... 

The  AcccountInfo component is being used to display submitting party on the review section.  I believe this was done because we were using the auth api  to get account information to determine if user is a Qualified supplier.  

The code above ensures that the account info component is populated so it is displayed...

ChasBoucher: There is supposed to be another ticket to allow for editing submitting party, which I believe will require this code to be refactored.

ChasBoucher: I made the change to localState.accountInfo as suggested.

doug-lovett (approved):

ppr-ui/src/components/tables/common/TableRow.vue
ChasBoucher: ok

dimak1:
@ChasBoucher I don't see any new commits, but the comments say the changes were made... maybe haven't been pushed yet. Will review it after the updates.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
dimak1: Not sure if we want to return an empty object `{}` for non-qualified-suppliers. Maybe `null` is better because of the check we have.

dimak1 (commented):
Just few comments to resolve. Will take another look after.

cameron-eyds (approved):
Looks good, thanks for the updates 👍

dimak1 (approved):
Thanks for updates. Looks much cleaner now.

PR #1165, opened 2023-02-01
14335 add UI column sort

owillborn:
*Issue #:* /bcgov/entity#14335

*Description of changes:*
- Added column sort for MHR Registration Table
- "Username" and "Owners" sort query is not returning a properly sorted list, believed to be a data issue. Implementation for sorting these columns is in place and should function once this is resolved.
image

### DIFF
--- a/ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
+++ b/ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
@@ -236,7 +236,7 @@ export const useNewMhrRegistration = () => {
     const draftFilings = await getMhrDrafts()
     const myMhrHistory = await mhrRegistrationHistory(true, sortOptions)
     const filteredMhrHistory = addHistoryDraftsToMhr(myMhrHistory, draftFilings, sortOptions)
-    setMhrTableHistory(myMhrHistory)
+    setMhrTableHistory(filteredMhrHistory)
   }
 
   function addHistoryDraftsToMhr (
@@ -246,11 +246,9 @@ export const useNewMhrRegistration = () => {
     MhRegistrationSummaryIF[] {
     const sortedDraftFilings = orderBy(mhrDrafts, ['createDateTime'], ['desc'])
 
-    const sortedMhrHistory = orderBy(mhrHistory, ['createDateTime'], ['desc'])
-
     var mhrTableData = []
     // add drafts to Registrations.
-    sortedMhrHistory.forEach(transfer => {
+    mhrHistory.forEach(transfer => {
       transfer.baseRegistrationNumber = transfer.mhrNumber
       //
       // Prepare existing changes
@@ -297,7 +295,7 @@ export const useNewMhrRegistration = () => {
         transfer.changes = orderBy(transfer.changes, ['createDateTime'], ['desc'])
       }
     })
-    if (sortOptions?.status !== mhApiStatusTypes.DRAFT) mhrTableData = sortedMhrHistory
+    if (sortOptions?.status !== mhApiStatusTypes.DRAFT) mhrTableData = mhrHistory
     return mhrTableData
   }
   /**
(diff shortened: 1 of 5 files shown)


ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -236,7 +236,7 @@ export const useNewMhrRegistration = () => {
     const draftFilings = await getMhrDrafts()
     const myMhrHistory = await mhrRegistrationHistory(true, sortOptions)
     const filteredMhrHistory = addHistoryDraftsToMhr(myMhrHistory, draftFilings, sortOptions)
-    setMhrTableHistory(filteredMhrHistory)
+    setMhrTableHistory(myMhrHistory)
```
cameron-eyds: is `filteredMhrHistory` still used? 
How does it handle the draft filings now?

ppr-ui/src/resources/tableHeaders.ts
```diff
@@ -441,56 +441,56 @@ export const registrationTableHeaders: Array<BaseHeaderIF> = [
...
     value: 'mhrNumber',
     display: true
   },
   {
     class: 'min-column-md',
-    sortable: false,
+    sortable: true,
     text: 'Registration Type',
-    value: 'registrationDescription',
+    value: 'registrationType',
```
cameron-eyds: Was this an api change? 
Or is it being mapped in the UI?

ppr-ui/src/resources/tableHeaders.ts
```diff
@@ -441,56 +441,56 @@ export const registrationTableHeaders: Array<BaseHeaderIF> = [
 export const mhRegistrationTableHeaders: Array<BaseHeaderIF> = [
   {
     class: 'min-column-md',
-    sortable: false,
+    sortable: true,
```
cameron-eyds, line 444: Same question for all here, using the Vuetify table sorting out of the box works?
No mapped value issues?

ppr-ui/src/resources/tableHeaders.ts
owillborn: API changed to use registrationType for sort in ticket 14336

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
ChasBoucher: Good questions.. I am wondering the same thing..

ppr-ui/src/resources/tableHeaders.ts
cameron-eyds: Copy that, thank you.
Does the API still have `registrationDescription` property available then? Otherwise i would have expected the table/values to have broken

ChasBoucher (approved):
Other than Cameron's comments.. 
Looks good.

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
owillborn: Working on a fix for dealing with drafts, I missed this

ppr-ui/src/resources/tableHeaders.ts
owillborn: I believe it uses the Vuetify sortable to make the header clickable and display the sort direction arrow, but the actual sort is done by orderBy in script. Please correct me if I'm wrong, but seems to be no mapping issues!

ppr-ui/src/resources/tableHeaders.ts
owillborn: I'll take a look

owillborn: The table did break as it was looking for registrationDescription but I made some changes in the template to fix that

owillborn: Just to be safe, I'm going to revert to `registrationDescription` as I see it used in other parts of the API. Going to pass `registrationType` in place of `registrationDescription` to the sort though (as done with `registeringParty` and `ownerNames` in `mhr-api-helper.ts` lines 608-612)

ppr-ui/src/resources/tableHeaders.ts
cameron-eyds: Sorry yes, i wasn't very clear, you are correct! This enables the button -> in turn the event -> in turn the function. 
Was definitely referring to the sort function itself, but if it works as is, thats great!

ppr-ui/src/resources/tableHeaders.ts
cameron-eyds: Sounds good, thanks for confirming 👍

cameron-eyds (approved):
Thanks for the updates/confirmations 👍

dimak1 (approved):

PR #1167, opened 2023-02-02
Ppr/Mhr Search UXA

cameron-eyds:
*Issue #:* /bcgov/entity###

*Description of changes:*
* Misc uxa and code clean up

### DIFF
--- a/ppr-ui/src/components/search/SearchBarList.vue
+++ b/ppr-ui/src/components/search/SearchBarList.vue
@@ -138,12 +138,15 @@ export default defineComponent({
         }
       }),
       isSingleSearchOption: computed((): boolean => {
-        return (hasPprEnabled.value && !hasMhrEnabled.value) || (!hasPprEnabled.value && hasMhrEnabled.value)
+        return ((hasPprEnabled.value && !hasMhrEnabled.value) || (!hasPprEnabled.value && hasMhrEnabled.value)) &&
+          !(isRoleStaff.value || isRoleStaffReg.value)
       }),
       displayItems: [],
       displayGroup: {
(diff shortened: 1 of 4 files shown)


doug-lovett (approved):
Looks good,

ppr-ui/src/components/search/SearchBarList.vue
```diff
@@ -138,12 +138,15 @@ export default defineComponent({
         }
       }),
       isSingleSearchOption: computed((): boolean => {
-        return (hasPprEnabled.value && !hasMhrEnabled.value) || (!hasPprEnabled.value && hasMhrEnabled.value)
+        return ((hasPprEnabled.value && !hasMhrEnabled.value) || (!hasPprEnabled.value && hasMhrEnabled.value)) &&
```
dimak1, line 141: Looks like we'll be using these combo flags a lot, maybe we can combine them in a new PR.

dimak1 (approved):

dimak1:
Is this the fix for the bug or simply refactor? @cameron-eyds

ppr-ui/src/components/search/SearchBarList.vue
cameron-eyds: Definitely!

cameron-eyds:
> Is this the fix for the bug or simply refactor? @cameron-eyds

A bit of both!
UX review called for some updated behaviour on when the MHR message shows. 
The refactor also fixes the condition for when Staff see the messaging. With the previous condition i put in, it was displaying when the staff accounts had the products as well, which wasn't considered originally.

ChasBoucher (approved):
Looks good.

PR #1169, opened 2023-02-03
App Session Timing Bug

cameron-eyds:
*Issue #:* /bcgov/entity#15220

*Description of changes:*
* Fixes minor bug that would impact SBC Staff accounts, attempting to request for Authentications before the Session storage would be propagated from the Common components sign in to the app session storage.
* Implemented 1s delay to allow for this propagation since the component which populates the session is a common/shared external component.

### DIFF
--- a/ppr-ui/src/App.vue
+++ b/ppr-ui/src/App.vue
@@ -271,9 +271,11 @@ export default class App extends Mixins(AuthMixin) {
         }
       }
 
-      // if we are already authenticated then go right to init
+      // When we are authenticated, allow time for session storage propagation from auth, then initialize application
       // (since we won't get the event from Signin component)
-      if (this.isAuthenticated) this.onProfileReady(true)
+      if (this.isAuthenticated) {
+        setTimeout(() => { this.onProfileReady(true) }, this.isJestRunning ? 0 : 1000)
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


owillborn (approved):
Looks good. Nice and simple

doug-lovett (approved):
Very nice.

ppr-ui/src/App.vue
```diff
@@ -271,9 +271,11 @@ export default class App extends Mixins(AuthMixin) {
         }
       }
 
-      // if we are already authenticated then go right to init
+      // When we are authenticated, allow time for session storage propagation from auth, then initialize application
       // (since we won't get the event from Signin component)
-      if (this.isAuthenticated) this.onProfileReady(true)
+      if (this.isAuthenticated) {
+        setTimeout(() => { this.onProfileReady(true) }, this.isJestRunning ? 0 : 1000)
```
dimak1, line 277: Is 1 sec a min we could go, have you tried a lower number? Not a big deal but maybe we could save 'some' time 😃

dimak1 (approved):

ppr-ui/src/App.vue
cameron-eyds: hahah potentially! I didn't test every milli increment and this could vary between environments as well.
I promise you, 1s wasn't noticeable 😆

ChasBoucher (approved):
Looks good.

ppr-ui/src/App.vue
dimak1: 😆 haha ok sounds good!

PR #1180, opened 2023-02-13
13280 remove service fee for non clients

owillborn:
*Issue #:* /bcgov/entity#13280

*Description of changes:*
- Changes to only display service fee and staff processing fee when they are greater than 0
image

- Minor UXA changes for 15030 (wording on certified checkbox text)


dimak1 (approved):

doug-lovett (approved):
Looks fine.

cameron-eyds (approved):
Code looks fine, no impact on PPR? 
Otherwise LGTM!

owillborn:
> Code looks fine, no impact on PPR? Otherwise LGTM!

@cameron-eyds good call, I'll take a look

dimak1 (approved):

doug-lovett (approved):
Looks good to me. Thanks for the screen shot.

cameron-eyds (approved):
Code looks fine to me.
This splits PPR into its own template section? 
Once again, since there is obvious impact here to existing users, will really need to give it a good regression test and inform Divya as well :) 

*Minor version conflicts to tackle

ppr-ui/src/composables/fees/FeeSummary.vue
```diff
@@ -132,6 +132,29 @@
...
+        <li
+          v-else-if="hasProcessingFee && feeSummary.processingFee > 0"
+          id="processing-fee-summary"
+          :class="[$style['fee-container'], $style['fee-list__item'], 'pb-4', 'pr-4', 'py-4']"
+        >
+          <div :class="$style['fee-list__item-name']">
+            Staff Processing Fee
+          </div>
+          <div :class="$style['fee-list__item-value']">
+            ${{ feeSummary.processingFee.toFixed(2) }}
```
dimak1, line 144: Hmm not sure about `toFixed()` here. It will round the number and could cause some unexpected results. I know this is not new, but do you know why we need to use it?

cameron-eyds: It just applies the decimals, if i had to guess. 
ie i don't believe `$25.00` is the value applied, it's probably `$25`. 
My guess is that it was to satisfy design, not necessarily do any rounding.

owillborn:
> Code looks fine to me. This splits PPR into its own template section? Once again, since there is obvious impact here to existing users, will really need to give it a good regression test and inform Divya as well :)
> 
> *Minor version conflicts to tackle

@cameron-eyds Yes essentially creates a PPR specific template. Simplest way I could think of to leave original functionality in PPR but change it for MHR

PR #1182, opened 2023-02-14
Hotfix for PPR Debtor DOB

dimak1:
*Issue #:* 
- bcgov/entity#15375

*Description of changes:*

- Quick fix to a PPR Debtor DOB if the browser cannot parse the date properly
- Would need to implement a proper solution

*Solution*

1. Check if the browser cannot create a Date instance from date string
2. Shorten `1/1/2000T00:00:00-08:00` to this -> `1/1/2000`
3. Then it can properly create a Date instance

*References*

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

**Before / After (Firefox)**

Screen Shot 2023-02-14 at 12 43 23 PM

Screen Shot 2023-02-14 at 12 43 03 PM

### DIFF
--- a/ppr-ui/src/composables/useParty.ts
+++ b/ppr-ui/src/composables/useParty.ts
@@ -21,7 +21,11 @@ export const useParty = () => {
 
   const getFormattedBirthdate = (party: PartyIF): string => {
     if (party.birthDate) {
-      const date = new Date(party.birthDate)
+      let date = new Date(party.birthDate)
+      // prod hotfix to check if browser does not support locales
+      if (!(date instanceof Date && !isNaN(date.valueOf()))) {
+        date = new Date(party.birthDate.substring(0, party.birthDate.indexOf('T')))
+      }
       return (
         date.toLocaleString('default', { month: 'long' }) +
         ' ' +


doug-lovett (approved):
Fast work.

ppr-ui/src/composables/useParty.ts
```diff
@@ -21,7 +21,11 @@ export const useParty = () => {
 
   const getFormattedBirthdate = (party: PartyIF): string => {
     if (party.birthDate) {
-      const date = new Date(party.birthDate)
+      let date = new Date(party.birthDate)
+      // prod hotfix to check if browser does not support locales
+      if (!(date instanceof Date && !isNaN(date.valueOf()))) {
+        date = new Date(party.birthDate.substring(0, party.birthDate.indexOf('T')))
+      }
```
cameron-eyds, line 28: Condition looks fine to me!
With regards to the conversion, future state could look into `convertDate` in `date-helper.ts`. 
Not sure if applicable but worth a look at a later date. 

Does this date value format suffice for any api validations on this property? ie this conversion files without issue?

cameron-eyds (approved):
LGTM!
Just a comment/question

ppr-ui/src/composables/useParty.ts
dimak1: I haven't tried the full flow. This fix only for the UI formatting/display. We do not update the date in the `party.birthDate`.

cameron-eyds: Sounds good. 
I recall Doug saying the data looked acceptable in the db, so if we don't mutate the original values, should be low risk!

PR #1188, opened 2023-02-15
Group Tenancy Types and Previous Owners

cameron-eyds:
*Issue #:* /bcgov/entity#14360

*Description of changes:*
* Updated handling for removed owners
* Updated Group assignment logic (ie we always start at 1 in the UI, handling sequentially by active groups)
* Update group tenancy types
* Added(but no implemented) custom SVGs to be used in Transfers and Registrations
* Misc UXA updates


ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -401,6 +402,12 @@ export default defineComponent({
       return hasNoOwners && index === 0
     }
 
+    const getGroupNumberById = (groupId: number): number => {
+      const activeOwnerGroups = getTransferOrRegistrationHomeOwnerGroups()
+        .filter(group => group.action !== ActionTypes.REMOVED)
+      return (activeOwnerGroups.findIndex(group => group.groupId === groupId)) + 1
+    }
+
```
owillborn, line 410: could this go into utils?

owillborn (approved):
LGTM one question in comments!

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
cameron-eyds: It's not really generic functionality. 
It's pretty specific to finding groups by ID, amongst a collection of active groups. 

I suppose it could be refactored down to take various arguments to replace the data set we're searching and the property we're searching for but i don't see a lot of value. 

If you feel strongly about this though, can do!

owillborn: Nope, you're correct! Thanks for clarifying, was mostly just curious :)

cameron-eyds: No prob! It's a totally valid consideration, something we should always be thinking about. 

I'm guilty of also forgetting to search through all the utils we have. Often we will create stuff locally to components or views when there is functionality existing external that could do the trick!

doug-lovett (approved):
Looks good.

dimak1 (approved):
Good stuff.

PR #1198, opened 2023-02-24
15413 staff payment component for transfer

owillborn:
*Issue #:* /bcgov/entity#15413

*Description of changes:*
- Added Staff Payment Component to transfer flow
- Added validations for Staff Payment


cameron-eyds (approved):
LGTM, can you confirm all options/arguments included in the request?

owillborn:
> LGTM, can you confirm all options/arguments included in the request?

will do

doug-lovett (approved):
Looks good.

ppr-ui/src/utils/mhr-api-helper.ts
```diff
@@ -407,9 +407,10 @@ export async function validateDocumentID (documentId: string) {
   }
 }
 
-export async function submitMhrTransfer (payloadData, mhrNumber) {
+export async function submitMhrTransfer (payloadData, mhrNumber, staffPayment) {
+  const paymentParams = `?${mhrStaffPaymentParameters(staffPayment)}`
   try {
-    const result = await axios.post(`transfers/${mhrNumber}`, payloadData, getDefaultConfig())
+    const result = await axios.post(`transfers/${mhrNumber}${paymentParams}`, payloadData, getDefaultConfig())
```
cameron-eyds, line 413: This will always apply the paymentParams but i'm guessing thats intentional? 
Not really an issue but should we check for `staffPayment` perhaps before applying?

owillborn: Intentional yes, would be beneficial to check staffPayment before though. Tested with no payment info and it formats the URL correctly, but I will add a check for safety's sake

cameron-eyds: I guess we do it in the `mhrStaffPaymentParameters` method itself, so i suppose it's ok.

PR #1204, opened 2023-03-01
fee summary updates on transfers

owillborn:
*Issue #:* /bcgov/entity#15295

*Description of changes:*
- Adjusted fee summary on transfers according to spec
- Fee summary hides on cancel

With no transfer type selected:
image

With transfer type selected:
image

### DIFF
--- a/ppr-ui/src/composables/fees/FeeSummary.vue
+++ b/ppr-ui/src/composables/fees/FeeSummary.vue
@@ -12,6 +12,9 @@
         >
           <div :class="$style['fee-list__item-name']">
             {{ feeLabel }}
+            <div v-if="isMhrTransfer" :class="[$style['fee-list__hint'], 'pt-2']">
+              {{ transferType ? transferType : 'Select Transfer Type' }}
+            </div>
           </div>
           <div
             v-if="feeSummary.feeAmount === 0"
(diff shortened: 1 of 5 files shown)


ppr-ui/src/composables/fees/FeeSummary.vue
```diff
@@ -223,6 +236,7 @@ export default defineComponent({
     const localState = reactive({
       feeType: props.setFeeType,
       registrationType: props.setRegistrationType,
+      transferType: props.setTransferType,
```
cameron-eyds: I know this is what has been done elsewhere with props in some older components but unless you are going to mutate this prop value locally, you can use it directly.

ppr-ui/src/composables/fees/FeeSummary.vue
```diff
@@ -352,6 +369,9 @@ export default defineComponent({
     watch(() => props.setRegistrationType, (val: UIRegistrationTypes) => {
       localState.registrationType = val
     })
+    watch(() => props.setTransferType, (val: UITransferTypes) => {
+      localState.transferType = val
+    })
```
cameron-eyds: You won't need this by using the prop directly

ppr-ui/src/composables/fees/FeeSummary.vue
owillborn: Gotcha, had it as is just to be consistent. Will change

cameron-eyds (commented):
Code looks good! Just need to refactor the prop, what is happening is artificial reactivity, just use the prop directly and let vue take care of the rest :)

ppr-ui/src/composables/fees/FeeSummary.vue
owillborn: Changed

cameron-eyds (approved):
Right on! Less is more, always 😁 
Gj 👍

doug-lovett (approved):
Looks good.

PR #1209, opened 2023-03-06
Business Lookup on PPR Debtor Search

cameron-eyds:
*Issue #:* /bcgov/entity#15472

*Description of changes:*
* Implements Business LookUp on for PPRs Business Debtor Search
* Minor Test updates 
* Minor Type Error fix

### DIFF
--- a/ppr-ui/src/components/search/BusinessSearchAutocomplete.vue
+++ b/ppr-ui/src/components/search/BusinessSearchAutocomplete.vue
@@ -1,6 +1,6 @@
 <template>
   <v-card
-    v-if="searchValue.length >= 3 && !searching"
+    v-if="searchValue && searchValue.length >= 3 && !searching"
     id="business-search-autocomplete"
     class="auto-complete-card"
     elevation="5"
--- a/ppr-ui/package-lock.json
+++ b/ppr-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ppr-ui",
-  "version": "0.4.27",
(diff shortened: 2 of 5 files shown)


ppr-ui/src/components/search/BusinessSearchAutocomplete.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <v-card
-    v-if="searchValue.length >= 3 && !searching"
+    v-if="searchValue && searchValue.length >= 3 && !searching"
```
owillborn, line 3: would it be possible to use something like `searchValue?.length` here instead?

cameron-eyds: That would be nice wouldn't it!
Optional Chaining is a TypeScript feature which is not supported in the Vue template itself.

There might be a way but i'm not sure if its related to our vue version or what. We've short circuited instead for some time, might be worth looking at if this is possible!

cameron-eyds: Was curious, a quick bit of research mentions support for this (typescript in the tempalte) in Vue 3 👍

owillborn: Neat! Ok thanks for the clarification

owillborn (approved):
LGTM

doug-lovett (approved):
Looks good to me.

PR #1215, opened 2023-03-09
15374 implement remove for DEAT flow

owillborn:
*Issue #:* /bcgov/entity#15374

*Description of changes:*
- Built DeathCertificate component to capture `deathCertificateNumber` and `deathDateTime` for deceased owner
- Field validations for  `deathCertificateNumber` and `deathDateTime`, more to be added in https://app.zenhub.com/workspaces/assets-team-space-61426544e7e2ea001d079597/issues/gh/bcgov/entity/15582
- Updates to homeOwnersTable

image


ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -119,23 +119,33 @@
...
+              <v-chip
+                v-if="isMhrTransfer && showDeathCertificate && isRemovedHomeOwner(row.item)"
+                class="badge-delete ml-8 mt-2"
+                label x-small
+                color="#grey lighten-2"
+                text-color="$gray9"
+                data-test-id="owner-deceased-badge"
+              >
+                <b>DECEASED</b>
+              </v-chip>
```
cameron-eyds, line 121: The only difference i see between the two badges is the chip content (ie Deleted vs Deceased) and the test-id. 
Perhaps we just template the content vs a whole new badge? 
ie `<b>{{ showDeathCertificate ? 'DECEASED' : 'DELETED' }}</b>`?

Side question though, have you looked at the spec, does it require a deleted or deceased action type to be passed in the payload or is it just assumed in this flow?

owillborn: oo good question, I'll check

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -241,6 +251,15 @@
...
         </tr>
+        <tr v-if="isRemovedHomeOwner(row.item) && showDeathCertificate" class="death-certificate">
+          <td :colspan="homeOwnersTableHeaders.length" class="py-0">
+            <v-expand-transition>
+              <DeathCertificate
+              :deceasedOwner="row.item"
+              />
+            </v-expand-transition>
+          </td>
+        </tr>
```
cameron-eyds: Looks good!
Will need an indent on your :prop

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -341,6 +362,12 @@ export default defineComponent({
         getValidation(MhrSectVal.REVIEW_CONFIRM_VALID, MhrCompVal.VALIDATE_STEPS)),
       showEditActions: computed((): boolean => !props.isReadonlyTable),
       homeOwnersTableHeaders: props.isReadonlyTable ? homeOwnersTableHeadersReview : homeOwnersTableHeaders,
+      transferType: computed(() => { return getMhrTransferType.value?.transferType }),
+      showDeathCertificate: computed(() => {
+        // Add other conditions for other transfer Types
+        return localState.transferType === ApiTransferTypes.SURVIVING_JOINT_TENANT ||
+               localState.transferType === ApiTransferTypes.TO_EXECUTOR_PROBATE_WILL
+      }),
```
cameron-eyds: Perfect type of condition/property that would fit into our useTransferOwners composable, since it's specific to a particular transfer owner flow.

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -499,6 +525,14 @@ export default defineComponent({
     background-color: $gray1 !important;
   }
 
+  .no-bottom-border {
+    border-bottom: none !important;
+  }
+
+  .death-certificate {
```
cameron-eyds: `death-certificate-row` might be a bit more accurate. 
The root component usually has an id that would be called the name of the component.

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
```diff
@@ -0,0 +1,180 @@
+<template>
+  <div class="mhr-death-certificate">
```
cameron-eyds: Might want to make it an id. 
Also, i'm not entirely sure we have to label it with mhr. 🤔

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
```diff
@@ -0,0 +1,180 @@
...
+              for="deathCertificate"
+              :class="{ 'error-text': validateDeathCertificate && hasError(deathCertificateRef) }"
+            >
+              Death Certificate Number
+            </label>
+          </v-col>
+          <v-col cols="9">
+            <v-text-field
+              id="death-certificate-number"
+              v-model="deathCertificate"
```
cameron-eyds: is this a number? or a document?

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
owillborn: "Deleted owners (deceased) or surviving owners must be a OWNER_BUS or OWNER_IND partyType."
This is under the TRAND section in the spec and makes me assume the API takes the DELETED action type. Didn't see anything more specific. @doug-lovett are you able to confirm?

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
```diff
@@ -0,0 +1,180 @@
...
+import { FormIF, MhrRegistrationHomeOwnerIF } from '@/interfaces' // eslint-disable-line no-unused-vars
+import { localTodayDate } from '@/utils'
+
+export default defineComponent({
+  name: 'DeathCertificate',
+  props: {
+    deceasedOwner: {
+      type: Object as () => MhrRegistrationHomeOwnerIF,
+      default: null
+    }
```
cameron-eyds, line 90: What information does this component require from the deceased HomeOwner?

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
owillborn: user would input a number referencing a document. So a number I think

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
owillborn: This component would set the `deathCertificateNumber` and `deathDateTime` for the deceased owner once valid

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
owillborn: Agreed

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
```diff
@@ -0,0 +1,180 @@
...
+    )
+
+    const localState = reactive({
+      validateDeathCertificate: false, // NEW VALIDATOR REQUIRED
+      isFormValid: false, // Death Certificate form without Death Date Picker
+      isDeathCertificateFormValid: computed((): boolean => localState.isFormValid && !!localState.dateOfDeath),
+      deathCertificate: null,
+      dateOfDeath: null,
+      hasCertificate: false, // Will be used for validation on UI side only (original certificate checkbox)
+      deceasedOwner: props.deceasedOwner,
```
cameron-eyds: Depending on your reasoning for the prop, i'm not entirely sure you need this.
If you are looking to update properties on a specific owner, it might be better suited to pass an id down and directly update that owner by id.

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
```diff
@@ -0,0 +1,180 @@
...
+    )
+
+    // Update deceased owner deathDateTime when value changes
+    watch(
+      () => localState.dateOfDeath,
+      (val: string) => {
+        localState.deceasedOwner.deathDateTime = val
+        setUnsavedChanges(true)
+      }
+    )
```
cameron-eyds: The watcher part works fine but the function is only mutating the components local copy of the prop you've passed down, if that makes sense. 
I think what you want to do is update the homeOwner (by id) not so much the local instance of the homeOwner. 
ie `watch( () => localState.dateOfDeath, (val: string) => { editSomeHomeOwner({ key: propertyName, val: someDate }) })`

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
doug-lovett: The api has no such action for the owner in a deleteOwnerGroup for this registration. If the death certificate information is present the owner is deceased. If the the death certificate information is not provided for an owner in the deleteOwnerGroup, that owner must be in the addOwnerGroup as a surviving owner.

doug-lovett: The OWNER_BUS, OWNER_IND partyType rule means the group cannot have an EXECUTOR/ADMINISTRATOR/TRUSTEE.

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
owillborn: Yes, was going to update the local values as so, then once proper validations are in place, set the owner values outside the local instance when confirmed valid. So pretty much waiting on validations if you get what I mean

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
cameron-eyds: Cool, sounds good! 
So its really just some interchangeable text 👍

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
cameron-eyds: right on, always a good idea to keep the ui property names in sync with the api, just saves mapping headaches (as much as possible 😅 )

cameron-eyds: Screen Shot 2023-03-09 at 3 47 30 PM

cameron-eyds: I suppose the same goes for the date property :)

owillborn: Makes a lot more sense, will change

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
cameron-eyds: For sure, i guess using whats existing, `editHomeOwner` it takes the owner and the groupId. 
That being said, i wouldn't copy the prop value to another local property, you will lose reactivity on the object. 

So when you go to set it, just spread it. ie `setSomeEditedHomeOwner({...props.homeOwner, deathCertNumber: someNewVal })` as an example. or create a new copy and then use that in the current edit functionality.

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
cameron-eyds: Copy that 👍 
You can go ahead and set to store without validations, typically is allowed, we can implement validation after the fact.
 
Few reasons, we validate everything anywho, users can put in whatever they want typically.
We have draft filings which have zero validation, user can save whatever they want, to some extent. 

I know it seems odd but its more of the less the reason for the validations. 
We can always go two ways with it, enforce specific valid entries or allow the user to enter/store whatever but provide validations and information when things are invalid.

cameron-eyds:
Looks good, lots of good stuff here! 👍

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
owillborn: Ohhh interesting! I'll remove the local instance then. I'm able to manipulate the prop then set it though right?

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
owillborn: Gotcha, on it

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -117,25 +117,25 @@
...
+          <td :class="{'no-bottom-border' : isRemovedHomeOwner(row.item) && showDeathCertificate}">
             <div :class="{'removed-owner': isRemovedHomeOwner(row.item)}">
               {{ toDisplayPhone(row.item.phoneNumber) }}
               <span v-if="row.item.phoneExtension"> Ext {{ row.item.phoneExtension }} </span>
             </div>
           </td>
-          <td v-if="showEditActions" class="row-actions text-right">
-
+          <td v-if="showEditActions" class="row-actions text-right"
+          :class="{'no-bottom-border' : isRemovedHomeOwner(row.item) && showDeathCertificate}">
```
cameron-eyds: Minor indentation here on the td props

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -341,6 +353,7 @@ export default defineComponent({
         getValidation(MhrSectVal.REVIEW_CONFIRM_VALID, MhrCompVal.VALIDATE_STEPS)),
       showEditActions: computed((): boolean => !props.isReadonlyTable),
       homeOwnersTableHeaders: props.isReadonlyTable ? homeOwnersTableHeadersReview : homeOwnersTableHeaders,
+      transferType: computed(() => { return getMhrTransferType.value?.transferType }),
```
cameron-eyds: I don't believe you need this local computed value. The getter itself is a global computed value.

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
```diff
@@ -0,0 +1,187 @@
+<template>
+  <div id="death-certificate" class="death-certificate">
+    <v-card flat class="pl-8 rounded" :class="{ 'border-error-left': showFormError }">
```
cameron-eyds: Another minor thing, we only need 1 root element in the template so i'm not sure the div is really required here. You could apply the class and id to the card itself and make it the root element. 
Unless you had a purpose for wrapping it in the div?

ppr-ui/src/components/mhrTransfers/DeathCertificate.vue
cameron-eyds: More or less. 
We never mutate the prop directly but spread is a means of shallow copying.

cameron-eyds (approved):
Nice work, just some minor stuff to tackle when your back before we merge 👍

doug-lovett (approved):
Scanned, rubber stamping after Cameron's great constructive comments.

PR #1239, opened 2023-03-22
Search report conditionally add exempt message.

doug-lovett:
*Issue #:* /bcgov/entity#15679

*Description of changes:*
- Conditionally add exempt status message to MH registration in search report.
- Also when matching owner names on TRAND (editing instead of adding) treat empty middle name and no middle name as equivalent.
- Also allow any party type for delete group on staff transfer due to death registrations. For example, allow TRUSTEE for TRANS_WILL
- Also add new TRANS_AFFIDAVIT rules: all deleted owners must have death cert info.

### DIFF
--- a/mhr_api/src/mhr_api/models/db2/utils.py
+++ b/mhr_api/src/mhr_api/models/db2/utils.py
@@ -694,7 +694,7 @@ def __build_summary(row, add_in_user_list: bool = True, mhr_list=None):
                     owner_names += name[1:] + '\n'
         # remove comma if exists at end of str
         if owner_names[-2] == ',':
-            owner_names = owner_names[:-2]      
+            owner_names = owner_names[:-2]
     summary = {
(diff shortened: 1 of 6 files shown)


cameron-eyds (approved):
Looks great!

mhr_api/src/mhr_api/models/db2/utils.py
```diff
@@ -694,7 +694,7 @@ def __build_summary(row, add_in_user_list: bool = True, mhr_list=None):
                     owner_names += name[1:] + '\n'
         # remove comma if exists at end of str
         if owner_names[-2] == ',':
-            owner_names = owner_names[:-2]      
+            owner_names = owner_names[:-2]
```
owillborn, line 697: Thanks for cleanup

owillborn (approved):
LGTM!

dimak1 (approved):

PR #1241, opened 2023-03-23
15432 Transaction History Bug Fix

owillborn:
*Issue #:* /bcgov/entity#15432

*Description of changes:*
 - Fix for request when submitting MHR Registration not having payment parameters
 - Text fix for 15417 (missing period)

### DIFF
--- a/ppr-ui/src/utils/mhr-api-helper.ts
+++ b/ppr-ui/src/utils/mhr-api-helper.ts
@@ -290,12 +290,12 @@ export async function searchMhrPDF (searchId: string): Promise<any> {
     })
 }
 
-export async function submitMhrRegistration (payloadData, queryParamData) {
+export async function submitMhrRegistration (payloadData, staffPayment) {
   try {
-    // assuming the queryParamData (staff payment) is always available because of validation
-    const queryParamString = new URLSearchParams(queryParamData).toString()
+    // assuming the staffPayment is always available because of validation
+    const paymentParams = mhrStaffPaymentParameters(staffPayment)
 
-    const result = await axios.post(`registrations?${queryParamString}`, payloadData, getDefaultConfig())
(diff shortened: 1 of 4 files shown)


ppr-ui/src/utils/mhr-api-helper.ts
```diff
@@ -290,12 +290,12 @@ export async function searchMhrPDF (searchId: string): Promise<any> {
     })
 }
 
-export async function submitMhrRegistration (payloadData, queryParamData) {
+export async function submitMhrRegistration (payloadData, staffPayment) {
```
dimak1, line 293: Is this the fix for 15432?

owillborn: Yes, including the changes to the parameter formatting below.

cameron-eyds (approved):
LG!

ppr-ui/src/utils/mhr-api-helper.ts
dimak1: So by changing the params during submission it fixed the history?

owillborn: The change on line 293 is just a var name change to make the code more readable. Refactored below to not use URLSearchParams as it was returning empty, plus we already have a function to parse staff payment parameters (mhrStaffPaymentParameters)

owillborn: Basically parameters were not being added to the URL for the registration POST call because URLSearchParams was returning nothing. Refactored to use mhrStaffPaymentParameters and the POST URL now contains payment info fixing the history

dimak1: Gotcha, thanks for clarifying.

dimak1 (approved):

doug-lovett (approved):
Very nice.

PR #1246, opened 2023-03-24
MHR add registration frozen state check.

doug-lovett:
*Issue #:* /bcgov/entity#15728

*Description of changes:*
- Conditionally set account registration summary statusType to FROZEN.
- Add a validation state check on registrations where the last registration was a TRANS_AFFIDAVIT. 

### DIFF
--- a/mhr_api/src/mhr_api/models/db2/utils.py
+++ b/mhr_api/src/mhr_api/models/db2/utils.py
@@ -139,7 +139,13 @@ SELECT mh.mhregnum, mh.mhstatus, d.regidate, TRIM(d.name), TRIM(d.olbcfoli), TRI
            AND og2.status IN ('3')) as owner_names,
        TRIM(d.affirmby),
        d.documtid as document_id,
-       d.docuregi as doc_reg_number
+       d.docuregi as doc_reg_number,
+       (SELECT d2.docutype
+          FROM document d2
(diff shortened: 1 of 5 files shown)


cameron-eyds (approved):
Looks good! 
Also just curious, would the same value apply to the `status` property which is part of the GET for MHR APIs `/registrations`? 

Screen Shot 2023-03-24 at 3 20 32 PM

Good question, I thought about making this change but decided not to for now. FROZEN is more a temporary, informal state. The registration JSON returned by the GET is also used with the search results, and I don't think we want this state showing up there. We could of course modify the search version but I don't think it's needed. If a registration is frozen for a different reason and a longer period of time we might make this change.

cameron-eyds:
> Looks good! Also just curious, would the same value apply to the `status` property which is part of the GET for MHR APIs `/registrations`?
> 
> Screen Shot 2023-03-24 at 3 20 32 PM
> 
> Good question, I thought about making this change but decided not to for now. FROZEN is more a temporary, informal state. The registration JSON returned by the GET is also used with the search results, and I don't think we want this state showing up there. We could of course modify the search version but I don't think it's needed. If a registration is frozen for a different reason and a longer period of time we might make this change.

Sounds good sir, thank you for the info!

dimak1 (approved):

owillborn (approved):

PR #1247, opened 2023-03-25
Correctly delete a Home Owner in Registration flow

dimak1:
*Issue #:* 
- bcgov/entity#15757

*Description of changes:*
- Fix for correctly deleting a Home Owners from the table in the Registration flow

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
@@ -461,10 +461,12 @@ export default defineComponent({
     const {
       getMhrRegistrationHomeOwnerGroups,
       getMhrTransferHomeOwnerGroups,
+      getMhrTransferHomeOwners,
       getMhrRegistrationValidationModel
     } = useGetters<any>([
       'getMhrRegistrationHomeOwnerGroups',
       'getMhrTransferHomeOwnerGroups',
+      'getMhrTransferHomeOwners',
       'getMhrRegistrationValidationModel'
     ])
     const {
@@ -495,8 +497,7 @@ export default defineComponent({
       editHomeOwner,
       showGroups,
       setShowGroups,
-      setGroupFractionalInterest,
-      getTransferOrRegistrationHomeOwners
+      setGroupFractionalInterest
     } = useHomeOwners(props.isMhrTransfer)
 
     const {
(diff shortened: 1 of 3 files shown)


ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
```diff
@@ -513,7 +513,9 @@ export default defineComponent({
 
     const defaultHomeOwner: MhrRegistrationHomeOwnerIF = {
       ...props.editHomeOwner,
-      ownerId: props.editHomeOwner?.ownerId || getTransferOrRegistrationHomeOwners().length + 1 || (DEFAULT_OWNER_ID++),
+      ownerId: props.editHomeOwner?.ownerId || props.isMhrTransfer
```
dimak1: We need to track the correct `ownerId` for Transfers vs Registration. Otherwise, we could have owners with the same Id (if deleting the first owner and then adding a new one).

cameron-eyds (approved):

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
cameron-eyds: Although i said to just remove the parenthesis on the Default on Friday, couldn't the Default be removed entirely? 
ie no ternary and just fallback to the `getTransferOrRegistrationHomeOwners().length `? Wouldn't it resolve the same?

cameron-eyds:
Just a few tests to fix but otherwise looks fine. 
Would confirm the ternary + default is actually required though.

ppr-ui/src/components/mhrRegistration/HomeOwners/AddEditHomeOwner.vue
dimak1: I did remove the parenthesis, and it worked for some scenarios, but there are conditions that didn't help. We can't fallback to `getTransferOrRegistrationHomeOwners().length` because in registration, if I add 2 owners and then remove first one and then add the third one - both owners will have `ownerId` of 2. Then removing one will remove both.

cameron-eyds: Ah yes, makes sense on the fallback 👍  Thanks for clarifying. 

What scenarios requires the ternary?

dimak1: Basically what I discovered, is that for new registrations, we need `ownerId` to be incremental starting at 1. But for transfers, we need `ownerId` to be incremental from the number of owners in the `getTransferOrRegistrationHomeOwners()` because there could be some existing ones.

cameron-eyds: You bet, always existing ones in Transfers! 

Fair enough, that being said, would make sense to just get `getMhrTransferHomeOwners` vs `getTransferOrRegistrationHomeOwners()`

doug-lovett (approved):

PR #1252, opened 2023-03-27
V3 Migration build Prep

cameron-eyds:
*Issue #:* /bcgov/entity#15560

*Description of changes:*
* Major reformatting of all Class Based Components to Composition API. 
* Minor test udates. 

*NOTE:*
- Whats left is a few Shared components (Staff Payment, DatePicker) etc, that will need to be localized in order to remove the package dependencies completely, in conjunction with Mixins.


ppr-ui/src/components/parties/party/SecuredParties.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <v-container fluid no-gutters id="secured-parties-component" class="pa-0">
-    <change-secured-party-dialog
+    <ChangeSecuredPartyDialog
```
dimak1, line 3: Nice to (finally) see full names without dashes

dimak1 (approved):
Great refactor!

ppr-ui/src/components/parties/party/SecuredParties.vue
cameron-eyds: Yes! Will try and catch these as we go :) 
I haven't looked at the templates too much yet but upgrading Vuetify we will definitely be in there 💥

doug-lovett (approved):
Scanned the files with large number of differences. Looks great, a lot of work.

ppr-ui/src/components/parties/party/SecuredParties.vue
dimak1: Awesome, thanks!

PR #1260, opened 2023-04-03
14866 updates to ownership for executor scenarios

owillborn:
*Issue #:* /bcgov/entity#14866

*Description of changes:*
- Updates to home owner tenancy type and group tenancy type in executor/trustee/admin scenarios
- Highlighted scenarios here: https://projects.invisionapp.com/share/2C133LCZGDW9#/screens/470980264_Step3_-_UPDATED_TENANCY

### DIFF
--- a/ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
+++ b/ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
@@ -9,7 +9,7 @@ import '@/utils/use-composition-api'
 
 import { readonly, ref, toRefs, watch } from '@vue/composition-api'
 import { useActions, useGetters } from 'vuex-composition-helpers'
-import { ActionTypes, HomeTenancyTypes } from '@/enums'
+import { ActionTypes, HomeTenancyTypes, HomeOwnerPartyTypes } from '@/enums'
 import { MhrCompVal, MhrSectVal } from '@/composables/mhrRegistration/enums'
 import { useMhrValidations } from '@/composables'
 import { find, findIndex, remove, set } from 'lodash'
@@ -48,6 +48,12 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
     'setMhrTransferHomeOwnerGroups'
   ])
 
+  const executorTrusteeAdmin = [
+    HomeOwnerPartyTypes.EXECUTOR,
+    HomeOwnerPartyTypes.TRUSTEE,
+    HomeOwnerPartyTypes.ADMINISTRATOR
+  ]
+
   // Get Transfer or Registration Home Owners
   const getTransferOrRegistrationHomeOwners = (): MhrRegistrationHomeOwnerIF[] =>
     isMhrTransfer ? getMhrTransferHomeOwners.value : getMhrRegistrationHomeOwners.value
@@ -76,6 +82,12 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
   const getHomeTenancyType = (): HomeTenancyTypes => {
     // Groups
     const groups = getTransferOrRegistrationHomeOwnerGroups().filter(owner => owner.action !== ActionTypes.REMOVED)
+    // Variable to track if owners has a valid combination of Executor/Trustee/Admin (ETA) Owners
+    let hasETAError = false
+    groups.forEach(group => {
+      if (hasETAGroupError(group)) hasETAError = true
+    })
+
     const commonCondition = isMhrTransfer ? groups.length > 1 : showGroups.value
 
     // Special case where a defined Group is orphaned using remove functionality, we want to preserve the Group Type.
@@ -93,7 +105,7 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
       // Added second condition, because when an owner exists as a Sole Ownership, editing and clicking Done,
       // will change status to Tenants in Common unless above logic is in place..
       return HomeTenancyTypes.SOLE
-    } else if (numOfOwners > 1) {
+    } else if (numOfOwners > 1 && !hasETAError) {
       // More than one owner without groups showing
       return HomeTenancyTypes.JOINT
     }
@@ -103,7 +115,7 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
   const getGroupTenancyType = (group: MhrRegistrationHomeOwnerGroupIF): HomeTenancyTypes => {
     const numOfOwnersInGroup = group.owners.filter(owner => owner.action !== ActionTypes.REMOVED).length
 
-    if (numOfOwnersInGroup > 1) {
+    if (numOfOwnersInGroup > 1 && !hasETAGroupError(group)) {
       return HomeTenancyTypes.JOINT
     } else if (getHomeTenancyType() === HomeTenancyTypes.SOLE) {
       return HomeTenancyTypes.SOLE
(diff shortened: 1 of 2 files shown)


ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -112,6 +124,22 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
     }
   }
 
+  const hasETAGroupError = (group: MhrRegistrationHomeOwnerGroupIF): boolean => {
+    // Variable to track if group has a valid combination of Executor/Trustee/Admin (ETA) Owners
+    let hasETAGroupError = false
+    // If a group contains multiple ETA's within one group, set invalid
+    if (group.owners.filter(owner => executorTrusteeAdmin.includes(owner.partyType)).length > 1) {
+      console.log('Multiple ETA Error')
+      hasETAGroupError = true
```
dimak1: Here we could just `return true` and skip the second condition check below.

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -112,6 +124,22 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
     }
   }
 
+  const hasETAGroupError = (group: MhrRegistrationHomeOwnerGroupIF): boolean => {
+    // Variable to track if group has a valid combination of Executor/Trustee/Admin (ETA) Owners
+    let hasETAGroupError = false
```
dimak1: We don't really need this variable, as we can simply return the boolean.

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -76,6 +82,12 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
   const getHomeTenancyType = (): HomeTenancyTypes => {
     // Groups
     const groups = getTransferOrRegistrationHomeOwnerGroups().filter(owner => owner.action !== ActionTypes.REMOVED)
+    // Variable to track if owners has a valid combination of Executor/Trustee/Admin (ETA) Owners
+    let hasETAError = false
+    groups.forEach(group => {
+      if (hasETAGroupError(group)) hasETAError = true
```
dimak1: Maybe once we get the ETA error we don't have to continue iterating thru groups?

owillborn: refactored to use arr.some() instead

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
owillborn: fixed

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
owillborn: fixed

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -48,6 +48,12 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
     'setMhrTransferHomeOwnerGroups'
   ])
 
+  const executorTrusteeAdmin = [
+    HomeOwnerPartyTypes.EXECUTOR,
+    HomeOwnerPartyTypes.TRUSTEE,
+    HomeOwnerPartyTypes.ADMINISTRATOR
+  ]
```
cameron-eyds: Unless using this elsewhere, i would define within the scope of the function expression it's used.

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -93,7 +103,7 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
       // Added second condition, because when an owner exists as a Sole Ownership, editing and clicking Done,
       // will change status to Tenants in Common unless above logic is in place..
       return HomeTenancyTypes.SOLE
-    } else if (numOfOwners > 1) {
+    } else if (numOfOwners > 1 && !hasETAError) {
```
cameron-eyds: Just curious how this validation rule ties into determining the Home Tenancy Type. 
By the logic, it implies that if there is ever ever an `ETAError` that we always fall back to a `NA` type? 
This was a requirement?

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -103,7 +113,7 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
   const getGroupTenancyType = (group: MhrRegistrationHomeOwnerGroupIF): HomeTenancyTypes => {
     const numOfOwnersInGroup = group.owners.filter(owner => owner.action !== ActionTypes.REMOVED).length
 
-    if (numOfOwnersInGroup > 1) {
+    if (numOfOwnersInGroup > 1 && !hasETAGroupError(group)) {
```
cameron-eyds: Same as above, designs call for Tenancy Type of NA in any scenario where there is a E T or A?

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -112,6 +122,17 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
...
+    // If a group contains multiple ETA's within one group, return invalid
+    if (group.owners.filter(owner => executorTrusteeAdmin.includes(owner.partyType)).length > 1) {
+      return true
+    } else
+    // If group has a mix of ETA and living owner, return invalid
+    if (group.owners.some(owner => executorTrusteeAdmin.includes(owner.partyType)) &&
+        group.owners.some(owner => !executorTrusteeAdmin.includes(owner.partyType))) {
+      return true
+    } else return false
+  }
```
cameron-eyds: I believe you can simplify this to ..
```
return (group.owners.filter(owner => executorTrusteeAdmin.includes(owner.partyType)).length > 1) ||
(group.owners.some(owner => executorTrusteeAdmin.includes(owner.partyType)) &&
        group.owners.some(owner => !executorTrusteeAdmin.includes(owner.partyType)))(group.owners.some(owner => executorTrusteeAdmin.includes(owner.partyType)) &&
        group.owners.some(owner => !executorTrusteeAdmin.includes(owner.partyType)))
```
Though doesn't the second condition always resolve as false? They look mutually exclusive no, as in if one is true the other is false and vice versa?

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
owillborn: The only case where there can be an ETA and not have tenancy type of NA would be Sole Ownership. So I think I could remove the hasETAError and replace with hasETA, checking if an ETA is just present. Thanks for calling this out!

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
cameron-eyds: What were the findings here?

cameron-eyds (approved):
Looks good! Thank you for the updates

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
owillborn: They are mutually exclusive as you suggested. Scrapped that function and replaced in most recent commit

cameron-eyds: oh right on, yes looks good, nice work 👍

dimak1 (approved):

PR #1263, opened 2023-04-04
Navigation + EsLint Updates

cameron-eyds:
*Issue #:* /bcgov/entity#15583

*Description of changes:*
* Updated breadcrumb + navigation to utilize vue router vs url on back or breadcrumb navigation
* Implements eslint vuetify plugin + misc eslint fixes
* Minor adjustments to Dashboard Tabs to prevent reloading of PPR registrations on Tab change + minor optimizations.
* Many many many.... ESLINT fixes to adhere to recent Vue and Vuetify lint rules.


ppr-ui/.eslintrc.js
```diff
@@ -7,8 +7,13 @@ module.exports = {
     'plugin:vue/essential',
     '@vue/standard',
     '@vue/typescript'
+    // ToDo: enable and fix Lint issues when capacity.
+    // 'plugin:vue/base',
+    // 'plugin:vuetify/base',
+    // 'plugin:vuetify/recommended'
```
cameron-eyds: With the plugin available, we can enable these and start to implement vuetify fixes as we go. 
This will help alleviate the deprecation fixes required when we upgrade vuetify.

ppr-ui/src/components/tables/mhr/MhrRegistrationTable.vue
```diff
@@ -1,587 +0,0 @@
-<template>
```
cameron-eyds, line 1: This was an obsolete component. 
We opted to utilize the same data table for both PPR and MHR registrations.

ppr-ui/.eslintrc.js
```diff
@@ -7,8 +7,13 @@ module.exports = {
...
   ],
   rules: {
+    'vue/multi-word-component-names': 'off',
```
cameron-eyds, line 15: Feels like a weird rule to me, so i turned it off. Anyone else?
ie `Folio` wouldn't be accepted but `FolioNumber` or `Folio-Number` would be. 
I'm not sure the basis but there is probably a reason, that being said though, didn't want to rename all the single name components in the application :P

cameron-eyds:
~Working through the lint fixes introduced via the updated versions.~ Completed

ppr-ui/.eslintrc.js
cameron-eyds: Enabled these, after fixing issues 👍

owillborn (approved):
lgtm

ppr-ui/src/components/collateral/Collateral.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <v-container fluid no-gutters
+  <v-container fluid
     id="collateral-component"
-    class="white pa-0 rounded-bottom"
+    class="white pa-0 rounded-bottom no-gutters"
```
dimak1, line 4: So some of the component params are now css classes? I see it throughout the PR.

ppr-ui/src/components/mhrTransfers/SupportingDocuments.vue
```diff
@@ -48,22 +48,21 @@ import { ApiTransferTypes, SupportingDocumentsOptions } from '@/enums/transferTy
...
-import { DeathCertificate } from '.'
 import { transferSupportingDocuments } from '@/resources/'
 
 export default defineComponent({
   name: 'SupportingDocument',
   props: {
     deletedOwner: {
       type: Object as () => MhrRegistrationHomeOwnerIF,
-      default: null
+      required: true
```
dimak1, line 58: Not sure how the null value made it there 😄

ppr-ui/src/components/mhrTransfers/SupportingDocuments.vue
```diff
@@ -48,22 +48,21 @@ import { ApiTransferTypes, SupportingDocumentsOptions } from '@/enums/transferTy
...
     },
     // Used to disable Death Cert when group has only one owner
     isSecondOptionDisabled: {
       type: Boolean
     }
   },
-  components: { DeathCertificate },
+  components: { },
```
dimak1, line 65: Do we need an empty components here?

dimak1 (approved):
Lots of changes but looks great!

ppr-ui/.eslintrc.js
doug-lovett: Disabling makes sense to me.

doug-lovett (approved):
C'est magnifique. Nice tidying.

ppr-ui/src/components/collateral/Collateral.vue
cameron-eyds: Great question. It's hard to describe in a concise way, but i'l try :)

In this example, `no-gutters` i believe has always been a style class, as with many of the other props. 
The means to apply it however was made easier by the props, ie `v-row no-gutters >...`

That being said, i think why the linter has caught some of these and applied them as classes, is where the prop isn't supported on that specific component. 
So in this instance, `v-container` doesn't support the `no-gutters` prop, so it was applied as a class.

Fundamentally `no-gutters` is very much a helper class, like most the margin and padding helpers. 
Does that sort of help describe why the linter has done this?

ppr-ui/src/components/mhrTransfers/SupportingDocuments.vue
cameron-eyds: All good! Linter caught it :)

ppr-ui/src/components/mhrTransfers/SupportingDocuments.vue
cameron-eyds: We definitely do not.
I'm guilty of leaving these in, in a few spots. I thought these component options make adding/editing a component more intuitive, like an empty `emits: []` as another example.

I'm 100% ok with removing though too, whatever you prefer!

PR #1265, opened 2023-04-05
Group numbering bug fix

owillborn:
*Issue #:* /bcgov/entity#14866

*Description of changes:*
QA found grouping issues which resulted in incorrect tenancy types being displayed. Refactored how executors are added to groups in the WILL flow. Newly added owners will now always be added to the group from which the owner was deceased, as opposed to creating a new group (happened when all original owners were deleted/marked deceased).

### DIFF
--- a/ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
+++ b/ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
@@ -9,7 +9,7 @@ import '@/utils/use-composition-api'
 
 import { readonly, ref, toRefs, watch } from '@vue/composition-api'
 import { useActions, useGetters } from 'vuex-composition-helpers'
-import { ActionTypes, HomeTenancyTypes, HomeOwnerPartyTypes } from '@/enums'
+import { ActionTypes, HomeTenancyTypes, HomeOwnerPartyTypes, ApiTransferTypes } from '@/enums'
 import { MhrCompVal, MhrSectVal } from '@/composables/mhrRegistration/enums'
 import { useMhrValidations } from '@/composables'
 import { find, findIndex, remove, set } from 'lodash'
@@ -30,14 +30,16 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
     getMhrRegistrationValidationModel,
     getMhrTransferHomeOwnerGroups,
     getMhrTransferHomeOwners,
-    getMhrTransferCurrentHomeOwnerGroups
+    getMhrTransferCurrentHomeOwnerGroups,
+    getMhrTransferType
   } = useGetters<any>([
     'getMhrRegistrationHomeOwners',
     'getMhrRegistrationHomeOwnerGroups',
     'getMhrRegistrationValidationModel',
     'getMhrTransferHomeOwnerGroups',
     'getMhrTransferHomeOwners',
-    'getMhrTransferCurrentHomeOwnerGroups'
+    'getMhrTransferCurrentHomeOwnerGroups',
+    'getMhrTransferType'
   ])
 
   const {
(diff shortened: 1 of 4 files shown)


ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -219,8 +220,15 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
       homeOwnerGroups = [...getMhrRegistrationHomeOwnerGroups.value]
     }
     // For Mhr Transfers with Removed Groups, assign a sequential groupId
-    const transferDefaultId = homeOwnerGroups.find(group => group.action !== ActionTypes.REMOVED)?.groupId ||
+    // If WILL flow, add new executor to existing group instead of incrementing the group
+    let transferDefaultId = null
```
dimak1: Setting `transferDefaultId = groupId` and removing the `else` clause will have the same effect.

dimak1 (changes requested):
All looks good but please add a unit test. You can reference existing test for WILL transfer in `MhrTransferHomeOwners.spec.ts`.

cameron-eyds:
Just a heads up, there was a very explicit requirement that a new owner added after all owners had been deceased was its own thing. Here is a pic of what i mean, hopefully this is preserved! 

Screen Shot 2023-04-05 at 4 56 53 PM

dimak1:
@cameron-eyds for WILL transfers, the requirement is that Executor is added to the group with one or all deleted owners. Also, there is an error message _All owners must be deceased and an executor added._ 


Screen Shot 2023-04-05 at 5 00 07 PM

cameron-eyds:
> @cameron-eyds for WILL transfers, the requirement is that Executor is added to the group with one or all deleted owners. Also, there is an error message _All owners must be deceased and an executor added._
> 
> Screen Shot 2023-04-05 at 5 00 07 PM

O see, so the logic is tied to the party type, sounds good!
Just wanted to highlight there was some other requirements in another scenario, similar but opposite :)

dimak1:
@cameron-eyds yes, looks like each transfer type has its own unique requirements.

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
```diff
@@ -229,7 +237,10 @@ export function useHomeOwners (isMhrTransfer: boolean = false) {
         (group: MhrRegistrationHomeOwnerGroupIF) => group.groupId === (groupId || fallBackId)
       ) || ({} as MhrRegistrationHomeOwnerGroupIF)
 
-    if (groupToUpdate.owners && groupToUpdate.action !== ActionTypes.REMOVED) {
+    // Allow update to "REMOVED" group if WILL flow
+    if (groupToUpdate.owners &&
+        (groupToUpdate.action !== ActionTypes.REMOVED ||
+         getMhrTransferType.value?.transferType === ApiTransferTypes.TO_EXECUTOR_PROBATE_WILL)) {
```
cameron-eyds, line 241: O wow, so you can remove the group as a whole and still add to it? 
Wouldn't the whole group in this state?

Screen Shot 2023-04-06 at 7 22 24 AM

cameron-eyds (commented):
Looks good, no other comments than Dima's!

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
dimak1: In Will Transfers, all owners in the group could be removed but we do not show 'Previous Owner Group' text as well as Group Undo button. To this flow is not like others.

doug-lovett (approved):
Looks good, no additional comments.

ppr-ui/src/composables/mhrRegistration/useHomeOwners.ts
cameron-eyds: A lot of layers!
I just feel bad for Divya trying to remember all these scenarios haha

dimak1: Yes, and implementing them is not fun either, a lot of conflicting validations.

dimak1 (approved):
@owillborn looks great with a test! Should make you feel more confident that your code works. Good job!

cameron-eyds (approved):

doug-lovett (approved):
Looks good.

dimak1:
I run into the same issue when working with WILL transfers. I will copy the code for now, but we should merge it as soon as we can.

PR #1277, opened 2023-04-17
Updates for Draft SoG/SJT Transfers

cameron-eyds:
*Issue #:* /bcgov/entity#15795

*Description of changes:*
* Updates made for Draft Transfers for both Sale and Gift and Transfer to surviving joint tenants (Other Transfer types still wip, can revisit those when complete)
* Fixed issue causing draft resume for tranfer
* Updates to accommodate Reg Type inclusion in API response (Would try to resume DRAFT Mhr vs Transfer)
* Minor refactors.

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
@@ -667,7 +667,7 @@ export default defineComponent({
         (isTransferToSurvivingJointTenant.value ? localState.isValidDeathCertificate : true) &&
         (isTransferToExecutorProbateWill.value ? TransWill.isValidTransfer.value : true)
       )
-    }, { deep: true })
+    }, { immediate: true, deep: true })
 
     watch(
       () => enableTransferOwnerGroupActions(),
--- a/ppr-ui/src/views/mhrInformation/MhrInformation.vue
+++ b/ppr-ui/src/views/mhrInformation/MhrInformation.vue
@@ -274,7 +274,7 @@
               </template>
             </section>
           </v-col>
-          <v-col class="pl-6 pt-5" cols="3" v-if="showMhrFeeSummary || isReviewMode">
+          <v-col class="pl-6 pt-5" cols="3" v-if="showTransferType || isReviewMode">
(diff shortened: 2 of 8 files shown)


ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -667,7 +667,7 @@ export default defineComponent({
         (isTransferToSurvivingJointTenant.value ? localState.isValidDeathCertificate : true) &&
         (isTransferToExecutorProbateWill.value ? TransWill.isValidTransfer.value : true)
       )
-    }, { deep: true })
+    }, { immediate: true, deep: true })
```
cameron-eyds, line 670: Will do an initial validation
( To support Draft data being resumed if the user makes no changes after resuming )

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -637,8 +643,6 @@ export default defineComponent({
 
         if (!mhrTransferFiling.error) {
           setUnsavedChanges(false)
-          // Delete the draft on successful submission
-          if (getMhrInformation.value.draftNumber) await deleteMhrDraft(getMhrInformation.value.draftNumber)
```
cameron-eyds, line 641: The API requires the `draftNumber` in the payload in order to DELETE the draft. 
We previously didn't include said draftNumber, so this was done in error. 

Now that we are including the draftNumber and following the spec, we no longer require this, the api does all the magic 👍

owillborn (approved):
LGTM, lots of logic

ppr-ui/src/views/mhrInformation/MhrInformation.vue
doug-lovett: Thanks for catching this.

doug-lovett (approved):
Looks great.

ppr-ui/src/views/mhrInformation/MhrInformation.vue
cameron-eyds: No prob at all, it was my mistake, so easy to find 😆

PR #1325, opened 2023-05-08
Fixes for Transfer of Surviving Tenants

dimak1:
*Issue #:* 
- bcgov/entity#16197

*Description of changes:*
- Issue 1 - For SJT transfers, when navigating from the Review & Confirm page back to the first page, Death Certificate component would not re-validate properly, causing errors. 
- Issue 2 - submission fails due to bad payload structure, work in progress

### DIFF
--- a/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
+++ b/ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
@@ -282,7 +282,6 @@
               <DeathCertificate
                 :deceasedOwner="row.item"
                 :validate="validateTransfer"
-                @isValid="isValidDeathCertificate = $event"
               />
             </v-expand-transition>
           </td>
@@ -358,7 +357,6 @@
                   <DeathCertificate
                     :deceasedOwner="row.item"
                     :validate="validateTransfer"
-                    @isValid="isValidDeathCertificate = $event"
                     :isDisabled="isGlobalEditingMode"
                   />
                 </template>
@@ -464,6 +462,7 @@ export default defineComponent({
       TransSaleOrGift,
       TransToExec,
       TransToAdmin,
+      TransJointTenants,
       getMhrTransferType
     } = useTransferOwners(!props.isMhrTransfer)
 
@@ -492,7 +491,6 @@ export default defineComponent({
       ownerToDecease: null as MhrRegistrationHomeOwnerIF,
       isEditingMode: computed((): boolean => localState.currentlyEditingHomeOwnerId >= 0),
       isAddingMode: computed((): boolean => props.isAdding),
-      isValidDeathCertificate: false,
       showTableError: computed((): boolean => {
         // For certain Transfers, we only need to check for global changes and do not show table error in other cases
         if (isTransferToExecutorProbateWill.value ||
@@ -753,7 +751,7 @@ export default defineComponent({
         localState.isValidAllocation &&
         !localState.hasGroupsWithNoOwners &&
         (isTransferDueToSaleOrGift.value ? !TransSaleOrGift.hasMixedOwners.value : true) &&
-        (isTransferToSurvivingJointTenant.value ? localState.isValidDeathCertificate : true) &&
+        (isTransferToSurvivingJointTenant.value ? TransJointTenants.isValidTransfer.value : true) &&
         ((isTransferToExecutorProbateWill.value || isTransferToExecutorUnder25Will.value)
           ? TransToExec.isValidTransfer.value : true)
       )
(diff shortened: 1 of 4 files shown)


ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
```diff
@@ -298,7 +298,8 @@ export const useMhrInformation = () => {
             return owner.individualName ? { ...owner, individualName: normalizeObject(owner.individualName) } : owner
           }),
           // Determine group tenancy type
-          type: ownerGroup.owners.filter(owner => owner.action === ActionTypes.REMOVED).length > 1
+          type: (ownerGroup.owners.filter(owner => owner.action === ActionTypes.REMOVED).length > 1 ||
+            getMhrTransferType.value?.transferType === ApiTransferTypes.SURVIVING_JOINT_TENANT)
```
dimak1, line 302: After looking into the API and mhr_api code, it seems that for SJT transfers, we need to have JOINT type for the deleted owner to have a successful submission. @doug-lovett @cameron-eyds

ppr-ui/src/components/mhrRegistration/HomeOwners/HomeOwnersTable.vue
```diff
@@ -753,7 +751,7 @@ export default defineComponent({
         localState.isValidAllocation &&
         !localState.hasGroupsWithNoOwners &&
         (isTransferDueToSaleOrGift.value ? !TransSaleOrGift.hasMixedOwners.value : true) &&
-        (isTransferToSurvivingJointTenant.value ? localState.isValidDeathCertificate : true) &&
```
dimak1, line 756: We can't use `isValidDeathCertificate` prop here, as it is always false when going back from Review & Confirm page, and it won't recompute without making changes in DeathCertificate component.

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1325-hygzcxga.web.app

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
cameron-eyds: Sounds good although may be worth checking out the matrix: https://docs.google.com/spreadsheets/d/1crIN1kjmva_64540WmMbyTV1BXj6FjCc/edit#gid=1836270235

I believe* there was intention to assign it an NA type but my memory is a little fuzzy on it.

cameron-eyds (approved):
Looks good!

orelbn (approved):
Looks Good!

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
doug-lovett: For the delete owner group, the type should always match the tenancy type returned by the API from the GET call.

doug-lovett (approved):
Looks good.

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
dimak1: Hmm, ok, I will double-check that, thanks.

dimak1: @doug-lovett the GET call you are referring is the one before the POST call, right? In case of transfer submission, the GET request is `other-registrations/${registrationNum}`, but I do not see tenancy type in the response?

Screen Shot 2023-05-09 at 1 31 07 PM

That leads me to think that the GET request you are talking about is the one when we load the registration via `/mhr/api/v1/registrations/150411?current=true`, which has the tenancy type. In this case, in order to be able to file an SJT Transfer, we need to have at least two owners in the Group, which is always JOINT?

doug-lovett: Yes, the GET request I am referring to is the one where the UI fetches the current state of the MH. In this case it is 
GET /mhr/api/v1/registrations/150411?current=true. It returns all the active groups, and each group contains a tenancy type. For groupId=1, here is what is returned:
    {
      "groupId": 1,
      "interest": "UNDIVIDED",
      "interestDenominator": 6,
      "interestNumerator": 1,
      "owners": [
        {
          "address": {
            "city": "LANGLEY",
            "country": "CA",
            "postalCode": "V3A 6H4",
            "region": "BC",
            "street": "6665 238TH STREET"
          },
          "individualName": {
            "first": "MARY-ANNE",
            "last": "BICKNELL"
          },
          "ownerId": 1,
          "partyType": "OWNER_IND",
          "phoneNumber": "6044620279",
          "type": "JOINT"
        },
        {
          "address": {
            "city": "NORTH SAANICH",
            "country": "CA",
            "postalCode": "V8L 5V4",
            "region": "BC",
            "street": "123-1640 ELECTRA BLVD"
          },
          "individualName": {
            "first": "BRYAN",
            "last": "HOLMES"
          },
          "ownerId": 2,
          "partyType": "OWNER_IND",
          "phoneNumber": "6044620279",
          "type": "JOINT"
        }
      ],
      "status": "ACTIVE",
      "tenancySpecified": true,
      "type": "JOINT"
    },
Whenever deleting an owner group, the groupId and type must match was is in the database to pass data validation.
If the group contains 2 or more owners it can be JOINT or NA if it contains an EXECUTOR/TRUSTEE/ADMINISTRATOR.

PR #1330, opened 2023-05-10
Does not show exempt and historical children + Code clean up

orelbn:
*Issue #:* /bcgov/entity#15910

*Description of changes:*
- Adds condition to check for exempt and historical children statusTypes and return an empty string
- Utilizes a map/object in the `getStatusDescription` to avoid looping and simplify code
- changes the hasDraft condition inthe `addHistoryDraftsToMhr` function to false, has it causes "* Draft Changes" to display when filtering by draft statusType for drafts. 
- Adds and updates tests

### DIFF
--- a/ppr-ui/src/composables/useRegistration.ts
+++ b/ppr-ui/src/composables/useRegistration.ts
@@ -1,5 +1,5 @@
 import { reactive, toRefs } from '@vue/composition-api'
-import { MhStatusTypes, StatusTypes } from '@/resources'
+import { MhrAPIToUIStatusTypesMap, PprAPIToUIStatusTypesMap } from '@/resources'
 import {
   APIAmendmentTypes,
   APIRegistrationTypes,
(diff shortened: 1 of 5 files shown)


pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1330-lrs9k4mo.web.app

doug-lovett (approved):
Looks good to me.

ppr-ui/src/resources/statusTypes.ts
```diff
@@ -38,3 +37,17 @@ export const MhStatusTypes = [
...
+  [APIStatusTypes.EXPIRED]: UIStatusTypes.EXPIRED,
+  [APIStatusTypes.DISCHARGED]: UIStatusTypes.DISCHARGED
+}
+
+export const MhrAPIToUIStatusTypesMap = {
+  [MhApiStatusTypes.DRAFT]: MhUIStatusTypes.DRAFT,
+  [MhApiStatusTypes.ACTIVE]: MhUIStatusTypes.ACTIVE,
+  [MhApiStatusTypes.EXEMPT]: MhUIStatusTypes.EXEMPT,
+  [MhApiStatusTypes.HISTORICAL]: MhUIStatusTypes.HISTORICAL
+}
```
cameron-eyds, line 53: I like this approach! 
No action required but have a look if the above value/text objects are something we still* need.

cameron-eyds (approved):
Nice work, looks great!
Just had one comment/question.

ppr-ui/src/resources/statusTypes.ts
orelbn: I believe the above array makes it simple to use vue model in rendering the select, since it specifies the value and text properties for the select, that's why I did not get rid of it.

image

cameron-eyds: Excellent, as long as you gave it consideration, which you clearly have :)

ppr-ui/src/composables/useRegistration.ts
```diff
@@ -42,21 +42,12 @@ export const useRegistration = (setSort: RegistrationSortIF) => {
     }
   }
 
-  const getStatusDescription = (status: APIStatusTypes | MhApiStatusTypes): string => {
+  const getStatusDescription = (status: APIStatusTypes | MhApiStatusTypes,
+    isChild: boolean, isPpr: boolean): string => {
```
dimak1, line 46: `isPpr` used only in tests? If yes then we need a default value, otherwise, it will be undefined on line 50.

dimak1 (commented):
Just one comment.

ppr-ui/src/composables/useRegistration.ts
orelbn: isPpr is passed in to the table row component, and I forward it to getStatusDescription from the tableRow, however I can add a default value if you believe it will be better!

image

dimak1: Sounds good, we should be ok then, thanks for clarifying.

dimak1 (approved):

PR #1344, opened 2023-05-18
Fix MHR Submission

dimak1:
*Issue #:* 
- bcgov/entity#16450

*Description of changes:*

- Fix to add `description` field as suffix for Owners when submitting

### DIFF
--- a/ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
+++ b/ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
@@ -213,9 +213,17 @@ export const useNewMhrRegistration = () => {
   }
 
   const parseOwnerGroups = (): MhrRegistrationHomeOwnerGroupIF[] => {
-    const ownerGroups: MhrRegistrationHomeOwnerGroupIF[] =
-      Object.values(cleanEmpty(getMhrRegistrationHomeOwnerGroups.value))
-    ownerGroups.forEach(ownerGroup => {
+    const ownersGroups: MhrRegistrationHomeOwnerGroupIF[] =
+      getMhrRegistrationHomeOwnerGroups.value.map(group => {
+        group.owners.forEach(owner => {
+          owner.description = owner.suffix
+        })
+
+        return group
+      })
+
+    const parsedOwnerGroups = Object.values(cleanEmpty(ownersGroups))
+    parsedOwnerGroups.forEach((ownerGroup: MhrRegistrationHomeOwnerGroupIF) => {
       ownerGroup.owners = Object.values(ownerGroup.owners)
 
       // @ts-ignore - TODO: Mhr-Submission - api asks for number, maybe fix this once step 3 is finished?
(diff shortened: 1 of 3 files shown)


dimak1:
@cameron-eyds maybe would make sense to drop the `suffix` and just keep `description` so the payload is cleaner and no extra code to copy `suffix` over

doug-lovett (approved):

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1344-uncbc3oe.web.app

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -213,9 +213,17 @@ export const useNewMhrRegistration = () => {
...
+      getMhrRegistrationHomeOwnerGroups.value.map(group => {
+        group.owners.forEach(owner => {
+          owner.description = owner.suffix
+        })
+
+        return group
+      })
+
+    const parsedOwnerGroups = Object.values(cleanEmpty(ownersGroups))
+    parsedOwnerGroups.forEach((ownerGroup: MhrRegistrationHomeOwnerGroupIF) => {
```
cameron-eyds, line 226: I thought Suffix field was still required for regular owner partyTypes, this seems to just map all instances of it to description. 
If we truly don't need the property suffix, than we could just change it in the AddEditOwner component no? 
My hunch was that suffix still applied for owner_ind and _bus,
If not, we should change the property we save too originally maybe.

Is suffix still a property we want to store the additional name information in for regular owners or is it obsolete with the addition of description? This might be a question for @tlebedovich  and/or @doug-lovett

orelbn (approved):

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
dimak1: Yea, for now I will leave it in, and will clarify with the rest of the team.

cameron-eyds: I guess thats on me not requesting changes immediately but in the future please don't merge with unresolved comments.

dimak1: Sorry, I thought it was a comment to look into the suffix at a later time (with the help of Tracey and Doug). What's the change you wanted me to make?

dimak1: As to the merge, I know you were going away till Tue, so I wanted to merge it so we can continue with the other tickets that depend on this fix.

cameron-eyds: All good!

The change being, do we need suffix for owners? If not, we just change the property to `description` in AddEditName etc where its captured. 
If we do, we change to description for the specific partyTypes that require it (exec, admin, trustee). 

As it stands now, we are passing the value from add name field in both suffix and description properties for all partyTypes. 
I'm not sure if this implications, such as in the outputs etc but its worth looking into.
While it may work, its just a bit of a decoy pattern as in some cases the properties may be obsolete depending on partyType.

PR #1356, opened 2023-05-26
Fix scroll to first element on filter

orelbn:
*Issue #:* /bcgov/entity#16455

*Description of changes:*
- Updates function to correctly identify first item in MHR

*Screenshots:*

Before when you filter by active (scrolls down):

image

After when you filter by active (stays at top of table or scrolls to top of table):
image

*Note for reviewers:*
- There is still some scroll issues, where the scrollIntoView function does not work when the browser is it a specific view on the table, I tried debugging it but it seems to be fairly complex and very low occurrence bug. But whenever the scrollIntoView function does trigger it now scrolls to the right item.

### DIFF
--- a/ppr-ui/src/components/tables/RegistrationTable.vue
+++ b/ppr-ui/src/components/tables/RegistrationTable.vue
@@ -513,13 +513,18 @@ export default defineComponent({
       const draftItem = item as DraftResultIF
       const firstBaseReg = localState.registrationHistory[0].baseRegistrationNumber
       const firstDocId = localState.registrationHistory[0].documentId
-      if (item.baseRegistrationNumber && item.baseRegistrationNumber === firstBaseReg) {
-        return true
-      }
-      if (!item.baseRegistrationNumber && draftItem.documentId === firstDocId) {
-        return true
-      }
-      return false
+      return item.baseRegistrationNumber
+        ? item.baseRegistrationNumber === firstBaseReg
+        : draftItem.documentId === firstDocId
+    }
+
+    const isFirstItemMhr = (item: MhRegistrationSummaryIF | MhrDraftIF): boolean => {
+      const draftItem = item as MhrDraftIF
+      const firstMhrNumber = localState.registrationHistory[0].mhrNumber
+      const firstDocId = localState.registrationHistory[0].draftNumber
+      return item.mhrNumber
+        ? item.mhrNumber === firstMhrNumber
+        : draftItem.draftNumber === firstDocId
     }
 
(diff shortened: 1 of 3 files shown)


pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1356-1ie5fhm6.web.app

doug-lovett (approved):
Looks good, thanks for the screen shots.

ppr-ui/src/components/tables/RegistrationTable.vue
```diff
@@ -513,13 +513,18 @@ export default defineComponent({
...
-        return true
-      }
-      return false
+      return item.baseRegistrationNumber
+        ? item.baseRegistrationNumber === firstBaseReg
+        : draftItem.documentId === firstDocId
+    }
+
+    const isFirstItemMhr = (item: MhRegistrationSummaryIF | MhrDraftIF): boolean => {
+      const draftItem = item as MhrDraftIF
```
dimak1: We probably do not need to define `draftItem` and simply use `item` instead (in the condition below). Unless typescript is requiring.

ppr-ui/src/components/tables/RegistrationTable.vue
```diff
@@ -513,13 +513,18 @@ export default defineComponent({
...
-      if (!item.baseRegistrationNumber && draftItem.documentId === firstDocId) {
-        return true
-      }
-      return false
+      return item.baseRegistrationNumber
+        ? item.baseRegistrationNumber === firstBaseReg
+        : draftItem.documentId === firstDocId
+    }
+
+    const isFirstItemMhr = (item: MhRegistrationSummaryIF | MhrDraftIF): boolean => {
```
dimak1, line 520: When calling `isFirstItemMhr(item as MhRegistrationSummaryIF)` you pass item only as a `MhRegistrationSummaryIF` type, so `MhrDraftIF` is not needed here.

dimak1 (approved):
Looks good, just two small comments.

ppr-ui/src/components/tables/RegistrationTable.vue
orelbn: good point, I updated the function call to reference `item as MhRegistrationSummaryIF | MhrDraftIf` to be more type accurate. I also removed draftItem assignment, as its not necessary.

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1356-1ie5fhm6.web.app

cameron-eyds (approved):

PR #1370, opened 2023-06-05
Fix Transfers validation for Execs and Admins

dimak1:
*Issue #:* /
- bcgov/entity#16290

*Description of changes:*
- Fix validations when adding and removing Executors and Admins in Transfers
- Add unit tests

### DIFF
--- a/ppr-ui/src/composables/mhrInformation/useTransferOwners.ts
+++ b/ppr-ui/src/composables/mhrInformation/useTransferOwners.ts
@@ -16,7 +16,6 @@ import {
 import { computed, reactive, toRefs } from 'vue-demi'
 import { isEqual, find, uniq } from 'lodash'
 import { normalizeObject } from '@/utils'
-import { useHomeOwners } from '@/composables'
 import {
   transferOwnerPrefillAdditionalName,
   transferOwnerPartyTypes,
@@ -44,10 +43,6 @@ export const useTransferOwners = (enableAllActions: boolean = false) => {
     getMhrTransferCurrentHomeOwnerGroups,
     getMhrTransferAffidavitCompleted
   } = storeToRefs(useStore())
-  const {
-    getGroupById,
-    getCurrentGroupById
-  } = useHomeOwners(true)
 
   /** Local State for custom computed properties. **/
   const localState = reactive({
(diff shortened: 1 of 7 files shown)


ppr-ui/src/composables/mhrInformation/useTransferOwners.ts
```diff
@@ -47,7 +47,7 @@ export const useTransferOwners = (enableAllActions: boolean = false) => {
   const {
     getGroupById,
     getCurrentGroupById
-  } = useHomeOwners(!!getMhrTransferType.value)
+  } = useHomeOwners(true)
```
dimak1: @cameron-eyds, you mentioned this could cause errors. We should discuss this.

cameron-eyds: It does indeed, will break MHRs for Tenants in common. 
To reproduce, create a basic tenants in common in the MHR and you'll notice with set to true, that the step will not get set to valid. 

Little bit of research done, this is a result of the validation watcher for home owners. Essentially the same issue you experienced but in reverse, it will look for the wrong data with regards to home owners. 

I think 2 potential solutions, remove the `useHomeOwners` from the `useTransferOwners` composable and just pass the `getGroupById` and `getCurrentGroupById` result in as an argument, to whatever functions use them in the transfer composable. Thats probably the best way to go with minimal work. 

The latter is a different dynamic property for this composable instance that doesn't clash with either flow but i'm not even sure if thats possible.

cameron-eyds (commented):
LGTM! Just the one comment, although it does require some action.

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1370-gm7jffyz.web.app

ppr-ui/src/composables/mhrInformation/useTransferOwners.ts
dimak1: Thanks for the breakdown. I agree, we should remove the useHomeOwners from the transfers composable. I will ask Patrick to create a separate ticket for it. Meanwhile, could you please block this PR from merging by 'Requesting Change', thanks!

cameron-eyds: I did this locally in about 10 minutes, do you want me to commit this? 
Wouldn't it make sense to include this here, so you could test again your current changes?

cameron-eyds: If you want, reference this code change: https://github.com/bcgov/ppr/compare/main...cameron-eyds:ppr:Composable-refactor

Or if there is a new ticket created, let me know and i can put this up for PR. or whatever works 👍

cameron-eyds (changes requested):
As per discussion, there is an issue with composables clashing with each other in some instances. 

Can either do here or we could pr: https://github.com/bcgov/ppr/compare/main...cameron-eyds:ppr:Composable-refactor if there is a new ticket.
That being said, unable to test against changes in pr but this does fix it for mhr's.

ppr-ui/src/composables/mhrInformation/useTransferOwners.ts
dimak1: Thanks for the reference, updated!

dimak1:
@cameron-eyds used your reference and updated code in my PR, run tests and all is green. Thanks!

ppr-ui/src/composables/mhrInformation/useTransferOwners.ts
cameron-eyds: Right on!

cameron-eyds (approved):
Looks good, thanks for the adjustment!

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1370-gm7jffyz.web.app

orelbn (approved):

orelbn (approved):

PR #1373, opened 2023-06-07
Manufacturer Registration Framework

orelbn:
*Issue #:* /bcgov/entity#16416

*Description of changes:*
- Updates state-model for manufacturer registration
- Updates tombstone, steps, and footer for manufacturer registration
- Makes MhrValidation to be more abstract to work with different state-models
- Adds state to store to identify manufacturer and manufacturer registration 
- Displays registration button to start manufacturer flow

*Screenshots:*

Adds launch button:
added-button

2-steps:
2-steps

Footers:

- Describe your home
bottom-footer

- Review and Confirm 
bottom-footer-review-confirm

Tombstone:

Tombstone

*Note for Reviewers:*

A Manufacturer registration can **NOT**  be completed at this point until the components are updated and validation state is set appropriately. The scope of this ticket is only establishing the framework.

We have a testing manufacturer account set up please message me if you need access.

Please confirm that only update needed for payload is making document ID optional.

Please let me know if I missed anything or if there are any issues.


ppr-ui/src/store/store.ts
```diff
@@ -70,7 +79,15 @@ export const useStore = defineStore('assetsStore', () => {
...
+    return (state.value.authorization?.authRoles.includes(AuthRoles.STAFF) || isRoleStaffSbc.value)
+  })
+  /** Whether the user has 'manufactuer' keycloak role. */
+  const isRoleManufactuer = computed((): boolean => {
+    return (state.value.authorization?.authRoles.includes(AuthRoles.MHR_REGISTER) &&
+    state.value.authorization?.authRoles.includes(AuthRoles.MHR_PAYMENT) &&
+    state.value.authorization?.authRoles.includes(AuthRoles.MHR_TRANSPORT) &&
+    state.value.authorization?.authRoles.includes(AuthRoles.MHR_TRANSFER_SALE) &&
+    !state.value.authorization?.authRoles.includes(AuthRoles.PPR_STAFF))
   })
```
orelbn: Identifies Manufacturer based on:

image

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
```diff
@@ -23,24 +28,20 @@ export const useMhrValidations = (validationState: any) => {
...
+      const components = Object.keys(validationState[section].value) as MhrCompVal[]
+      // Sets all components to false
+      for (const component of components) {
+        setValidation(section, component, false)
+      }
+    }
+    // Default for ADD_EDIT_OWNERS_VALID -> OWNERS_VALID is true
+    if (sections.includes(MhrSectVal.ADD_EDIT_OWNERS_VALID)) {
+      setValidation(MhrSectVal.ADD_EDIT_OWNERS_VALID, MhrCompVal.OWNERS_VALID, true)
+    }
```
orelbn: This resets all validations to their default values based on the model:

image

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
```diff
@@ -13,6 +16,8 @@ export const useMhrValidations = (validationState: any) => {
 
   /** Is true when all flags are true in specified section. */
   const getStepValidation = (section: MhrSectVal): boolean => {
+    // If section is not part of validation model return true
+    if (!validationState[section]) return true
     return Object.values(validationState[section].value).every(val => val)
```
orelbn: Makes it so if a component tries to get validation for a section that is not part of the current flow it automatically returns true.

For example: In the manufacture flow we populate the home owners section, so we don't check it is validation. If the HomeOwnersReview component tries to check for validation it will return true,

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1373-pglims1k.web.app

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
```diff
@@ -13,6 +16,8 @@ export const useMhrValidations = (validationState: any) => {
 
   /** Is true when all flags are true in specified section. */
   const getStepValidation = (section: MhrSectVal): boolean => {
+    // If section is not part of validation model return true
+    if (!validationState[section]) return true
```
dimak1: Wouldn't this be a false positive, if we return true for the specific section (that doesn't even exist in the val model)?

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
```diff
@@ -23,24 +28,20 @@ export const useMhrValidations = (validationState: any) => {
...
-    // Reset home owner validations
-    setValidation(MhrSectVal.HOME_OWNERS_VALID, MhrCompVal.OWNERS_VALID, false)
-    // Reset home location validations
-    setValidation(MhrSectVal.LOCATION_VALID, MhrCompVal.LOCATION_TYPE_VALID, false)
-    setValidation(MhrSectVal.LOCATION_VALID, MhrCompVal.CIVIC_ADDRESS_VALID, false)
-
-    // Reset Review validations
-    setValidation(MhrSectVal.REVIEW_CONFIRM_VALID, MhrCompVal.VALIDATE_STEPS, false)
+    // Gets all sections in model
+    const sections = Object.keys(validationState) as MhrSectVal[]
```
dimak1, line 29: Appreciate the explanation of the events. We can also simplify without extra declarations. Either way is fine.

Object.keys(validationState).forEach(section =>
  Object.keys(validationState[section].value).forEach(component =>
    setValidation(section, component, false)
  )
)

ppr-ui/src/store/store.ts
cameron-eyds: This is good! 
Originally i was concerned about overlap with Qualified Suppliers (they are identified with `mhr_transfer_sale` role.

That being said, that role does grant access to Transfers etc, so we may need to refine that permission. A discussion you could bring up with the team perhaps! ie are manufacturers allowed to perform transfers?

ppr-ui/src/components/registration/RegistrationBar.vue
```diff
@@ -4,7 +4,7 @@
       <v-col>
         <div class="actions">
           <v-btn
-            v-if="isMhr && (!isRoleQualifiedSupplier || isRoleStaff)"
+            v-if="isMhr && (!isRoleQualifiedSupplier || isRoleStaff || (isRoleManufactuer && hasMhrEnabled))"
```
cameron-eyds: Although the right thought, i don't believe you need the `hasMhrEnabled` flag here. 
This was originally created when the only part of mhr included in PPR was for MHR Search. There is a little note in the default flag section in the utils file. 

`mhr-registration-enabled` would be the flag for Registrations, that being said, this flag currently is in place and will block the entire MHR Table Tab for anyone but Staff.

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
```diff
@@ -3,7 +3,10 @@ import { MhrCompVal, MhrSectVal } from '@/composables/mhrRegistration/enums'
 export const useMhrValidations = (validationState: any) => {
   /** Set specified flag */
   const setValidation = (section: MhrSectVal, component: MhrCompVal, isValid: boolean): void => {
-    validationState[section].value[component] = isValid
+    // Only sets specified flag if the section and component are part of the validation model
+    if (validationState[section]?.value[component] !== undefined) {
+      validationState[section].value[component] = isValid
+    }
```
cameron-eyds, line 9: Great addition here, it was too forgiving previously and by allowing the creation of new properties, it was prone to error. Good check 👍

ppr-ui/src/interfaces/mhr-registration-interfaces/MhrRegistrationIF.ts
```diff
@@ -36,7 +36,7 @@ export interface MhrRegistrationIF {
 
 export interface NewMhrRegistrationApiIF {
   draftNumber?: string
-  documentId: string
+  documentId?: string
```
dimak1, line 24: Why do we need to make it optional?

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
```diff
@@ -275,14 +276,16 @@ export const useNewMhrRegistration = () => {
...
-      documentId: getMhrRegistrationDocumentId.value,
       ownLand: getMhrRegistrationOwnLand.value,
       submittingParty: parseSubmittingParty(),
       ownerGroups: parseOwnerGroups(),
       location: parseLocation(),
       description: parseDescription(),
       ...(isRoleStaffReg.value && !!getStaffPayment.value && {
         clientReferenceId: getStaffPayment.value.folioNumber
+      }),
+      ...(!isMhrManufactuerRegistration.value && {
```
dimak1: The `documentId` is not required when registering a Manufacturer? Is that right?

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
cameron-eyds: `addEditOwnersvalid` doesn't really follow the 'steps` model per se but i'm assuming this was pre-existing. 
A refactor for another day. 
Should* probably be part of the homeOwnersValid block.

ppr-ui/src/store/store.ts
```diff
@@ -180,6 +208,13 @@ export const useStore = defineStore('assetsStore', () => {
       useMhrValidations(modelRef).getStepValidation(MhrSectVal.HOME_OWNERS_VALID) &&
       useMhrValidations(modelRef).getStepValidation(MhrSectVal.LOCATION_VALID)
   })
+  const isMhrManufactuerRegistrationReviewValid = computed<boolean>(() => {
+    const modelRef = toRefs(getMhrRegistrationValidationModel.value)
+    return state.value.mhrValidationManufactuerState.reviewConfirmValid.authorizationValid &&
+      state.value.mhrValidationManufactuerState.reviewConfirmValid?.attentionValid &&
+      state.value.mhrValidationManufactuerState.reviewConfirmValid?.folioOrRefNumValid &&
+      useMhrValidations(modelRef).getStepValidation(MhrSectVal.YOUR_HOME_VALID)
+  })
```
cameron-eyds: Wouldn't the `getStepValidation` function serve the same purpose? 
This does appear to consider the filing as a whole as well as review.

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
orelbn: Agreed I thought the same!

cameron-eyds (commented):
Looks really good, great work!
Just a few comments.

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
orelbn: I was considering this, at first I added logic individual components to not run getStepValidation() for the manufacturer registration, for the steps that we are not validating, which was a bit of extra checks, but is more concrete.

For example I added in 4 different spots. `!isManufactuerRegistration.value &&  getStepValidation(...)`

Which I can foresee leading to a lot of checks in many places.

You are correct that It can lead to false positive if the model is not set up correctly or the wrong model is passed to the composable. However, I believe that proper testing and API validations might be able to capture that.

 I can look into adding logic to validation composable to verify the model passed to it is correct for the current flow.

However, if you prefer checking in individual components I can do that.

@cameron-eyds if you have any thoughts on this as well. 

Another option is to add all of the sections that we are auto populating to the model and setting those sections to true by default thus making it similar to the other model (i.e. has the same sections).

We could add a resource, where we store model-validations defaults, than in the resetAllValdation function we reset all the validations based on the models defaults. Which will make it easy to set defaults with some validations being true

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
orelbn: I am good with simplifying it thanks! 😄

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
orelbn: Based on the ticket yes.

image

ppr-ui/src/interfaces/mhr-registration-interfaces/MhrRegistrationIF.ts
orelbn: The ticket says no Document ID for Manufacturer.

I can make it optional or populate it with any empty string I guess.
Do you have any insights on which approach is better?

ppr-ui/src/store/store.ts
orelbn: Yeah I think it does, I mostly copied this section from what was already there. So was not sure if there is something I might be missing.

image

Update: the only differences between it and the stepValidation, is that getStepValidation(...REVIEW_CONFIRM) considers, `validateSteps` and `validateApp`. So we might need them to be different?

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
orelbn: Just for more context this one place where I was running into issues with it checking validations not on the model. 
`HomeOwnersReview`
image

In the manufacturer flow we will have the HomeOwnersReview component with data fetched from database, but we don't have the add home owners step where we would be setting the validation.

ppr-ui/src/components/registration/RegistrationBar.vue
orelbn: Sounds good ill change it to not include hasMhrEnabled

ppr-ui/src/store/store.ts
orelbn: Ill bring it up thanks.

orelbn: Manufacturers can perform mhr_transfer_sale

image

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
dimak1: Ok cool, and we are reusing the same IF for registering Manufacturers?

ppr-ui/src/store/store.ts
dimak1: @orelbn is this screenshot from the chat? If yes, we should post it somewhere for everyone to see. Any requirements or rules (that come up) should be added to Figma or tickets. Thanks!

ppr-ui/src/interfaces/mhr-registration-interfaces/MhrRegistrationIF.ts
dimak1: Since it's just one field, I'm not too opposed - we can make it optional here, instead of using a new (dedicated) interface for Manufacturers' Registrations. If we end up modifying too much, we'll create a new IF.

orelbn: Agreed, I had the same thought process!

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
dimak1: For separation of concerns and simplicity, can we just call a different step validation function (with its own validation model) when validating steps for Manf. Reg?

cameron-eyds: The composable is agnostic of the model, so it can be reused for any 'stepper' style filing we create. 
Functions for a specific type of validation model would sort of defeat the purpose of the reusable composable no? 

I think less is more and although reuse of anything can introduce some measure of complexity, it should always be the goal if we can.

ppr-ui/src/store/store.ts
cameron-eyds: Ah i see. Fair enough!
Yes, i'm sure it has it's purpose, although both validate steps and app would be true at the point of considering if the filing as a whole was good to go. 
It MAY work with the common getSteps function, worth a try but no action required 👍

ppr-ui/src/store/store.ts
cameron-eyds: Definitely agree, good item to highlight!

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
dimak1: Ok, in that case, can we pass a different validation model, that would have the right sections to validate? Have you tried that @orelbn?

ppr-ui/src/composables/mhrRegistration/useNewMhrRegistration.ts
orelbn: At the moment yes, we can change that if there is a lot of differences in the payload.

ppr-ui/src/composables/mhrRegistration/useMhrValidations.ts
orelbn: I am not clear about what right sections to validate means.

dimak1: For example, if we have only two Steps in the flow, we can create a validation model with only these two steps. That way, we do not need to reuse other validation model (that has more Steps) and return true for Steps that do not exist.

cameron-eyds (approved):
I know there is some unit test and slight refactors pending but assuming you will get those in the next commit 👍 

Keep in mind, there will be some minor conflicts to resolve.
Let me know if you wanted any help there.

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1373-pglims1k.web.app

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1373-pglims1k.web.app

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
```diff
@@ -215,14 +216,10 @@ export default defineComponent({
           scrollToInvalid(MhrSectVal.LOCATION_VALID, 'mhr-home-location')
           break
         case RouteNames.MHR_REVIEW_CONFIRM:
+          let stepsValidation = getSteps.value.filter((step : StepIF) => step.valid)
```
dimak1: Filtering only valid steps? Do we need extra conditions here?
Since it's an MHR Registration flow as well, we just need to be cautious to not break it.

ppr-ui/tests/unit/Breadcrumb.spec.ts
```diff
@@ -268,4 +269,78 @@ describe('Breadcrumb component tests', () => {
...
+    for (let i = 0; i < tombstoneBreadcrumbRegistration.length; i++) {
+      expect(breadcrumbs.at(i).text()).toContain(wrapper.vm.handleStaff(tombstoneBreadcrumbRegistration[i].text))
+    }
+
+    // Tear down
+    defaultFlagSet['mhr-ui-enabled'] = false
+    store.setUserProductSubscriptionsCodes([])
+  })
+
+  it('renders on Mhr Registration: Manufactuer - MHR and PPR', async () => {
```
dimak1: Typo here. Not a big deal, but if someone will be doing a global search for "Manufacturer" it won't come up.

ppr-ui/tests/unit/Breadcrumb.spec.ts
```diff
@@ -268,4 +269,78 @@ describe('Breadcrumb component tests', () => {
...
+    // Set up
+    await store.setAuthRoles(mockedManufactuerAuthRoles)
```
dimak1: Typo here as well.

doug-lovett (approved):
Looks tres magnifique. Cameron and Dima thoroughly reviewed so rubber stamping.

ppr-ui/tests/unit/ButtonFooter.spec.ts
```diff
@@ -378,3 +381,125 @@ describe('Button events', () => {
...
+
+  afterAll(async () => {
+    await store.setAuthRoles([])
+    await store.setRegistrationType(null)
+  })
+
+  beforeEach(async () => {
+    const localVue = createLocalVue()
+    localVue.use(VueRouter)
+    wrapper2 = shallowMount((YourHome as any), { localVue, store, router, vuetify })
```
dimak1: Do we need wrapper2?

dimak1 (approved):
Great stuff, thanks for all the updates and unit tests!

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
orelbn: Sorry that was a typo. It should be map step .valid, I will fix it. Thanks for catching that!

I checked that validation was still highlighting, but forget double check that scroll was working.

ppr-ui/tests/unit/Breadcrumb.spec.ts
orelbn: Thanks!

ppr-ui/tests/unit/Breadcrumb.spec.ts
orelbn: Thanks Ill fix all the typos

ppr-ui/tests/unit/ButtonFooter.spec.ts
orelbn: I was doing it based on the previous tests. But I don't believe its needed. Ill remove it.

ppr-ui/src/views/newMhrRegistration/MhrReviewConfirm.vue
dimak1: Ok, but interestingly enough that all tests were still passing. I know we don't have 100% coverage, so that's why we need to be careful when changing something in the existing code.

orelbn: Should be fixed now. I verified on the UI.
Agreed. Ill merge this PR to move this ticket along, and ill work on separate PR to add some tests for scrolling functionality.

orelbn:
Once the dev link is ready. Ill do one more look through before merging.

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1373-pglims1k.web.app

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1373-pglims1k.web.app

PR #1385, opened 2023-06-14
Amendment Staff Check hotfix

cameron-eyds:
*Issue #:* /bcgov/entity#

*Description of changes:*
* Amendment Staff check hotfix

### DIFF
--- a/ppr-ui/src/views/amendment/ConfirmAmendment.vue
+++ b/ppr-ui/src/views/amendment/ConfirmAmendment.vue
@@ -576,7 +576,7 @@ export default defineComponent({
         scrollToInvalid()
         return
       }
-      if ((isRoleStaffReg) || (isRoleStaffSbc)) {
+      if ((isRoleStaffReg.value) || (isRoleStaffSbc.value)) {
         localState.staffPaymentDialogDisplay = true
       } else {
         submitAmendment()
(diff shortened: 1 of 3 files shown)


dimak1 (approved):

ppr-ui/src/views/amendment/ConfirmAmendment.vue
```diff
@@ -576,7 +576,7 @@ export default defineComponent({
         scrollToInvalid()
         return
       }
-      if ((isRoleStaffReg) || (isRoleStaffSbc)) {
+      if ((isRoleStaffReg.value) || (isRoleStaffSbc.value)) {
```
dimak1, line 579: How was it working before without getting the values?

doug-lovett (approved):
Fantastic response time.

orelbn (approved):

ppr-ui/src/views/amendment/ConfirmAmendment.vue
cameron-eyds: More or less in JS would resolve as check 'Does this property exist?' yes, return true.  vs checking the refs value, which is actually a property that exists on the ref. 

ie isRoleStaffReg === { value: someBoolean }.

PR #1456, opened 2023-07-27
Add new create manufacturer info endpoint.

doug-lovett:
*Issue #:* /bcgov/entity#17063

*Description of changes:*
- Add POST /mhr/api/v1/manufacturers to support auth manufacturer account onboarding.

### DIFF
--- a/mhr_api/requirements.txt
+++ b/mhr_api/requirements.txt
@@ -59,4 +59,4 @@ sentry-sdk==1.20.0
 six==1.16.0
 strict-rfc3339==0.7
 urllib3==1.26.11
-git+https://github.com/bcgov/registry-schemas.git@1.7.9#egg=registry_schemas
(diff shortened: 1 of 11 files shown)


dimak1 (approved):
Looks good to me.

cameron-eyds (approved):
Looks great, thanks sir!

Side question, i believe we will need a GET as well for the qualified supplier information captured, is that something we've scoped out in another ticket?

doug-lovett:
To confirm, the GET was already implemented and data for existing legacy manufacturers is available. We just need to link the BCRS account id's to the records when the accounts are created.

cameron-eyds:
> To confirm, the GET was already implemented and data for existing legacy manufacturers is available. We just need to link the BCRS account id's to the records when the accounts are created.

Right on! Thank you sir

PR #1485, opened 2023-08-14
Change historical to cancelled

orelbn:
*Issue #:* /bcgov/entity#17032

*Description of changes:*
- Changes historical to cancelled and adds tests
- Clear filters when returning to dashboard

### DIFF
--- a/ppr-ui/src/components/tables/RegistrationTable.vue
+++ b/ppr-ui/src/components/tables/RegistrationTable.vue
@@ -292,6 +292,7 @@
 import {
   computed,
   defineComponent,
+  onBeforeMount,
   onUpdated,
   reactive,
   ref,
@@ -704,6 +705,11 @@ export default defineComponent({
       },
       { deep: true }
     )
+
+    // Ensures filtering is cleared when returing to dashboard from registrations
+    onBeforeMount(() => {
+      clearFilters()
+    })
 
     onUpdated(() => {
       // needed to set overrideWidth to true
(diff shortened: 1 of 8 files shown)


ppr-ui/src/components/tables/RegistrationTable.vue
```diff
@@ -705,6 +706,11 @@ export default defineComponent({
       { deep: true }
     )
 
+    // Ensures filtering is cleared when returing to dashboard from registrations
+    onBeforeMount(() => {
+      clearFilters()
+    })
+
```
cameron-eyds, line 713: Is this a request by design?

cameron-eyds (approved):
LGTM!

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1485-14w86ipi.web.app

ppr-ui/src/components/tables/RegistrationTable.vue
orelbn: No I can remove this, but I thought it was strange behaviour that when you return from a registration at the moment, it leaves the filter columns showing even though it is not filtering.

image

@tlebedovich is my assumption of the expected behaviour accurate?

`Clear filters when returning to dashboard`

cameron-eyds: Ah i see. 
Yea, i think that makes sense to clear, the above scenario is definitely confusing.

doug-lovett (approved):
Looks good.

tlebedovich:
@orelbn I think it make sense for now to clear the filters if it indeed is not actually filtering. But I will check with scott in terms of how this should actually be working and can create a ticket to make further updates in terms of when/if to save the table filtered states.

PR #1487, opened 2023-08-15
MHR API bc assessment data feed legacy owner fix

doug-lovett:
*Issue #:* /bcgov/entity###

*Description of changes:*
- legacy system BC Assessment data feed requires that the db2 table amhrtdb.owner.ownseqno increments and is the same value as amhrtdb.owner.ownerid when the tenancy type is 'JT'.


cameron-eyds (approved):
Looks good!

PR #1497, opened 2023-08-18
Separates user remarks from generated remarks

orelbn:
*Issue #:* /bcgov/entity#16833

*Description of changes:*
- Separates generated remarks from user remarks per design.
- Updates test.

Design:

image

Before:

image

After:

image


ppr-ui/src/components/unitNotes/UnitNoteContentInfo.vue
```diff
@@ -120,8 +129,23 @@ export default defineComponent({
...
   },
-  setup () {
+  setup (props) {
     const { isNoticeOfCautionOrRelatedDocType } = useMhrUnitNote()
+
+    const localState = reactive({
+      separatedRemarks: computed(() : string[] | null => {
+        if (props.note.documentType === UnitNoteDocTypes.CONTINUED_NOTE_OF_CAUTION && !props.note?.expiryDateTime) {
+          const remarks = props.note.remarks
+          const generatedRemarks = 'Continued until further order of the court. '
```
orelbn: Can be abstracted if needed in the future.

pwei1018:
Temporary Url for review: https://bcregistry-assets-dev--pr-1497-2m8mh5m3.web.app

doug-lovett (approved):
Looks good.

cameron-eyds (approved):
LGTM!

ppr-ui/src/components/unitNotes/UnitNoteContentInfo.vue
```diff
@@ -120,8 +129,23 @@ export default defineComponent({
   components: {
     BaseAddress
   },
-  setup () {
+  setup (props) {
     const { isNoticeOfCautionOrRelatedDocType } = useMhrUnitNote()
+
+    const localState = reactive({
+      separatedRemarks: computed(() : string[] | null => {
+        if (props.note.documentType === UnitNoteDocTypes.CONTINUED_NOTE_OF_CAUTION && !props.note?.expiryDateTime) {
```
cameron-eyds: very small thing but you may want to optional chain the first instance of `props.note`. if its undefined, it may not reach the second instance.

ppr-ui/src/components/unitNotes/UnitNoteContentInfo.vue
dimak1: These generated remarks could either be in a resource file or a prop text.

dimak1 (commented):
Do we need to include this additional text as part of filing?

ppr-ui/src/components/unitNotes/UnitNoteContentInfo.vue
orelbn: Given that note prop is required for this component, is it necessary to add the optional chaining?

image

I think I should remove the 2nd optional chaining actually, since note should be defined. expiryDateTime might not be defined.

cameron-eyds: Probably not, although i'm not sure what type of error you get when you don't pass a required prop. 
Either way, not sure the optional chaining would protect from a type error where it's at.

orelbn:
> Do we need to include this additional text as part of filing?

Based on the current design. It is the text shown for one of expiry date selection options. 

image

Why it is added to remarks section, is a question for the designers not something I can answer.

dimak1:
@orelbn based on the screenshot, the added text in Remarks shows _after filing_, so the answer is no, it's not included for filing. Thanks!

dimak1 (approved):

ppr-ui/src/components/unitNotes/UnitNoteContentInfo.vue
orelbn: Updated to remove the optional chaining, as I believe its not necessary.

cameron-eyds: Sounds good

PR #1514, opened 2023-08-28
search changes for data migration switch to PosgreSQL.

doug-lovett:
*Issue #:* /bcgov/entity#14522

*Description of changes:*
- Update the DB2 model unit tests to conditionally execute only if the env variable USE_LEGACY_DB is true.
- Search Step 1 Endpoint POST /mhr/api/v1/searches Update search queries for all existing search types to query the modernized PostgreSQL database using the same business requirements as the DB2 queries. Response JSON should match legacy data source JSON.
- Search Step 2 endpoint POST /mhr/api/v1/search-results/{searchId} add ability to use the modernized PosgreSQL database as the data source. Response JSON (used for the search report) should match legacy data source JSON as closely as possible.
- Review and update all search related code to conditionally use DB2 or PostgreSQL as the data source based on the USE_LEGACY_DB environment variable.
- Create or update PostgreSQL search related unit tests based on the legacy unit tests.


dimak1 (approved):

orelbn (approved):
Excellent!

PR #1572, opened 2023-10-06
Search, transfer report minor fixes.

doug-lovett:
*Description of changes:*
*Issue #:* /bcgov/entity#18097
- Search tenants in common display fractional ownership when no interest text.

*Issue #:* /bcgov/entity#18113
- UXA transfer registration report updates. 

### DIFF
--- a/mhr_api/src/mhr_api/reports/v2/report_utils.py
+++ b/mhr_api/src/mhr_api/reports/v2/report_utils.py
@@ -512,7 +512,11 @@ def format_description(description: str) -> str:
     doc_desc: str = doc_desc.lower().title()
     doc_desc = doc_desc.replace(' Of ', ' of ')
     doc_desc = doc_desc.replace(' To ', ' to ')
+    doc_desc = doc_desc.replace(' Or ', ' or ')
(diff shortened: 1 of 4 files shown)


mhr_api/src/mhr_api/reports/v2/report_utils.py
```diff
@@ -512,7 +512,11 @@ def format_description(description: str) -> str:
     doc_desc: str = doc_desc.lower().title()
     doc_desc = doc_desc.replace(' Of ', ' of ')
     doc_desc = doc_desc.replace(' To ', ' to ')
+    doc_desc = doc_desc.replace(' Or ', ' or ')
```
dimak1, line 515: There is no way to update the description itself?

dimak1 (approved):

mhr_api/src/mhr_api/reports/v2/report_utils.py
doug-lovett: Yes there is, but the MHR API follows the PPR API convention, where the database values are used as is, and the database values are uppercase. The legacy outputs are all uppercase. The case change is for report formatting. The API consumers receive the same descriptions as they do pre-migration.

PR #1583, opened 2023-10-13
Exemptions Review, Validation, Submissions and misc Component/Style Updates

cameron-eyds:
*Issue #:* /bcgov/entity#17420

*Description of changes:*
* Builds out Exemptions Review for Staff and QS
* Ties UI to exemption state
* Integrate exemption validations
* Minor config and resource updates for various common components and staff/qs specific content
* Style updates
-**Note: Trying to move towards global styling, if you noticed component local styling being removed, its because it's used in more than one spot**
-**Note: Have begun the transition to <Col 3 / Col 9> format. Will be doing some more updates to bring everything into alignment, this is going to be standardize, so we can globally adjust the widths of such in one place. This has also been confirmed with Tracey. Don't be alarmed if some components fall out of alignment in whatever filing flow, will be fixed accordingly**

* Updates common party intake form to include person / business options
* Updates schema structure to handle optional properties and dynamic 'optional' labels via the schema
* Builds common review card, to use for basic review component content. 
* Updates a few common components with Slots for easier re-use while preserving current implementation. 

**Review Notes:**
- As mentioned above, there will be a follow up Commit/PR for further column adjustments as well as updated unit testing for the new components. * Potentially AA unit test inclusions as well. 


bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1583-6hmxi5qu.web.app

ppr-ui/src/assets/styles/layout.scss
```diff
@@ -77,6 +77,10 @@
   }
 }
 
+.col-3, .col-sm-3 {
+  max-width: 25%;
+}
+
```
cameron-eyds, line 83: This is actually default by vuetifys grid system.
Why i've explicitly set this though, if we are to move towards standardizing our spacing i believe it might be important to bring visibility to it, if this were a shared file across projects ect.

I've discussed with Tracy the implications, if we stick to our 3-9 column formats for all components, reuse will become much easier. 
To clarify as well, the 3 column is typically our labels/side label column and the 9 col would contain whatever form or content we want to display. 

Seems like a simple thing, but having this consistent means we could adjust the initial label columns and/or form columns in a single place for the vast majority of our apps components.

ppr-ui/src/components/common/Attention.vue
```diff
@@ -71,7 +76,7 @@ export default defineComponent({
     const { maxLength } = useInputRules()
 
     const localState = reactive({
-      config: isRoleManufacturer.value ? attentionConfigManufacturer : attentionConfig,
+      config: props.configOverride || isRoleManufacturer.value ? attentionConfigManufacturer : attentionConfig,
```
cameron-eyds, line 79: This is just one example but ideally we don't couple our components to specific configurations (if they are configurable). 
Utilizing props or slots is always preferred for reuse. 
Being agile, we end up like this sometimes, all good, just providing a path forward for the component, can always clean up as we go.

ppr-ui/src/components/common/CertifyInformation.vue
```diff
@@ -28,8 +28,8 @@
             <!-- Table Body -->
             <tbody v-if="registeringParty.length > 0">
               <tr v-for="(item, index) in registeringParty" :key="`${item}: ${index}`" class="party-row">
-                <td class="list-item__title title-text" style="padding-left:30px">
-                  <v-icon class="v-icon mt-n2">mdi-account</v-icon><span>{{ legalName }}</span>
+                <td class="list-item__title title-text icon-text" style="padding-left:30px">
```
cameron-eyds, line 31: Simple generic class that aligns the icons with the text associated to it.

ppr-ui/src/utils/form-helper.ts
```diff
@@ -12,11 +12,31 @@ export function focusOnFirstError (formName: string): void {
...
+ **/
+export async function scrollToFirstVisibleErrorComponent (): Promise<void> {
+  const errorElements = document.getElementsByClassName('border-error-left')
+  for (const [index, element] of Array.from(errorElements).entries()) {
+    if (isElementInViewport(element)) {
+      await scrollToFirstErrorComponent(index)
+      return
+    }
+  }
+}
```
cameron-eyds, line 36: @dimak1 Really liked your simple scrollTo function, extended it a bit so we can also only scroll to visible elements. 
Context basically, in stepper flows where we have error instances across many steps in the dom, we only scroll to whats in our current viewport (ie step). 

Very simple, we just call this at root view layer and will scroll to any error border currently visible. Much improved way of scroll to vs try to track and feed identifiers to a function! Referring to what i did earlier in Registrations, doesn't age well and a bit complex.

doug-lovett (approved):
Looks great. Not that big a whale after all - lots of files with small changes.

ppr-ui/src/components/common/CertifyInformation.vue
dimak1: We should cleanup inline styles at some point.

ppr-ui/src/assets/styles/base.scss
```diff
@@ -890,4 +912,29 @@ tr.registration-row.draft-registration-row.added-reg-effect td.actions-cell,
 
 .overlap {
   z-index: 7;
+}
+
+// Contact Type Radio Btns
+.person-radio {
+  width: 47%;
```
dimak1, line 919: Another thing to cleanup (after design system in place) is these relative widths and use flexbox.

ppr-ui/src/assets/styles/layout.scss
dimak1: Hmm, `col-3` is usually a third part of row, but 25% is closer to a quoter size.

ppr-ui/src/components/common/Attention.vue
dimak1: Looking closely at the config, it is actually a simple content text. The parent should take care of it (based on role) and just pass the right content to the component. It has been done to a few components. And then we won't need configs or overrides.

ppr-ui/src/components/common/ReviewCard.vue
```diff
@@ -0,0 +1,93 @@
...
+    },
+    showIncomplete: {
+      type: Boolean,
+      default: true
+    },
+    showNotEntered: {
+      type: Boolean,
+      default: false
+    },
+    returnToRoutes: {
```
dimak1, line 58: So this is essentially a router `path` that's constructed from an array of routes?

ppr-ui/src/components/mhrTransfers/ConfirmCompletion.vue
```diff
@@ -20,117 +20,119 @@
...
+              class="py-7 px-4 my-0 confirm-checkbox"
+              :hide-details="true"
+              id="checkbox-certified"
+              v-model="confirmCompletion"
+              data-test-id="confirm-completion-checkbox"
+            >
+              <template v-slot:label>
+                <span data-test-id="confirm-checkbox-label" :class="{ 'invalid-color': showErrorComponent }">
+                  I, <strong>{{ legalName }}</strong
+                >, confirm that all of the requirements listed above have been completed.
```
dimak1: Trailing closing tag

ppr-ui/src/composables/common/usePayment.ts
```diff
@@ -0,0 +1,73 @@
+import { nextTick, reactive } from 'vue-demi'
+import { StaffPaymentOptions } from '@bcrs-shared-components/enums'
+import { StaffPaymentIF } from '@bcrs-shared-components/interfaces'
+import { useStore } from '@/store/store'
+
+export const usePayment = () => {
```
dimak1, line 9: Could you please add a general comment about this new composable?

ppr-ui/src/components/common/CertifyInformation.vue
cameron-eyds: 100%

ppr-ui/src/views/exemption/Exemptions.vue
```diff
@@ -133,8 +142,28 @@ export default defineComponent({
...
 
     const submit = async (): Promise<void> => {
+      localState.loading = true
+      localState.validate = true
+      await nextTick()
+      await scrollToFirstVisibleErrorComponent()
+
+      if (isMhrExemptionValid.value) {
+        // Construct payload
+        const payload = buildExemptionPayload()
```
dimak1, line 152: I like the _build payload_ naming, and will rename the _build api data_ functions. 👍🏻

dimak1 (approved):
Really good work 👍🏻 no complaints! Just few comments.

dimak1:
@cameron-eyds for upcoming unit tests and accessibility improvements (as mentioned). I think AA should either be mentioned in the ticket (for proper estimation) or be a separate ticket. So, I would say just adding functional unit tests would be good enough. But it's really up to you. This seems to be a big piece of work already.

ppr-ui/src/utils/form-helper.ts
dimak1: Thank you! So the scroll to visible should be used where we have the stepper.

ppr-ui/src/assets/styles/layout.scss
cameron-eyds: In Vuetify, the grid system consists of 12 Cols.

ppr-ui/src/components/common/Attention.vue
cameron-eyds: The content may be simple but in my opinion, a single config or resource object is preferable over 5-6 props etc. 
But definitely agree it should be passed to the child component in 1 or the other form.

ppr-ui/src/utils/form-helper.ts
cameron-eyds: Definitely! Seems to work good in that scenario.

ppr-ui/src/components/common/ReviewCard.vue
cameron-eyds: Exactly, you bet, array so we can pass multiple in the event of nested routes. 
Any flow with a stepper, traditionally has a `/view-route/step-route` make up etc.

ppr-ui/src/components/mhrTransfers/ConfirmCompletion.vue
cameron-eyds: Fixed!

ppr-ui/src/composables/common/usePayment.ts
cameron-eyds: Done and done.
Basically this code is duplicated in several places with Staff Payment Component. 
If we were not knocking on Vue3's door i would refactor this this now but we will localizing the staff payment shortly, can do so at that point.

ppr-ui/src/views/exemption/Exemptions.vue
cameron-eyds: Right on!

cameron-eyds:
> @cameron-eyds for upcoming unit tests and accessibility improvements (as mentioned). I think AA should either be mentioned in the ticket (for proper estimation) or be a separate ticket. So, I would say just adding functional unit tests would be good enough. But it's really up to you. This seems to be a big piece of work already.

Probably right, perhaps we leave for those dedicated AA tickets we've already groomed!

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1583-6hmxi5qu.web.app

ppr-ui/src/components/common/Attention.vue
dimak1: Yea only one resource object could be passed (with all the props) which can be added/removed without changing the component's signature, eg:
```
 :content="{
          title: 'Document ID',
          description: 'Enter the 8-digit Document ID number.',
          sideLabel: 'Document ID'
        }"
```

ppr-ui/src/composables/common/usePayment.ts
dimak1: Thank you!

PR #1615, opened 2023-11-06
Vuetify Fixes for MHR and Misc Features

cameron-eyds:
*Issue #:* /bcgov/entity#18391

*Description of changes:*
* Following the patterns for PPR, continued the refactors into MHR Related flows. 
* Replaces v-mask with maska package for our phone number masking, as a global directive
* Registration Table refactors - leverage virtual scroller so we can pull in the entirety of the registration data set and prevent having pseudo pagination. Improves performance greatly for filtering and swapping tabs. 
* Minor fixes to svg styling for stepper

* Flows covered in this ticket: 
 MHR Registrations Table
 MHR Registration Flows
 User Access Flows
 MHR Unit Notes
 MHR Information Flows
 Exemption Flow

*Notes:*
- Keep in mind this was functionality first, so there will be some minor UX or stylings i'm sure i missed. 
- We are also still missing a Date Picker, so there is some errors where that component is referenced (Future Ticket)


doug-lovett (approved):
Looks great Cameron, a fair bit of cleanup and refactoring.

cameron-eyds:
> Looks great Cameron, a fair bit of cleanup and refactoring.

Thank you!

ppr-ui/src/main.ts
```diff
@@ -8,6 +8,7 @@ import { getVueRouter } from '@/router'
 import { getPiniaStore } from '@/store'
 import * as Sentry from '@sentry/vue'
 import vuetify from './plugins/vuetify'
+import { vMaska } from 'maska'
```
dimak1, line 11: 👍🏻

ppr-ui/src/utils/feature-flags.ts
```diff
@@ -10,10 +10,10 @@ export const defaultFlagSet: LDFlagSet = {
   'search-registration-number': true,
   'search-serial-number': true,
   'mhr-ui-enabled': true, // Mhr Search - default true: Should remove from codebase
-  'mhr-registration-enabled': false, // Enables MHR table tab
-  'mhr-transfer-enabled': false, // Enables changes to base MHR Home Owners within the MHR Information flow
+  'mhr-registration-enabled': true, // Enables MHR table tab
```
dimak1, line 13: Are we permanently enabling these now?

ppr-ui/src/resources/maskConfigs.ts
dimak1, line 1: Good idea to keep 'em clean.

dimak1 (approved):
Looks great! Good one 👏🏻 

Just curious if it is required to add dashes to the props of the components? Maybe we can update the lint rule. The reason is that it's easier to search for it in files.

cameron-eyds:
> Looks great! Good one 👏🏻
> 
> Just curious if it is required to add dashes to the props of the components? Maybe we can update the lint rule. The reason is that it's easier to search for it in files.

@dimak1 Great point for sure!
It was recommended by the linter as you've already deduced (ie it was the basic community standards) but i totally agree, this might be something we want to override. 
Your reason is alone is a great reason to do so. 
Will 100% do this as part of the final linting sweep, will include as part of the unit testing ticket scope.

ppr-ui/src/views/Dashboard.vue
```diff
@@ -131,23 +131,23 @@
             class="mt-13"
             :app-loading-data="loading"
             :app-ready="appReady"
-            @snackBarMsg="snackBarEvent($event)"
+            @snack-bar-msg="snackBarEvent($event)"
```
cameron-eyds, line 134: Absolutely, we can/will override this. The reason you mentioned is a very good reason to do so. There is a lot of linting fixies to come in, will make this adjustment in the unit testing ticket. 
I see that ticket as a means of getting both linting and unit test CI tests Greeen.

ppr-ui/src/utils/feature-flags.ts
cameron-eyds: Not permanently, no, just left for ease of development.

ppr-ui/src/main.ts
cameron-eyds: Super easy replacement. Defined globally, so we can use in any template, just need to import basically the mask config. That being said, we don't really mask anything but phone numbers but the options there :)

dimak1:
@cameron-eyds great, yea that would be good if we are in agreement (about overriding the lint rule). So do we need to go back (in final lint sweep) and update the names to remove the dashes?

ppr-ui/src/utils/feature-flags.ts
dimak1: Yes, it would make it easier, so I don't have to revert these changes each time I rebase :)

cameron-eyds:
> @cameron-eyds great, yea that would be good if we are in agreement (about overriding the lint rule). So do we need to go back (in final lint sweep) and update the names to remove the dashes?

I believe (am hopeful) the editor will take care of this, once the rule is overridden. 
If not, i will adjust the rules and following the lint issues to fix alongside the other misc lint items.

ppr-ui/src/main.ts
dimak1: If we use just one mask (for phone) do we even need this package? On the other hand, it is small and has no additional dependencies so maybe we can keep it.

ppr-ui/package.json
```diff
@@ -1,14 +1,14 @@
...
+  "version": "3.0.2",
   "private": true,
   "appName": "Assets UI",
   "sbcName": "SBC Common Components",
   "scripts": {
     "dev": "vite",
     "build": "vite build",
     "serve": "vite preview",
-    "lint": "eslint . --ext .js,.ts,.vue",
+    "lint": "eslint . --ext .js,.ts,.vue --fix",
```
cameron-eyds: I'll have to revert this

ppr-ui/src/main.ts
cameron-eyds: I suppose its a nice option for future. The package is pretty small and very easy to implement/maintain.

ppr-ui/package.json
cameron-eyds: Done

cameron-eyds:
@dimak1 Set up some additional linting rules and also helped identify certain vue warnings as errors. 
In doing so, it helps the auto linter fix some of said issues. 
Reverted all props and @ events to camelCase and put in some lint rules to enforce that.

dimak1:
> @dimak1 Set up some additional linting rules and also helped identify certain vue warnings as errors. In doing so, it helps the auto linter fix some of said issues. Reverted all props and @ events to camelCase and put in some lint rules to enforce that.

@cameron-eyds thank you for doing that!

dimak1:
@cameron-eyds since we are on linter topic, maybe we could also look into prettier as well, just at the high level. Because I stopped using it a long time ago, since it was causing conflicts between other devs (I know Orel wasn't using it either because of it). Maybe it would be a good opportunity for all of us start using an external prettier config, so we don't have conflicts. By no means I want to increase the scope further, just putting it out here...

cameron-eyds:
> @cameron-eyds since we are on linter topic, maybe we could also look into prettier as well, just at the high level. Because I stopped using it a long time ago, since it was causing conflicts between other devs (I know Orel wasn't using it either because of it). Maybe it would be a good opportunity for all of us start using an external prettier config, so we don't have conflicts. By no means I want to increase the scope further, just putting it out here...

Absolutely, another great point!
Lets make a note to have a closer look at that.
 
Lots to do :) Just need to find the scope to do so, i think now is a great time to do it though, i'll give it a go in the Unit Test/Clean up Ticket.

PR #1637, opened 2023-12-01
Transfer Validation and Submission Updates

cameron-eyds:
*Issue #:* /bcgov/entity#18842

*Description of changes:*
- Fixes allocation validation from being false when dealing with Group -> Sole/JT Home Tenancy Transfers
- Fixes ownLand from falling back to `null` ie `None` which fails in the schema and prevents transfer submissions. 

### DIFF
--- a/ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
+++ b/ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
@@ -353,7 +353,7 @@ export const useMhrInformation = () => {
       declaredValue: getMhrTransferDeclaredValue.value,
       consideration: getMhrTransferConsideration.value,
       transferDate: getMhrTransferDate.value,
-      ownLand: getMhrTransferOwnLand.value || null,
+      ownLand: getMhrTransferOwnLand.value,
       ...(getMhrTransferDocumentId.value && {
         documentId: getMhrTransferDocumentId.value
       }),
--- a/ppr-ui/package-lock.json
+++ b/ppr-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ppr-ui",
-  "version": "2.1.26",
+  "version": "2.1.27",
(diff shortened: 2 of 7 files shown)


ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
```diff
@@ -353,7 +353,7 @@ export const useMhrInformation = () => {
       declaredValue: getMhrTransferDeclaredValue.value,
       consideration: getMhrTransferConsideration.value,
       transferDate: getMhrTransferDate.value,
-      ownLand: getMhrTransferOwnLand.value || null,
+      ownLand: getMhrTransferOwnLand.value,
```
dimak1, line 356: Hey @RuoxuanPengBC, I remember you had similar default value in your PR (for Land Lease radios)?

dimak1 (approved):

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1637-6bawadpa.web.app

ppr-ui/src/composables/mhrInformation/useMhrInformation.ts
cameron-eyds: Totally!
I think in the Registration flow, null was set to determine no-entry-made when the user navigated to the Review Confirm step. 

In the Transfer flow, it's required to navigate to review confirm, with no indeterminate state ie non selection state etc, and the schema doesn't like those `null` properties when types are enforced.

If it was an optional property (i didn't check), it would have to be omitted from the payload vs a null value.

dimak1: Thanks for clarification. I recall this was a mandatory in at least one of the flows... maybe Roy would have some input.

doug-lovett (approved):
Looks good.

PR #1646, opened 2023-12-06
Axios version and interceptor upgrades

cameron-eyds:
*Issue #:* /bcgov/entity#16983

*Description of changes:*
* Upgrade axios version in both Assets and SBC CC projects 
* SBC version bump to pull in axios version bump
* Axios interceptor updates to handle updated header object structure
* Restore Preview build command in package.json

### DIFF
--- a/ppr-ui/package.json
+++ b/ppr-ui/package.json
@@ -1,11 +1,11 @@
 {
   "name": "ppr-ui",
-  "version": "3.0.5",
+  "version": "3.0.6",
   "private": true,
   "appName": "Assets UI",
   "sbcName": "SBC Common Components",
   "scripts": {
-    "dev": "vite",
+    "dev": "vite dev",
     "build": "vite build",
     "serve": "vite preview --port 8080",
     "lint": "eslint . --ext .js,.ts,.vue",
@@ -36,7 +36,7 @@
     "moment": "^2.29.1",
     "moment-timezone": "^0.5.34",
     "pinia": "^2.1.3",
-    "sbc-common-components": "^4.0.3",
+    "sbc-common-components": "4.0.3",
     "vue": "^2.0.0 || >=3.3.4",
     "vue-pdf-embed": "1.1.6",
     "vue-plugin-helper-decorator": "^0.0.12",
--- a/ppr-ui/package.json
+++ b/ppr-ui/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "serve": "vite dev",
+    "serve": "vite preview --port 8080",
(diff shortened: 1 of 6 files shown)


ppr-ui/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "serve": "vite dev",
+    "serve": "vite preview --port 8080",
```
cameron-eyds, line 10: @dimak1 `npm run dev` for local dev 👍 
`serve` for viewing production ready build locally, following a local build.

dimak1: I had a suspicion about this one, I looked at the docs quickly and didn't see a `preview` for prod builds. Good find! 👍🏻  We probably need to have another run command for local dev as well

dimak1 (approved):

ppr-ui/package.json
cameron-eyds: `npm run dev` is the one for local dev. 
Vite doesn't need an argument for a dev build it seems, `vite` === `dev build` by default looks like.

https://vitejs.dev/guide/cli

severinbeauvais: Since you've changed this file, I recommend updating the patch version on line 3.

ppr-ui/package.json
```diff
@@ -29,14 +29,14 @@
...
-    "axios": "^0.26.1",
+    "axios": "^1.6.2",
     "flush-promises": "^1.0.2",
     "http-status-codes": "^2.1.4",
     "material-design-icons": "^3.0.1",
     "moment": "^2.29.1",
     "moment-timezone": "^0.5.34",
     "pinia": "^2.1.3",
-    "sbc-common-components": "4.0.2",
+    "sbc-common-components": "^4.0.3",
```
severinbeauvais: You know this is risky because you might pick up changes you don't know about, right?

ppr-ui/package.json
cameron-eyds: Done!

severinbeauvais (approved):
Yikes. We're all grateful you've worked through this!

ppr-ui/package.json
cameron-eyds: Perhaps eventually, you bet!
At the moment, i'm the only one contributing to this version of SBC-CC but still, your probably right in that i will forget to change 😁

cameron-eyds:
> Yikes. We're all grateful you've worked through this!

@seeker25 Helped me discover the root causes here from Axios, luckily the latest versions had resolved this. 
Others definitely bumped into this!

doug-lovett (approved):
Looks great.

PR #1686, opened 2024-01-11
User Access + Global Address Fixes

cameron-eyds:
*Issue #:* /bcgov/entity#16952
/bcgov/entity#17647
/bcgov/entity#17463

*Description of changes:*
* Misc UXA fixes for Home Location and QS Information view
* Correct default values for address country and regions to v-selects thinking there was a selected value.
The above is a vuetify issue (or feature if you ask them) that allows empty strings to present as a legit value. 
https://github.com/vuetifyjs/vuetify/issues/16372

* Includes authHandler in app.vue to verify the CURRENT_ACCOUNT is set before initializing the application, prevent any timing issues where auth is slower than app initialization.
* Minor update to transfer type selector to prevent missed event firings on selection
* Modify CivicAddress to take it's schema as a prop, for reuse. 

### DIFF
--- a/ppr-ui/src/App.vue
+++ b/ppr-ui/src/App.vue
@@ -233,13 +233,21 @@ export default defineComponent({
           }
         }
 
-        // When we are authenticated, allow time for session storage propagation from auth, then initialize application
-        // (since we won't get the event from Signin component)
-        if (isAuthenticated.value) {
-          setTimeout(() => { onProfileReady(true) }, 1000)
-        }
+        authVerificationHandler()
       }
     })
+
+    /**
+     * When we are authenticated, allow time for session storage propagation from auth, then initialize application
+     * (since we won't get the event from signin component)
+     */
+    const authVerificationHandler = () => {
+      setTimeout(() => {
+        isAuthenticated.value && !!sessionStorage.getItem(SessionStorageKeys.CurrentAccount)
+          ? onProfileReady(true)
+          : authVerificationHandler()
+      }, 2000)
+    }
 
     const payErrorDialogHandler = (confirmed: boolean) => {
       const flowType = getRegistrationFlowType.value
@@ -663,7 +671,12 @@ export default defineComponent({
 
     /** Called when profile is ready -- we can now init app. */
     watch(() => localState.profileReady, async (val: boolean) => {
+      console.log(sessionStorage.getItem(SessionStorageKeys.CurrentAccount))
       await onProfileReady(val)
+    })
+
+    watch(() => sessionStorage.getItem(SessionStorageKeys.CurrentAccount), async (val: object) => {
+      console.log(val)
     })
 
     return {
--- a/ppr-ui/src/App.vue
+++ b/ppr-ui/src/App.vue
@@ -671,12 +671,7 @@ export default defineComponent({
 
     /** Called when profile is ready -- we can now init app. */
     watch(() => localState.profileReady, async (val: boolean) => {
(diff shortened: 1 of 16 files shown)


ppr-ui/src/components/mhrTransfers/TransferType.vue
```diff
@@ -64,6 +64,7 @@
                   class="copy-normal gray7"
                   v-bind="props"
                   :title="null"
+                  @click="handleTypeChange(item.raw)"
```
dimak1, line 67: Nice, so this fixes the issue with not firing sometimes?

ppr-ui/src/resources/userAccessOrgLookup.ts
```diff
@@ -3,16 +3,15 @@ import { MhrSubTypes } from '@/enums'
 
 export const UserAccessOrgLookupConfig: Record<MhrSubTypes, OrgLookupConfigIF> = {
   [MhrSubTypes.LAWYERS_NOTARIES]: {
```
dimak1, line 5: Should the enum be updates as well? Because the title changed.

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1686-ne5p4br3.web.app

ppr-ui/src/App.vue
```diff
@@ -233,14 +233,22 @@ export default defineComponent({
...
+    /**
+     * When we are authenticated, allow time for session storage propagation from auth, then initialize application
+     * (since we won't get the event from signin component)
+     */
+    const authVerificationHandler = () => {
+      setTimeout(() => {
+        isAuthenticated.value && !!sessionStorage.getItem(SessionStorageKeys.CurrentAccount)
+          ? onProfileReady(true)
+          : authVerificationHandler()
+      }, 2000)
```
dimak1, line 249: That's the lowest we can go, is it noticeable delay? Or you wanted to have some buffer for the future?

dimak1 (approved):
👍🏻

ppr-ui/src/components/mhrTransfers/TransferType.vue
cameron-eyds: It did in my testing, yes. 
It only seemed to miss firing when the selection was made on the very edges of the element because the event was bound to the item-title, not the item itself. 
So hopefully this resolves whatever associated issues were happening here!

ppr-ui/src/resources/userAccessOrgLookup.ts
cameron-eyds: I'm not sure what you mean, is this in reference to the pluralTitle property value changing?
If so, i did ask Tracey explicitly about this one, for Lawyers and Notaries, design wanted this particular label to be generic (ie Qualified Suppliers).

ppr-ui/src/App.vue
cameron-eyds: It's not really noticeable no. 
This function is in itself a buffer, in the rare event CURRENT ACCOUNT is missing, it will call itself again after 2seconds. 
I think if we were to add more time here, it would just delay that initial initialization when everything was good, which will be the scenario 9/10 times pretty sure.

ppr-ui/src/resources/userAccessOrgLookup.ts
dimak1: Yea it was a reference to the `pluralTitle`. Sounds like it's a harmless content change.

doug-lovett (approved):
Looks great.

PR #1690, opened 2024-01-15
Manufacturer Submission

cameron-eyds:
*Issue #:* /bcgov/entity#17464

*Description of changes:*
* Factor in Manufacturer payload and submission for QS Application

Submission works, see screenshot 😆 
Screenshot 2024-01-15 at 3 17 25 PM

### DIFF
--- a/ppr-ui/src/composables/userAccess/useUserAccess.ts
+++ b/ppr-ui/src/composables/userAccess/useUserAccess.ts
@@ -1,16 +1,19 @@
 import { computed, ComputedRef, nextTick } from 'vue'
 import {
   cleanEmpty,
-  convertDate, createManufacturer,
+  convertDate,
+  createManufacturer,
   createQualifiedSupplier,
   fromDisplayPhone,
   getAccountInfoFromAuth,
   getFeatureFlag,
-  getKeyByValue, getMhrManufacturerInfo,
+  getKeyByValue,
+  getMhrManufacturerInfo,
   getQsServiceAgreements,
   getQualifiedSupplier,
   hasTruthyValue,
-  requestProductAccess, updateManufacturer,
+  requestProductAccess,
+  updateManufacturer,
   updateQualifiedSupplier,
   updateUserSettings
 } from '@/utils'
--- a/ppr-ui/src/composables/userAccess/useUserAccess.ts
+++ b/ppr-ui/src/composables/userAccess/useUserAccess.ts
@@ -1,24 +1,33 @@
 import { computed, ComputedRef, nextTick } from 'vue'
 import {
   cleanEmpty,
-  convertDate,
+  convertDate, createManufacturer,
   createQualifiedSupplier,
   fromDisplayPhone,
   getAccountInfoFromAuth,
   getFeatureFlag,
-  getKeyByValue,
+  getKeyByValue, getMhrManufacturerInfo,
   getQsServiceAgreements,
   getQualifiedSupplier,
   hasTruthyValue,
-  requestProductAccess,
+  requestProductAccess, updateManufacturer,
   updateQualifiedSupplier,
   updateUserSettings
 } from '@/utils'
-import { MhrSubTypes, ProductCode, ProductStatus, RouteNames, SettingOptions } from '@/enums'
+import {
+  ApiHomeTenancyTypes,
+  HomeLocationTypes,
+  HomeOwnerPartyTypes,
+  MhrSubTypes,
+  ProductCode,
+  ProductStatus,
+  RouteNames,
+  SettingOptions
+} from '@/enums'
 import { storeToRefs } from 'pinia'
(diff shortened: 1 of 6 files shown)


bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1690-kg0je390.web.app

doug-lovett (approved):
Looking good.

ppr-ui/src/composables/userAccess/useUserAccess.ts
```diff
@@ -324,6 +328,90 @@ export const useUserAccess = () => {
...
+      ...cleanEmpty(getMhrQsInformation.value) as MhrQsPayloadIF,
+      authorizationName: getMhrQsAuthorization.value.authorizationName,
+      phoneNumber: fromDisplayPhone(getMhrQsInformation.value.phoneNumber)
+    }
+
+    // Check for current QS Record
+    const hasQsRecord = await getQualifiedSupplier()
+
+    // Create or Update based on previous record
+    return hasQsRecord
```
dimak1, line 347: So we are making api cal `getQualifiedSupplier` to check for QS record and then we making another api call to either `update` to `create` qualified supplier. UI is controlling this logic? Should this be done on backend, or we are reusing existing api's? cc: @doug-lovett

ppr-ui/src/interfaces/mhr-registration-interfaces/MhrRegistrationHomeLocationIF.ts
```diff
@@ -1,8 +1,8 @@
 import { HomeLocationTypes } from '@/enums'
 
 export interface MhrRegistrationHomeLocationIF {
```
dimak1, line 3: Just for clarification, why set _almost_ all fields to optional?

dimak1 (approved):
Looks good, just few clarification comments.

ppr-ui/src/interfaces/mhr-registration-interfaces/MhrRegistrationHomeLocationIF.ts
cameron-eyds: Definitely!

Most of the home location properties are optional to begin with, this goes back to the MH Registration itself. They could have been made optional then, when first implemented. I have done so now so we could utilize the same interface, for the same model, in the different flow. 
But yea, they are all optional, only some are ever required and that is really dependent on the type of selection the user makes for home location type.

dimak1: I see. The interfaces should help us with the development, to enforce type safety and hopefully have less issues. With all fields optional, yes it makes more flexible/reusable, but it's harder to catch potential errors. 
Should we make it more strict and/or create another IF or keep as is? Thoughts?

cameron-eyds: I understand the purpose of interfaces. 

It's about as strict as it should be i think. If you reference the api schema, its of similar design. ie some required, some not. 
If you want to create an individual interface for every data permutation, have at er, but you will still end up with optional properties where the properties are... optional. 

I would argue the interface does it job well enough in informing us what should/could be in this data set. 
If i was to create another interface it would look near identical because this data set has a lot of optional properties.

ppr-ui/src/composables/userAccess/useUserAccess.ts
cameron-eyds: Are these questions for me or for Doug?

cameron-eyds: I can answer what i know. 

We are making a call to see if the record exists for the user, yes (either qs or mf). If it exists we will update it, if not we will create it. 
Yes the UI is doing this, as you can see. 

Should we be doing this in only one request? I'm not sure, i don't think so. 
Doug could probably answer much more clearly what best practices are but in my opinion isolated resources is probably a better way to follow the REST principles.

dimak1: Yea, I can see the UI is doing multiple get/post calls, which ideally should be only one post submission call, and updated or create should be done on the backend.

cameron-eyds: A POST should create, a PUT should update. 
In my opinion, i don't think a POST should do both.
 
Is that what your asking?

dimak1: Yes, I agree, I just think it should be handled on the backend (as it's more of a business logic) rather than driven by the UI. But I can get used to this approach as well.

cameron-eyds: Either way, it's a good point to bring up and consider 👍

PR #1702, opened 2024-01-23
Fix location type for mhr submission

dimak1:
Fix MHR Submission bug where `locationType` is not properly set.


cameron-eyds (approved):
LG!

PR #1706, opened 2024-01-26
Update PPR party client code formatting and matching.

doug-lovett:
*Issue #:* /bcgov/entity#19436

*Description of changes:*
- Format client party codes with leading zeroes. Currently 0115 returns no results when 3 codes exist with a head office code of 115.
- The first 4 digits will always match on the head office code. Currently 001 and 0001 match on 00000001. 001 will match on head office codes 1 and 10 to 19. 0001 will match on head office code 1.
- When matching on more than 4 digits, match on the formatted values with leading zeroes.
- Sort by party code when multiple results exist.


dimak1 (approved):

cameron-eyds (approved):
Looks good!

PR #1724, opened 2024-02-07
Get current MH info conditionally add new property exemptDateTime.

doug-lovett:
*Issue #:* /bcgov/entity#19705

*Description of changes:*
- If the home is in an exempt state, include exemptDateTime in the GET /mhr/api/v1/registrations/{mhr_number} response.


cameron-eyds (approved):
Looks good!

dimak1 (approved):

PR #1739, opened 2024-02-16
Updated Badge Component

cameron-eyds:
*Issue #:* /bcgov/entity#19503

*Description of changes:*
* Created small badge component which takes two data sets, compares them and renders when there is differences. 
* Deep comparison function will normalize strings, take numbers, booleans, objects and arrays, including nested objects and arrays (see unit tests).
* Lots of unit tests
* Included screens for example implementation:

(baseline would take the snapshot value, just hardcoded this for the example)
Screenshot 2024-02-16 at 9 29 46 AM
Screenshot 2024-02-16 at 9 29 26 AM
Screenshot 2024-02-16 at 9 29 20 AM

### DIFF
--- a/ppr-ui/src/components/common/UpdatedBadge.vue
+++ b/ppr-ui/src/components/common/UpdatedBadge.vue
@@ -1,0 +1,32 @@
+<template>
+  <v-chip
+    v-if="hasChanges"
+    id="updated-badge-component"
+    xSmall
+    variant="flat"
+    color="primary"
+    :data-test-id="`${action}-badge`"
+  >
+    <b>{{ action }}</b>
+  </v-chip>
+</template>
+<script setup lang="ts">
+import { computed } from 'vue'
+import { BaseDataUnionIF } from '@/interfaces'
+import { deepChangesComparison } from '@/utils'
+
+const props = withDefaults(defineProps<{
+  action?: string,
+  baseline: BaseDataUnionIF,
+  currentState: BaseDataUnionIF
+}>(), {
+  action: 'CORRECTED',
+  baseline: null,
+  currentState: null
+})
+
+/** Is true when there is a difference between the baseline and current state **/
+const hasChanges = computed(() => {
+  return deepChangesComparison(props.baseline, props.currentState)
+})
+</script>
--- a/ppr-ui/src/utils/utilities.ts
+++ b/ppr-ui/src/utils/utilities.ts
@@ -1,3 +1,5 @@
+import { BaseDataUnionIF } from '@/interfaces'
+
 export function isSigningIn (): boolean {
   const path = window.location.pathname
   return path.includes('/login') || path.includes('/signin')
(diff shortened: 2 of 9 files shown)


bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1739-04mdadnb.web.app

ppr-ui/src/components/common/UpdatedBadge.vue
```diff
@@ -0,0 +1,32 @@
...
+
+const props = withDefaults(defineProps<{
+  action?: string,
+  baseline: BaseDataUnionIF,
+  currentState: BaseDataUnionIF
+}>(), {
+  action: 'CORRECTED',
+  baseline: null,
+  currentState: null
+})
```
cameron-eyds, line 26: @dimak1 Small heads up here, when deconstructing props to apply defaults as in MhrTransportPermit line 258, it has reactivity implications. I followed that same approach initially and was losing reactivity on those props. 

Screenshot 2024-02-16 at 11 18 39 AM

cameron-eyds: Might just be when trying to type our props, as it states in the documentation there. 
Either way, not sure if it was an issue in the TransportPermit, just something to be aware of!

doug-lovett (approved):
Very nice.

ppr-ui/src/utils/utilities.ts
```diff
@@ -81,3 +83,31 @@ export const filterDuplicates = (list: Array<any>, filterBy: string) => {
...
+ * @param {*} current - The second value to compare.
+ * @returns {boolean} - Returns true if the values are different, false if they are equal.
+ */
+export const deepChangesComparison = (base: BaseDataUnionIF, current: BaseDataUnionIF): boolean => {
+  // Object safety-check
+  const isObject = (value) => typeof value === 'object' && value !== null
+  // String normalization safety-check: Case Insensitive
+  const caseInsensitiveStringCompare = (val1, val2) => {
+    if (typeof val1 === 'string' && typeof val2 === 'string') {
+      return val1.toUpperCase() !== val2.toUpperCase()
```
dimak1, line 100: So we are ignoring the case? Liz was ok with that?

dimak1 (approved):
The only thing I can think of is performance of the deep comparison. In case we have many badges (Review page) and it takes long to load, we can future proof by making it async. It is not a requirement however, so time will show.
Nice work!

ppr-ui/src/utils/utilities.ts
cameron-eyds: It was an explicit requirement in corrections, so it should be the case here too yea. 
If they have issue, we can option this.

cameron-eyds:
> The only thing I can think of is performance of the deep comparison. In case we have many badges (Review page) and it takes long to load, we can future proof by making it async. It is not a requirement however, so time will show. Nice work!

Performance impact is negligible but it depends on the implementation. It hinges on the length and depth of the objects/arrays being passed. 
This is why we are only passing in the isolated models of the component/input we are evaluating vs compare an entire filing model against an entire snapshot.

PR #1740, opened 2024-02-17
UI amend permit API add permitLandStatusConfirmation

doug-lovett:
*Issue #:* /bcgov/entity#19870

*Description of changes:*
- When amending a transport permit the UI needs to know the land status confirmation value of the latest transport permit. This value is only available for existing, modernized transport permit registrations. It does not exist in the legacy data model. Conditionally add a new property named "permitLandStatusConfirmation" to the GET /mhr/api/v1/regstrations/{mhr_number}?current=true response.

Note: the API will not return permitLandStatusConfirmation if amending a legacy transport permit.


dimak1 (approved):

cameron-eyds (approved):
Looks great!

PR #1746, opened 2024-02-21
Enabled/Disabled Transport Permit Address Condition

cameron-eyds:
*Issue #:* /bcgov/entity#19626

*Description of changes:*
* Enable/Disable Transport permits condition updated: now compares Dealer or Manufacturer Address against Home Civic Address instead of the Owners Address.
* User Access Phone Formatting

### DIFF
--- a/ppr-ui/src/composables/userAccess/useUserAccess.ts
+++ b/ppr-ui/src/composables/userAccess/useUserAccess.ts
@@ -32,7 +32,13 @@ import {
 import { storeToRefs } from 'pinia'
 import { useStore } from '@/store/store'
 import { useAuth, useNavigation } from '@/composables'
-import { MhrManufacturerInfoIF, MhrQsPayloadIF, UserAccessMessageIF, UserProductSubscriptionIF } from '@/interfaces'
+import {
+  AddressIF,
+  MhrManufacturerInfoIF,
+  MhrQsPayloadIF,
+  UserAccessMessageIF,
+  UserProductSubscriptionIF
+} from '@/interfaces'
 
 export const useUserAccess = () => {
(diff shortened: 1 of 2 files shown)


bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1746-nyq3ux6m.web.app

ppr-ui/src/composables/userAccess/useUserAccess.ts
```diff
@@ -449,7 +451,7 @@ export const useUserAccess = () => {
       submittingParty: {
         businessName: getMhrQsSubmittingParty.value.name,
         address: getMhrQsSubmittingParty.value.mailingAddress,
-        phoneNumber: getMhrQsSubmittingParty.value.phoneNumber,
+        phoneNumber: fromDisplayPhone(getMhrQsSubmittingParty.value.phoneNumber),
```
dimak1, line 454: This would keep only the digits? I had problems when the submitting party was pre-filled, then it had dashes and submission failed...

dimak1 (approved):

doug-lovett (approved):
Rocks!

ppr-ui/src/composables/userAccess/useUserAccess.ts
cameron-eyds: Yep, exactly! 
It should have been formatted all along. 

This is returned from the api but the real issue was in the user access submission (specifically Manufacturers),
It was saving the first 10 digits and wasn't correctly formatted in submissions payload. 

So we were submitting `(123) 123-4567` and it would store `(123) 123-` for the owners.
More of a safe-guard here but more importantly fixed in the payload submission down below.

PR #1761, opened 2024-03-01
Amended badges on Review and Confirm page

dimak1:
*Issue #:* 
- bcgov/entity#19861

*Description of changes:*
- Amended badges on Review page
- Registration Status amended badge
- Submission update
- Unit tests

Screenshot 2024-03-01 at 09 09 01

### DIFF
--- a/ppr-ui/src/components/tombstone/TombstoneDynamic.vue
+++ b/ppr-ui/src/components/tombstone/TombstoneDynamic.vue
@@ -183,7 +183,7 @@ import { useMhrCorrections, useMhrInformation, useTransportPermits } from '@/com
 import { storeToRefs } from 'pinia'
 import { MhrCorrectionClient, MhrCorrectionStaff } from '@/resources'
 import MhrStatusCorrection from '@/components/mhrRegistration/MhrStatusCorrection.vue'
-import { UpdatedBadge } from '../common'
+import { UpdatedBadge } from '@/components/common'
 
 export default defineComponent({
   name: 'TombstoneDynamic',
--- a/ppr-ui/src/components/tombstone/TombstoneDynamic.vue
+++ b/ppr-ui/src/components/tombstone/TombstoneDynamic.vue
@@ -40,18 +40,17 @@
           class="fs-16 pr-5"
           noGutters
         >
-          <v-col cols="10" />
-          <v-col
-            class="generic-label text-no-wrap"
-            cols="1"
-          >
-            <span class="float-right">Registration Status: </span>
-          </v-col>
-          <v-col
-            class="pl-3"
-            cols="1"
-          >
-            <p>{{ statusType }}</p>
+          <v-col>
+            <div class="float-right">
+              <span class="generic-label mr-2">Registration Status: </span> {{ statusType }}
+              <UpdatedBadge
+                v-if="showAmendedRegStatusBadge"
+                action="AMENDED"
+                :baseline="getMhrOriginalTransportPermitRegStatus"
+                :currentState="getMhrInformation.statusType"
+                style="position: absolute"
+              />
+            </div>
           </v-col>
         </v-row>
       </v-col>
@@ -180,14 +179,15 @@ import { useStore } from '@/store/store'
 import { formatExpiryDate, pacificDate } from '@/utils'
 import { RegistrationTypeIF } from '@/interfaces'
 import { MhApiStatusTypes, MhUIStatusTypes } from '@/enums'
-import { useMhrCorrections, useMhrInformation } from '@/composables'
+import { useMhrCorrections, useMhrInformation, useTransportPermits } from '@/composables'
 import { storeToRefs } from 'pinia'
 import { MhrCorrectionClient, MhrCorrectionStaff } from '@/resources'
 import MhrStatusCorrection from '@/components/mhrRegistration/MhrStatusCorrection.vue'
+import { UpdatedBadge } from '../common'
 
 export default defineComponent({
   name: 'TombstoneDynamic',
-  components: { MhrStatusCorrection },
+  components: { MhrStatusCorrection, UpdatedBadge },
   props: {
     isMhrInformation: {
       type: Boolean,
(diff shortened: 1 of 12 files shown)


bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1761-o7ymfxuz.web.app

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1761-o7ymfxuz.web.app

doug-lovett (approved):
Looks great.

ppr-ui/src/components/tombstone/TombstoneDynamic.vue
```diff
@@ -180,14 +179,15 @@ import { useStore } from '@/store/store'
 import { formatExpiryDate, pacificDate } from '@/utils'
 import { RegistrationTypeIF } from '@/interfaces'
 import { MhApiStatusTypes, MhUIStatusTypes } from '@/enums'
-import { useMhrCorrections, useMhrInformation } from '@/composables'
+import { useMhrCorrections, useMhrInformation, useTransportPermits } from '@/composables'
 import { storeToRefs } from 'pinia'
 import { MhrCorrectionClient, MhrCorrectionStaff } from '@/resources'
 import MhrStatusCorrection from '@/components/mhrRegistration/MhrStatusCorrection.vue'
+import { UpdatedBadge } from '../common'
```
cameron-eyds: Will the alias path work here?
`import { UpdatedBadge } from '@/components/common'`

dimak1: Probably, it was auto-imported, will update

cameron-eyds (approved):
Looks great!

ppr-ui/src/components/tombstone/TombstoneDynamic.vue
dimak1: Done!

ppr-ui/src/composables/mhrInformation/useTransportPermits.ts
```diff
@@ -263,7 +275,8 @@ export const useTransportPermits = () => {
         reserveNumber: '',
         exceptionPlan: ''
       } as MhrRegistrationHomeLocationIF,
-      ownLand: null
+      ownLand: null,
+      registrationStatus: ''
```
dimak1, line 279: @cameron-eyds just a note here, this is original status for the registration, might not be the best place...

ppr-ui/src/interfaces/mhr-registration-interfaces/MhrTransportPermitIF.ts
```diff
@@ -11,4 +11,5 @@ export interface MhrTransportPermitIF {
   ownLand: boolean,
   landStatusConfirmation?: boolean,
   amendment?: boolean
+  registrationStatus?: string // UI only prop for Amended badge
```
dimak1, line 14: @cameron-eyds just a note here, this is original status for the registration, might not be the best place...

cameron-eyds: I'm not sure the full context here. 

Is a mutable copy of the statusType?  Amend Transport Permits allows the changing of statusType?

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1761-o7ymfxuz.web.app

ppr-ui/src/interfaces/mhr-registration-interfaces/MhrTransportPermitIF.ts
cameron-eyds: Just food for thought, if Amendments doesn't allow the direct manipulation of the `statusType` i'd recommend not tracking the `statusType` and simple display the amendment badge directly via the condition. 

As far as i know, the API will handle this status change modification, so the UI doesn't really need to manage this unless they mutate and submit it directly.

dimak1: Amend TP allows statusType change (if new location is back to BC). We do not pass any `statusType` to API, so it is only for the UI (API will determine the new status). This prop is to track the original `statusType`, which is compared to new status and the badge is shown. If the Amendment flow is cancelled we fetch MHR info from backend, so all changes are discarded. Let me know if it helps.

Screenshot 2024-03-01 at 09 08 39

cameron-eyds: I understand completely, i think perhaps you misunderstand me that the UI doesn't change the status, so why do we track it like it does? 
You said it yourself, the api does all of this so why write more code in the UI when it's not needed?

All the UI needs to do is indicate to the user the status 'Will' change, not track and submit a request to do a change. 
Tracking of the status in UI and comparing changes between two states which the UI never changes or even submits to api is decoy code. 

The requirements for displaying this badge is described in the designs, the UI doesn't change the status in amendments, so a little confused why the extra steps and complexity. 

Screenshot 2024-03-01 at 11 38 40 AM

cameron-eyds: Just a lot of code to update a string and add a badge when its a simple condition. 
Exempt Home moving into BC 🤷

dimak1: UI will not submit the new registration status.
The requirement (how I understand it) is to actually update the status in the UI in the tombstone. Originally it is showing as Exempt, and when we change Civic Address province to BC, then the status would show Active with a badge. That's why we need to track the status. So not sure which code is not needed?

Screenshot 2024-03-01 at 16 01 10

dimak1: This is my understanding of the status update requirement:
Screenshot 2024-03-01 at 16 17 25

cameron-eyds: Exactly, UI doesn't submit status. It doesn't change it.
All it does is update a label and display a badge, when an exempt home is moving into BC. 

I don't doubt the code works, what i'm suggesting is you can do the exact same with much much less code and complexity but whatever, just trying to help.

dimak1: Yea I know, for sure, there are many ways to do it, I didn't want to get into weird label overrides. I don't see it being complex.

dimak1: You suggest to have a computed on the `region === 'BC'` ? Maybe that's another way to do it.

cameron-eyds: No action required, all good!

PR #1766, opened 2024-03-05
Case Sensitive Amend Badges

dimak1:
*Description of changes:*
- Add case sensitive check for Amend Badges for Transport Permit

### DIFF
--- a/ppr-ui/src/utils/utilities.ts
+++ b/ppr-ui/src/utils/utilities.ts
@@ -98,8 +98,6 @@ export const deepChangesComparison = (
   current: BaseDataUnionIF,
   isCaseSensitive: boolean = false
 ): boolean => {
-  console.log(base, current, isCaseSensitive);
-
   if (isCaseSensitive) {
     return !isEqual(base, current)
   }
--- a/ppr-ui/src/utils/utilities.ts
+++ b/ppr-ui/src/utils/utilities.ts
@@ -1,4 +1,5 @@
 import { BaseDataUnionIF } from '@/interfaces'
+import { isEqual } from 'lodash'
 
 export function isSigningIn (): boolean {
   const path = window.location.pathname
@@ -89,11 +90,22 @@ export const filterDuplicates = (list: Array<any>, filterBy: string) => {
  *
  * @param {*} base - The first value to compare.
  * @param {*} current - The second value to compare.
+ * @param {*} isCaseSensitive - Flag for case-sensitive string comparison. Defaults to false.
  * @returns {boolean} - Returns true if the values are different, false if they are equal.
  */
-export const deepChangesComparison = (base: BaseDataUnionIF, current: BaseDataUnionIF): boolean => {
+export const deepChangesComparison = (
+  base: BaseDataUnionIF,
+  current: BaseDataUnionIF,
+  isCaseSensitive: boolean = false
+): boolean => {
+  console.log(base, current, isCaseSensitive);
+
+  if (isCaseSensitive) {
+    return !isEqual(base, current)
+  }
+
   // Object safety-check
-  const isObject = (value) => typeof value === 'object' && value !== null
+  const isObject = value => typeof value === 'object' && value !== null
   // String normalization safety-check: Case Insensitive
   const caseInsensitiveStringCompare = (val1, val2) => {
     if (typeof val1 === 'string' && typeof val2 === 'string') {
@@ -107,7 +119,11 @@ export const deepChangesComparison = (base: BaseDataUnionIF, current: BaseDataUn
     const keys1 = Object.keys(base)
     const keys2 = Object.keys(current)
 
-    return keys1.length !== keys2.length || keys1.some(key => deepChangesComparison(base[key], current[key]))
+    if (keys1.length !== keys2.length) {
+      return true
+    }
+    return keys1.some(key => deepChangesComparison(base[key], current[key], isCaseSensitive))
   }
+
   return caseInsensitiveStringCompare(base, current)
 }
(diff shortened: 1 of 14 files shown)


bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1766-zecotmr4.web.app

ppr-ui/src/utils/utilities.ts
```diff
@@ -107,7 +117,11 @@ export const deepChangesComparison = (base: BaseDataUnionIF, current: BaseDataUn
     const keys1 = Object.keys(base)
     const keys2 = Object.keys(current)
 
-    return keys1.length !== keys2.length || keys1.some(key => deepChangesComparison(base[key], current[key]))
+    if (keys1.length !== keys2.length) {
+      return true
+    }
+    return keys1.some(key => deepChangesComparison(base[key], current[key], isCaseSensitive))
```
cameron-eyds: Why expand the simplified statement into long form with an `if` condition and double returns? 

`return keys1.length !== keys2.length || keys1.some(key => deepChangesComparison(base[key], current[key], isCaseSensitive))` Is the exact same with less code and more concise no?

cameron-eyds (approved):
Looks good to me! 
Just one question

Edit* I guess another question and maybe a business question, why are we case-sensitive on amendments and not corrections, what was the rationale there from the requirements end?

ppr-ui/src/utils/utilities.ts
dimak1: I was getting `...reaching memory limit: JS heap out of memory` error when running the tests. This fixed it, because it does not need to evaluate the second condition (and won't go deeper into recursion).

Screenshot 2024-03-04 at 16 49 42

dimak1:
> Edit* I guess another question and maybe a business question, why are we case-sensitive on amendments and not corrections, what was the rationale there from the requirements end?

@cameron-eyds I asked Liz if we want to make it consistent (twice). Didn't get a clear answer as to why, but she asked to implement the requirement. 🤷🏻

ppr-ui/src/utils/utilities.ts
cameron-eyds: This doesn't seem right then, there was no issue before. 

Please look a bit deeper into this, changing code syntax to pass unit tests is somewhat of a red flag, especially considering the simplicity of the statement.

cameron-eyds (changes requested):
Requesting a deeper look into the condition refactored to pass unit tests.

cameron-eyds:
> > Edit* I guess another question and maybe a business question, why are we case-sensitive on amendments and not corrections, what was the rationale there from the requirements end?
> 
> @cameron-eyds I asked Liz if we want to make it consistent (twice). Didn't get a clear answer as to why, but she asked to implement the requirement. 🤷🏻

Fair enough, thank you for inquiring!

doug-lovett (approved):
Nothing to add to Cameron's review.

ppr-ui/src/utils/utilities.ts
cameron-eyds: Let me know if you want me to run the code and help with the unit test debugging.

dimak1: Yea, thanks, I spent few hours today trying to figure it out. I was surprised as well to get this error. But I have another solution, so will try it. Will let you know.

dimak1: Also, the build tests didn't have the same js heap issue.

cameron-eyds: Ok, fair enough, if you've already spent time looking into it than all good. 
It does seem related to test set up, which is why i lean to leaving the code as it was and refactoring the test setup but i'll leave it to you to decide.

cameron-eyds (approved):

ppr-ui/src/utils/utilities.ts
dimak1: I restarted my machine, cleared cache, reinstalled the node, etc. and the tests passed, once... Now it is failing again. Very strange.

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1766-zecotmr4.web.app

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1766-zecotmr4.web.app

dimak1:
@cameron-eyds yay it's 🟢 please take a look at latest changes, thanks
Edit: tests are passing now, hopefully won't cause issues in future PR.

ppr-ui/src/utils/utilities.ts
dimak1: Regarding this, I tried to revert to original logic, but it failed with memory error.

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1766-zecotmr4.web.app

ppr-ui/src/utils/utilities.ts
cameron-eyds: Hmm yea, as you've noted in your other PR it's not really related to this specific logic, so i think it's safe to revert and continue to debug, or rerun the unit test suite until it passes since it's intermittent. 

We can focus on this as a single issue and not let it block this PR but we can now be pretty certain it's not related to the || statement, as your now seeing this in all the PR's. 
My recommendation is to revert and continue to run the tests until they pass and we can tackle this issue standalone, i don't think it's related to any code in this PR.

dimak1: Yea I agree it's not probably related to this. However, I still think that we could avoid logical OR, just because we don't want to run the second conditional statement (`keys1.some(key => deepChangesComparison...`) if first condition fails. It might be fast compare, but unnecessary?

dimak1: Updated as per your recommendation.

PR #1776, opened 2024-03-10
Fix account search history response inProgress mapping.

doug-lovett:
*Issue #:* /bcgov/entity#20254

*Description of changes:*
- Use search_result.api_result, not search_request.username to set the response inProgress value.
- Return a username if either a user first name or last name exists in the users table for the user matching the search user id. 


dimak1 (approved):

cameron-eyds (approved):
Looks good!

PR #1797, opened 2024-03-22
Prevent Multiple MHR Submissions

dimak1:
*Issue #:* 
- bcgov/entity#20458

*Description of changes:*
- Add `debounce()` of 500ms to create a delay emitting MHR `submit`, to prevent multiple submission calls

### DIFF
--- a/ppr-ui/src/components/common/ButtonsStacked.vue
+++ b/ppr-ui/src/components/common/ButtonsStacked.vue
@@ -106,6 +106,7 @@ import {
   reactive,
   toRefs
 } from 'vue'
+import { debounce } from 'lodash'
 
 export default defineComponent({
   name: 'ButtonsStacked',
@@ -147,9 +148,10 @@ export default defineComponent({
     const cancel = () => {
       emit('cancel', true)
     }
-    const submit = () => {
+    const submit = debounce(() => {
       emit('submit', true)
-    }
+    }, 500) // prevent multiple submissions by adding a small delay
+
     const save = () => {
       emit('save', true)
     }
--- a/ppr-ui/package-lock.json
+++ b/ppr-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ppr-ui",
-  "version": "3.0.69",
+  "version": "3.0.70",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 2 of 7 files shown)


ppr-ui/src/components/common/ButtonsStacked.vue
```diff
@@ -147,9 +148,10 @@ export default defineComponent({
     const cancel = () => {
       emit('cancel', true)
     }
-    const submit = () => {
+    const submit = debounce(() => {
```
doug-lovett, line 151: Will this allow resubmission before the API response? Maybe we should get clarification on requirements?

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1797-qrgum57e.web.app

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1797-qrgum57e.web.app

cameron-eyds (approved):

doug-lovett (approved):
Looks good.

ppr-ui/src/components/common/ButtonsStacked.vue
cameron-eyds: I think it technically could, you bet.
This will prevent multiple multiple function calls within this set timeframe (ie double clicking or whatever) but it's not directly awaiting an api response, so the function might be invokable after that set timeframe. 

I'm not sure if lodash.debounce has some internal await on the original function being returned however, might have to dig in to there documentation to know, if that was the case it may help, but don't know offhand. 

We could do the latter as well or instead of, it that is the requirement for sure. 
I missed most the context behind the issue but we can discuss at stand up!

cameron-eyds: Just to follow up here, Dima addressed the reason for Debounce in the ticket itself. 

The jist: 
There is a loading state in place that will await the network request, the issue is the UI cycle it takes to trigger that state. 
The implementation of Debounce here will cover the gap between the button press and the loading state triggering (which will await the request response) 

So i guess the answer is Debounce wouldn't prevent the resubmission by itself, but it is working in conjunction with that UI loading state. 
I think this should be A-OK!

PR #1799, opened 2024-03-26
Fix debtor search empty string middle name exact match.

doug-lovett:
*Issue #:* /bcgov/entity#2475

*Description of changes:*
- Fix individual debtor search with a middle name excluding exact match when the middle name is an empty string. 


cameron-eyds (approved):
Looks splendid!

PR #1832, opened 2024-04-17
All outputs TC add owner tenancy type.

doug-lovett:
*Issue #:* /bcgov/entity#20879

*Description of changes:*
- In all reports with a tenant in common owner structure, display the owner group tenancy type.
- Example reports attached to ticket.


cameron-eyds (approved):
Looks great!

PR #1835, opened 2024-04-18
Fix regc, puba report generation no location change.

doug-lovett:
*Issue #:* /bcgov/entity#20833

*Description of changes:*
Fix a recent change to report generation location information for REGC, PUBA registrations where there is no location change.


cameron-eyds (approved):
Looks good!

PR #1838, opened 2024-04-22
Fix Tenancy Type for Business and Individual roles

dimak1:
*Issue #:* 
- bcgov/entity#18733

*Description of changes:*
- Fix Home Tenancy Type by treating Business and Individual home owners as the same type


doug-lovett (approved):
Looks good.

PR #1850, opened 2024-04-26
Cancel transport permit updates.

doug-lovett:
*Issue #:* /bcgov/entity#19004

*Description of changes:*
- Implement staff and non-staff cancel transport permit registration.
- Limit non-staff access by account id.
- Prevent transport permit registrations when an active transport permit exists.
- Registrations filter by registration type allow filter criteria to be a document description or a document type.  


dimak1 (approved):
LG!

cameron-eyds (approved):
Looks great!

PR #1862, opened 2024-05-08
Misc Staff Transfers Submission

dimak1:
*Issue #:* 
- bcgov/entity#21076

*Description of changes:*
- Submit all remaining staff transfers
- Filter remaining staff transfers in MHR table


ppr-ui/src/enums/registrationTypes.ts
```diff
@@ -194,17 +194,17 @@ export enum APIMhrDescriptionTypes {
...
-  COU = 'COU',
-  FORE = 'FORE',
-  GENT = 'GENT',
-  TRANS_LAND_TITLE = 'TRANS_LAND_TITLE',
-  REIV = 'REIV',
-  REPV = 'REPV',
-  SZL = 'SZL',
-  TAXS = 'TAXS',
-  VEST = 'VEST'
+  ABAN = 'TRANSFER DUE TO ABANDONMENT AND SALE',
```
dimak1, line 197: @doug-lovett I had to remove the codes and use the full description because of the Registration Type filter in MHR Table, otherwise after selecting the Type the code would be displayed in the Registration Type filter.

doug-lovett: Maybe the code could only be used when submitting the API request? Using the description is more brittle. If the match is not exact it won't work. The API returns the description when filtering by document type code.

doug-lovett (approved):
LGTM.

ppr-ui/src/enums/registrationTypes.ts
dimak1: @doug-lovett I updated to use only the codes when filtering the table, but not all codes work. Maybe need to update API. I sent you a message in Teams.

dimak1: I added the mapping for codes, so the filtering is using the codes now, and it works. Thank you!

PR #1872, opened 2024-05-14
AddEdit Notices - Notice History Panels - Court/Commission Order Placeholders 

cameron-eyds:
*Issue #:* /bcgov/entity#21105

*Description of changes:*
- Build out Add/Edit Notice Forms
- Build out custom Expansion Panels for Notice history and management(actions)
- Court and Commission order placeholders
- Updated SAN Access permissions utilize api-provided user preference property
- Registration Model updates to support the start of SAN related data.
- Unit Testing

Screenshot 2024-05-14 at 9 56 44 AM

Screenshot 2024-05-14 at 9 56 52 AM

Screenshot 2024-05-14 at 9 56 58 AM


bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1872-4bdkxrvl.web.app

ppr-ui/src/interfaces/registration-interfaces/security-act-notice-interfaces.ts
```diff
@@ -0,0 +1,5 @@
+
+export interface AddEditSaNoticeIF {
+  noticeType: boolean
```
doug-lovett: Is noticeType boolean?

doug-lovett (approved):
Looks great.

ppr-ui/src/interfaces/registration-interfaces/security-act-notice-interfaces.ts
cameron-eyds: No sir it is not, good catch, thank you!
Should be `SaNoticeTypes`, will fix

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1872-4bdkxrvl.web.app

ppr-ui/src/components/registration/securities-act-notices/AddEditCommissionOrder.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <FormCard
+    id="add-edit-commission-order"
```
dimak1, line 3: Not a big deal (since we used it across many components already), but I ran into scenario when same component is used multiple times on the page, and they all have same ids. Quick workaround was to overwrite the id (when adding component to template), but maybe we can think of some long term solution.

dimak1 (approved):
Looks great, just one comment.

ppr-ui/src/components/registration/securities-act-notices/AddEditCommissionOrder.vue
cameron-eyds: I see, and what were the implications?
Each component should have an identifier at it's root level, imo. It's a practice we've always followed.

I think what you did is a viable solution for multiple instances of a single component, very much the same in this example. 
FormCard for example is a common component, in this case, we've applied an identifier unique to this implementation. 

I'm not sure there is a catch all solution but definitely open to your thoughts!

PR #1894, opened 2024-05-23
Fix legacy setting of amend permit expiry date.

doug-lovett:
*Issue #:* /bcgov/entity#21002

*Description of changes:*
- Fix to set the expiry date correctly in DB2 for an amend transport permit registration when there are multiple previous permits.


cameron-eyds (approved):
LG!

dimak1 (approved):

PR #1904, opened 2024-05-30
PPR API allow callback URLs when large report data size.

doug-lovett:
*Issue #:* /bcgov/entity#

*Description of changes:*
- PPR API accept optional callback URLs and send notifications to API accounts when search results report data size is large but the number of registrations is less than 75.


cameron-eyds (approved):
Looks good!

dimak1 (approved):

PR #1938, opened 2024-06-14
MHR permit active exists, legacy owner names.

doug-lovett:
*Issue #:* 
- /bcgov/entity#21715
- /bcgov/entity#21765

*Description of changes:*
- 21715 Conditionally allow a transport permit registration when an active permit exists.
- 21765 Individual owner names legacy save update when there are multiple middle names: conditionally save to the DB2 owner suffix. Handle Administrator/Executor/Trustee with multiple middle names.


cameron-eyds (approved):
Looks great!

dimak1 (approved):

PR #1950, opened 2024-06-24
PPR account drafts mark as used.

doug-lovett:
*Issue #:* /bcgov/entity#21975

*Description of changes:*
- Improve accounts drafts query - update draft account id when a draft is used in a registration. 


cameron-eyds (approved):
FanTasTico!

PR #1954, opened 2024-06-26
Update account IDs when registrations become historical.

doug-lovett:
*Issue #:* /bcgov/entity#21973

*Description of changes:*
- Create sql script to update accounts IDs for existing registrations that are historical.
- Create job to update accounts IDs for registrations that became historical within the last day.
- Update account ID's when removing/restoring registrations created by the account.
- SQL script run in DEV. Job deployed, configured, tested in DEV. Scheduled to run once a week in non-PROD: Wednesdays at 10:15 PM.


cameron-eyds (approved):
Looks good!

PR #1959, opened 2024-06-28
Tenants in common order groups by sequence number

doug-lovett:
*Issue #:* /bcgov/entity#22014

*Description of changes:*
- Data migration capture legacy group sequence numbers.
- Patch existing registrations.
- Add a sequence number to represent TC group order
- Update test data and unit tests.
- Update response and report data to order TC groups by the sequence number.

Note: for TC transfers, corrections, amendments, re-registrations, the addOwnerGroups group.groupId is now the sequence number of the new group.
So if only modifying the second group, the addOwnerGroups[0].groupId value is 2. A UI change may be required.

*Issue #:* /bcgov/entity#21829

*Description of changes:*
- Legacy registration if no description.baseInformation.circa value set the default to false.


cameron-eyds (approved):
Looks good!

mhr_api/src/mhr_api/models/db2/descript.py
```diff
@@ -289,6 +289,8 @@ def registration_json(self):
             description['engineerDate'] = model_utils.format_local_date(self.engineer_date)
         if self.circa == '?':
             description['baseInformation']['circa'] = True
+        else:
+            description['baseInformation']['circa'] = False
```
dimak1, line 293: 👍🏻

dimak1 (approved):

PR #1963, opened 2024-07-03
Add Focus State for Tabbable Elements

dimak1:
*Issue #:* 
- bcgov/entity#21781

*Description of changes:*
- Use :focus-visible pseudo class to apply outline styling only on Tab 
- Vue plugin to add event listeners for v-text-fields to fulfill the Tab-only requirement (no focus ring on mouse click)

Screenshot 2024-07-03 at 14 05 19

### DIFF
--- a/ppr-ui/package-lock.json
+++ b/ppr-ui/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ppr-ui",
-  "version": "3.2.31",
+  "version": "3.2.32",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "ppr-ui",
-      "version": "3.2.31",
+      "version": "3.2.32",
       "dependencies": {
         "@bcrs-shared-components/input-field-date-picker": "^1.0.0",
         "@lemoncode/fonk": "^1.5.1",
--- a/ppr-ui/package.json
+++ b/ppr-ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ppr-ui",
-  "version": "3.2.31",
+  "version": "3.2.32",
   "private": true,
   "appName": "Assets UI",
   "sbcName": "SBC Common Components",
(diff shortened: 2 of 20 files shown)


dimak1:
@cameron-eyds just some context here. The requirement was to apply styling only when tabbing into the fields or buttons. The pseudo class :focus-visible is built specifically for it, and I was able to use it for buttons. However, for text input fields, it does not work - the :focus and :focus-visible are applied at the same time, so there is no way to differentiate it. 

To solve this, I created a plugin that would add tab event listeners to all v-text-fields and style the fields on tab only. It could also be achieved via directive, but then we need to add it to all v-text-fields, which is more work.

Let me know if you have questions/comments. Thanks!

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1963-xvx3zta2.web.app

cameron-eyds:
Code looks fine but some styling on the Registration Table Tabs is gone. The inactive tab would shrink slightly lower than the active tab. Maybe it's ok, would confirm with @LizGovier

cameron-eyds (approved):
Code looks good but pending approval from @LizGovier with regards to the removal of the tab styling.

Edit* FWIW it's a very small thing and the tab designs for our MHR history don't include shrink/grow styling, so hopefully all good with design.

dimak1:
@cameron-eyds thanks, tabs were tricky and bunch of styles needed to be updated in order for focus outline to show. I left some comments for Liz in the ticket, so we'll see.

doug-lovett (approved):
Looks great.

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1963-xvx3zta2.web.app

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1963-xvx3zta2.web.app

bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1963-xvx3zta2.web.app

PR #1979, opened 2024-07-15
Fix SE reg type amendment set up.

doug-lovett:
*Issue #:* /bcgov/entity#22195

*Description of changes:*
- SE current view of base registration remove amendNoticeId from notices, amendOrderId from orders.
- Verification report set up handle case where all orders are deleted from an amended notice.


cameron-eyds (approved):
Looks great!

PR #1981, opened 2024-07-16
Search History Table AA - Screen Reading/Testing

cameron-eyds:
*Issue #:* /bcgov/entity#22054

*Description of changes:*
- Minor A11Y definitions to support preferred Screen Reading approach
- Includes AXE testing on Search History Table
- Added Screen reading support for ToolTip - if this is preferred by UXA/QA we can move forward with this approach

**Notes** 
- Lots of good findings from this ticket have been documented in the ticket itself - please read!
- Also we need to align on what screen reading tools we use for which browsers. They are not all equal and do not all work the same. I've documented findings in the ticket but the jist is: 

VoiceOver (macOs): Recommended for Safari
NCDA/ChromeVox: Recommended for Chrome/Firefox

Each of these tools yields different results, so in order for accurate development and testing we need to be aware of this and standardize what we tool we use for the respective browsers. 

### DIFF
--- a/ppr-ui/src/components/tables/SearchHistory.vue
+++ b/ppr-ui/src/components/tables/SearchHistory.vue
@@ -1,5 +1,8 @@
 <template>
-  <v-container class="main-results-div pa-0 bg-white">
+  <v-container
+    class="main-results-div pa-0 bg-white"
+    role="region"
+  >
     <v-row
       noGutters
       class="pt-4"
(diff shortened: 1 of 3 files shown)


bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-1981-0vwva9st.web.app

doug-lovett (approved):
One small step for man....

cameron-eyds:
> One small step for man....

One giant leap for finding that missing aria-label...

flutistar (approved):
LGTM

PR #1990, opened 2024-07-20
Allow QS dealer transfers.

doug-lovett:
*Issue #:* /bcgov/entity#22328

*Description of changes:*
- Allow QS dealers to submit API transfer TRANS registration type requests with data validations.
- Fix existing legacy AFFE incorrectly preventing registrations because the query order is incorrect. 

*Issue #:* /bcgov/entity#22379

*Description of changes:*
- Exemption report add owner suffix/description.


dimak1 (approved):
LG!

cameron-eyds (approved):
Looks good!

PR #1991, opened 2024-07-22
enabled change ownership button when hasLien and isRoleStaffReg are true

flutistar:
*Issue #:* /bcgov/entity#22366

*Description of changes:*
Enabled `Change Ownership` button when there is a lien and logged in as a staff.

### DIFF
--- a/ppr-ui/src/views/mhrInformation/MhrInformation.vue
+++ b/ppr-ui/src/views/mhrInformation/MhrInformation.vue
@@ -892,8 +892,9 @@ export default defineComponent({
         'Ownership Transfer or Change'
       ),
       isChangeOwnershipBtnDisabled: computed((): boolean => {
-        if(isRoleStaffReg.value && hasLien.value)
-          return false  
+        if(isRoleStaffReg.value && hasLien.value && !isFrozenMhrDueToAffidavit.value){
+          return false
+        }  
 
         const isFrozenMhr = isFrozenMhrDueToAffidavit.value || isFrozenMhrDueToUnitNote.value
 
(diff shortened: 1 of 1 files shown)


dimak1 (approved):
LGTM! Would be nice to support with a small unit test, as an option.

flutistar:
> LGTM! Would be nice to support with a small unit test, as an option.

Sure!, I will add a test case for this.

doug-lovett (approved):

ppr-ui/src/views/mhrInformation/MhrInformation.vue
```diff
@@ -892,6 +892,9 @@ export default defineComponent({
         'Ownership Transfer or Change'
       ),
       isChangeOwnershipBtnDisabled: computed((): boolean => {
+        if(isRoleStaffReg.value && hasLien.value)
```
cameron-eyds: This may not work totally as intended, as this will ignore the rest of the conditions within the computed value. 
They may not all be applicable to staff but i think if an MHR is FROZEN and due to AFFE we would want the button to be disabled, if we add this it will be enabled if there is also a lien registered against the MHR.

Can either modify the functions return or include the !isFrozenMhrDueToAffidavit.value in your condition here. 

Also would use probably use { } if putting the return on a new line, generally.

doug-lovett: Nice catch Cameron.

flutistar: @cameron-eyds  addressed your comment.

cameron-eyds (approved):
Looks good, thank you

ppr-ui/src/views/mhrInformation/MhrInformation.vue
cameron-eyds: Thank you!

flutistar:
> Looks good, thank you

Thanks @cameron-eyds 
I am gonna merge this PR.

PR #1994, opened 2024-07-29
Ignore search results callbackURL if callback condition does not exist.

doug-lovett:
*Issue #:* /bcgov/entity#22539

*Description of changes:*
- Ignore a search result request with a callback URL and request for a PDF is a callback condition does not exist.


cameron-eyds (approved):
Looking good!

PR #2004, opened 2024-08-20
fixed error on transfer pop up modal

flutistar:
*Issue #:* /bcgov/entity#16198

*Description of changes:*
I changed the method for generating `transferRequiredDialogOptions`. Initially, it works correctly with `text.replace()`, but subsequent changes to the value aren't reflected, even though the `computed` function is called.


cameron-eyds (approved):
LG!

doug-lovett (approved):

PR #2033, opened 2024-10-08
Include REG_103 when setting current reg changePermit

doug-lovett:
*Issue #:* /bcgov/entity#23541

*Description of changes:*
- Include REG_103E doc type when setting current registration changePermit property.
- Add new property permitExtended and set to true when a permit is extended with a REG_103E doc type.


cameron-eyds (approved):
Looks good!

PR #2045, opened 2024-10-23
MHR API fix reg summary frozen status fix.

doug-lovett:
*Issue #:* /bcgov/entity#23931

*Description of changes:*
- Legacy data migration document records do not always match the registration sequence. Use the registration timestamp, not the document record ID, to determine the last registration document type for setting the locked/frozen status. Fix for summary incorrectly setting the status as FROZEN.


cameron-eyds (approved):
Looking good, thanks sir!

PR #2047, opened 2024-10-24
Preserve date filter on registration table

flutistar:
*Issue #:* /bcgov/entity#23445
*Description of changes:*

- [ ] Passed defaultStartDate and defaultEndDate to DatePicker
- [ ] Updated placeholder text accordingly
- [ ] A function that returns PST date object.


bcregistry-sre:
Temporary Url for review: https://bcregistry-assets-dev--pr-2047-vo6ktze4.web.app

ppr-ui/src/components/common/RangeDatePicker.vue
```diff
@@ -30,6 +31,7 @@
         </b>
         <BaseDatePicker
           class="date-selection__picker mt-2"
+          :default-selected-date="endDate"
```
cameron-eyds, line 34: Small thing but in this project we've been going with camelCase for the prop names

cameron-eyds (approved):
Looks good!
So the issue was just the display of the date in the picker? 
The date was still filtering by the date under the hood?

doug-lovett (approved):
LGTM

flutistar:
> Looks good! So the issue was just the display of the date in the picker? The date was still filtering by the date under the hood?

Displaying the date and place holder text.
When the page is loaded, useRegistration sets placeholder text as empty.
And on the component, there is a `watch` function that initializing `startDate` and `endDate` if `dateTxt` is empty.
So we have both dates in store but not in local state.

I think that's why.

flutistar:
@cameron-eyds 
FYI, I addressed your comment :)

PR #2060, opened 2024-11-08
API: Unrestricted rule for home dealers

flutistar:
*Issue #:* /bcgov/entity#24056
*Description of changes:*

- Removed location restrictions on transport permits
- sole owner restrictions on transfer due to sale or gift

NOTE: Since I am not comfortable to update the DB yet, I left the other 2 tasks.


mhr-api/src/mhr_api/utils/registration_validator.py
```diff
@@ -276,16 +276,17 @@ def validate_permit(
             error_msg += validator_utils.validate_registration_state(
                 registration, staff, MhrRegistrationTypes.PERMIT, MhrDocumentTypes.REG_103, json_data
             )
-            if registration and group_name and group_name == MANUFACTURER_GROUP:
```
doug-lovett, line 279: From the ticket description the removal of the restriction applies only to dealers. MANUFACTURER_GROUP applies to QS manufactures, which I do not think has changed.

mhr-api/tests/unit/utils/test_permit_validator.py
```diff
@@ -376,18 +376,18 @@
     ('Missing land status confirm MH_PARK', False, '000900', LOCATION_PARK_2,
      validator_utils.STATUS_CONFIRMATION_REQUIRED,
      REQUEST_TRANSPORT_PERMIT),
-    ('MANUFACTURER no existing lot', False, '000919', None, validator.MANUFACTURER_DEALER_INVALID,
```
doug-lovett: If the validation change only applies to QS dealers then the manufacturer tests can remain as is.

doug-lovett (commented):
Looks good, just a comment on scope - if QS manufacturer validation is changing? From the ticket description it looks like only QS dealer validation is changing.

mhr-api/src/mhr_api/utils/registration_validator.py
cameron-eyds: Definitely correct, only Dealers are changing.

flutistar: Yep! I will update it.

mhr-api/tests/unit/utils/test_permit_validator.py
flutistar: I will revert this change!

PR #2068, opened 2024-11-19
Fix report set up to work with TC group structure.

doug-lovett:
*Issue #:* /bcgov/entity#24303

*Description of changes:*
- Fix report set up for changed transfer owners to work with the TC group structure.


cameron-eyds (approved):
Looks great!

PR #2119, opened 2025-03-14
RL to CL transition discharge

doug-lovett:
*Issue #:* /bcgov/entity#26486

*Description of changes:*
- RL discharges convert RL registration type to CL after CLA is enacted.
- Example reports before and after attached to the ticket.


cameron-eyds (approved):
Looking great!!

PR #2134, opened 2025-04-05
PPR API omit debtor email addresses.

doug-lovett:
*Issue #:* /bcgov/entity#26960

*Description of changes:*
- Do not save debtor email with new registrations.
- Do not return existing debtor email addresses in API registration and search responses.


cameron-eyds (approved):
Looking great!

PR #2135, opened 2025-04-08
Corrections Status Bug: Duplicate Identifier

cameron-eyds:
*Issue #:* /bcgov/entity#26998

*Description of changes:*
- Fixed duplicate radio btn identifier which would cause functional issues when vite build the application and the duplicate identifiers were rendering at the same time. 


flutistar (approved):
Super!

PR #2145, opened 2025-05-01
MHR API process credit card payments

doug-lovett:
*Issue #:* /bcgov/entity#27259

*Description of changes:*
See ticket for more details
- Add optional new request parameter (ccPayment=true) for all public registration endpoints to indicate CC payment (if preferred account payment is not CC). UI should know the payment method in advance to handle the mhr api response and redirect user to the CC payment url. 
- Save draft record without marking it as consumed if pay api call is successful.
- Associate the pay api response invoice id with the draft.
- For change registrations lock the manufactured home to prevent further changes until payment completes.
- Return to the UI the pay api response invoice ID. Response HTTP status is 202 accepted; new payload paymentPending property is set to true.
- Create new endpoint to receive payment complete requests and create the registration.
- Add/update unit tests.


cameron-eyds (approved):
Outstanding!

flutistar (approved):
Looks good

PR #2147, opened 2025-05-07
MHR CC Handling

cameron-eyds:
*Issue #:* /bcgov/entity#27255

*Description of changes:*
- Include CC required handling for client Manufactured Home: Registrations, Transfers, Transport Permits.
- Include placeholder CC messaging (to be expanded)


flutistar (approved):
Looks good

ppr-ui/src/utils/mhr-api-helper.ts
```diff
@@ -109,6 +111,9 @@ function mhrStaffPaymentParameters (staffPayment: StaffPaymentIF) {
     if (staffPayment.isPriority) {
       paymentParams = paymentParams + '&priority=true'
     }
+    if (staffPayment.isCreditCard) {
```
doug-lovett: Should this be ccPayment?

doug-lovett (approved):
Looks great, just the one comment.

ppr-ui/src/utils/mhr-api-helper.ts
cameron-eyds: You are correct! copypasta error in the param itself. 

On having another look, i need to refactor into the switch itself, as this will be a explicit payment option, not a checkbox(ie in addition to payment)

PR #2158, opened 2025-05-26
27432 extend sbc connect layer

flutistar:
*Issue #:* /bcgov/entity###

*Description of changes:*
- upgraded Nuxt version from 3.14 to 3.16
- Extended SBC connect layer


ppr-ui/src/assets/css/tw.css
```diff
@@ -0,0 +1,247 @@
+/* Lines below re-import Tailwind base styles to fix issues where utilities didn't work due to Vuetify's global reset */
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
+@import "tailwindcss" theme(static);
+@import "@nuxt/ui";
+@source "../../../node_modules/@sbc-connect/nuxt-core-layer-beta";
```
cameron-eyds, line 7: We may be able to set up an alias for @sbc-connect, in the same way we did for sbc common components. 
Especially if we are going to be calling this in more than one place.

cameron-eyds (approved):
Looks great to me!

ppr-ui/src/assets/css/tw.css
deetz99: > We may be able to set up an alias for @sbc-connect, in the same way we did for sbc common components. Especially if we are going to be calling this in more than one place.

There is an alias, you can use `import x from '#core'`.

I haven't tried this in css files though.

cameron-eyds: I think Brandon mentioned there was some issues in the css file, definitely worth a try though, thanks! 
@flutistar Your call sir, we can also resolve later too, not a blocker :)

flutistar: I’m sorry, but I added an alias in nuxt.config.ts and tried to use it in a CSS file, but it doesn’t work. I’ll take another look

doug-lovett (approved):
Looks good, no additional comments.

PR #2159, opened 2025-05-26
CC Redirect url

cameron-eyds:
*Issue #:* /bcgov/entity#27516

*Description of changes:*
- Pull redirect url from session as base url is set in env's


flutistar (approved):
Looks good!

PR #2176, opened 2025-06-24
MHR, PPR API add draftNumber to account registrations.

doug-lovett:
*Issue #:* /bcgov/entity#29100

*Description of changes:*
Account registrations add identifier to link registration with a payment pending draft by draft number. Pending registrations have a draft number with a prefix of 'P'. The prefix is removed when the payment completes and the registration is created.
- For MHR add draftNumber to the GET /mhr/api/v1/registrations endpoint.
- For PPR add draftNumber (draft documentId) to the GET /mhr/api/v1/registrations endpoint.

### DIFF
--- a/mhr-api/pyproject.toml
+++ b/mhr-api/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mhr-api"
-version = "2.0.14"
+version = "2.0.15"
 description = ""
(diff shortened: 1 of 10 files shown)


cameron-eyds (approved):
Looks great, thank you!

flutistar (approved):
Looks good!

PR #2195, opened 2025-07-17
MHR API new reg pay api add reg number.

doug-lovett:
*Issue #:* /bcgov/entity#29054

*Description of changes:*
- New MH registrations generate MHR numbers before the pay api call. Include the MHR number in the pay api payload.
- Update the pay api details info that shows up in the payment receipt and the account transactions to include the MHR number.
- Add/update unit tests.
- Add logic to reuse a MHR number when the registration fails.

### DIFF
--- a/mhr-api/pyproject.toml
+++ b/mhr-api/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mhr-api"
-version = "2.0.15"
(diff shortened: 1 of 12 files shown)


cameron-eyds (approved):
Fantastic!

flutistar (approved):
Looks great!

PR #2201, opened 2025-07-29
PPR API add paging of account search history.

doug-lovett:
*Issue #:* /bcgov/entity#29739

*Description of changes:*
- Start page number is 1. Default if no pageNumber request parameter is 1.
- The page size is 1000. Specifying the next page when there are no more results will return an empty list.
- The first item in the list if results exist contains the total history count for the account as property searchHistoryTotal.


flutistar (approved):
Looks good!

cameron-eyds (approved):
Muy Bueno!

PR #2233, opened 2025-09-04
PPR API allow staff historical registrations access.

doug-lovett:
*Issue #:* /bcgov/entity#27065

*Description of changes:*
- PPR expired/discharged historical registrations: allow staff to add, remove, see in the account registrations, download reports.


flutistar (approved):
Legendary PRs, as always!

PR #2236, opened 2025-09-09
Refactor selection function in Search Results table

flutistar:
*Issue #:* /bcgov/entity#30001

*Description of changes:*
- Updated selection function to work with an array object instead of individual items
- Adjusted logic since item is now passed from the child component


doug-lovett (approved):
Looks great.

PR #2238, opened 2025-09-09
MHR API QS lawyer/notary submit Transfer WILL registrations.

doug-lovett:
*Issue #:* /bcgov/entity#30487

*Description of changes:*
- Model changes to support new staff approval registrations. Initially only the WILL document type.
- Create a copy of the registration data as the staff approval registration. 
- Lock the home.
- Allow QS priority fee payment
- Use the same QS access rules for the WILL document type as those for the DEAT type.


flutistar (approved):
Wow, amazing!

PR #2244, opened 2025-09-19
Search History Table

flutistar:
*Issue #:* /bcgov/entity#30005

*Description of changes:*
- Sort by any columns
- Removed 3 letters delay
- Fixed issue on cleaning date filter

image

### DIFF
--- a/ppr-ui/src/components/search/SearchBarList.vue
+++ b/ppr-ui/src/components/search/SearchBarList.vue
@@ -77,7 +77,7 @@
 import { computed, defineComponent, reactive, ref, toRefs, nextTick, watch } from 'vue'
 import { useStore } from '@/store/store'
 import { MHRSearchTypes, SearchTypes } from '@/resources'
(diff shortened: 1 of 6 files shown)


cameron-eyds (approved):
Looks great sir! 

Quick question, maybe for design, but perhaps we should increase the view height of the search history table?

With the addition of the filter rows, we can't see more than 3 entries in the table, previously it was around 5.

What are your thoughts, should we increase the height a bit to include a few additional entries in the view?

flutistar:
> Looks great sir!
> 
> Quick question, maybe for design, but perhaps we should increase the view height of the search history table?
> 
> With the addition of the filter rows, we can't see more than 3 entries in the table, previously it was around 5.
> 
> What are your thoughts, should we increase the height a bit to include a few additional entries in the view?

You are absolutely right!

flutistar:
image

doug-lovett (approved):
Looks great.

PR #2269, opened 2025-12-02
31394 - Fix amendment response for trust indenture

chenhongjing:
*Issue #:* /bcgov/entity#31394

*Description of changes:*

- Fix bug when a PPR amendment is submitted to update trust indenture, `addTrustIndenture` and `removeTrustIndenture` will incorrectly return in response. The ticket only describes one scenario, but this fix is to resolve all the scenarios.
- Scenarios for trust indenture changes

| Before Amendment | After Amendment | Current Response | Expected Response |
| -------------------------------- | ------------------------------- | ---------------- | ----------------- |
| N                                | N                               | add + remove     | remove            |
| N                                | Y                               | add + remove     | add               |
| Y                                | Y                               | add + remove     | add               |
| Y                                | N                               | add + remove     | remove            |

- Update unit tests

### DIFF
--- a/ppr-api/tests/conftest.py
+++ b/ppr-api/tests/conftest.py
@@ -94,6 +94,7 @@ def session(db, request):  # pylint: disable=redefined-outer-name
 
     def commit():
         db.session.flush()
+        db.session.expire_all()
 
     # patch commit method
     old_commit = db.session.commit
--- a/ppr-api/pyproject.toml
+++ b/ppr-api/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "ppr-api"
-version = "1.3.3"
+version = "1.3.4"
 description = ""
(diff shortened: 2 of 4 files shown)


ppr-api/tests/conftest.py
```diff
@@ -94,6 +94,7 @@ def session(db, request):  # pylint: disable=redefined-outer-name
 
     def commit():
         db.session.flush()
+        db.session.expire_all()
```
chenhongjing, line 97: I added this to mimic the behaviour of real `commit()` so that all cached objects and relationships will expire and reload.

For the unit tests I updated, the code can correctly see the relationship updates around trust indenture and go into the correct path.

doug-lovett (approved):
Looks good, could you also increase the version in pyproject.toml.

cameron-eyds (approved):
Looks great!

PR #2274, opened 2025-12-17
MHR API account registrations fix missing child registrations.

doug-lovett:
*Issue #:* /bcgov/entity#31696

*Description of changes:*
- Fix account registrations list missing registrations when the account registers a new home and another account creates a change registration on the home.
- Interim fix to improve PROD staff account registrations performance when no filters - only include staff registrations created in the last 14 days in the query. No date restriction on registrations added. Note: a background job that runs once a week in PROD removes all staff registrations more than 14 days old from the staff list.


cameron-eyds (approved):
Looking good!

chenhongjing (approved):
LGTM! 😃

PR #2275, opened 2025-12-19
10653 Show account name on account registrations for API User registrations

chenhongjing:
*Issue #:* /bcgov/entity#10653

*Description of changes:*
- Create API user if the record doesn't exist during `post_financing_statements()`
- Update unit tests
- Update query & alembic script to make sure no leading/trailing whitespace for `registering_name`
- ppr-api version = 1.3.5


chenhongjing:
`registering_name` is updated

from 

```sql
(SELECT CASE WHEN r.user_id IS NULL THEN ''
                    ELSE (SELECT CASE WHEN u.lastname = '' or u.lastname IS NULL THEN u.firstname
                                 ELSE u.firstname || ' ' || u.lastname END
                            FROM users u
                           WHERE u.username = r.user_id FETCH FIRST 1 ROWS ONLY) END) AS registering_name
```

to

```sql
(SELECT CASE
            WHEN r.user_id IS NULL OR r.user_id = '' THEN ''
            ELSE COALESCE((
                SELECT CONCAT_WS(' ', NULLIF(TRIM(u.firstname), ''), NULLIF(TRIM(u.lastname), ''))
                FROM users u
                WHERE u.username = r.user_id FETCH FIRST 1 ROWS ONLY
            ), '') END) AS registering_name
```

doug-lovett (approved):
Looks great, thanks for updating the db views. I reviewed the db migration script and the upgrade views look correct.

cameron-eyds (approved):
Well done!

PR #2278, opened 2026-01-16
30747 MHR transfer queue staff review: Ensure status in header is reactive to the assign/un-assign

eve-git:
*Issue #:* /bcgov/entity###
https://github.com/bcgov/entity/issues/30747

*Description of changes:*
1. Display readable transfer-queue status 
2. Keep the tombstone transfer status in sync on the Transfer Details page when it changes.

### DIFF
--- a/ppr-ui/src/composables/analystQueue/resources.ts
+++ b/ppr-ui/src/composables/analystQueue/resources.ts
@@ -1,11 +1,9 @@
 import { FilterTypes } from '@/enums'
 import { ReviewStatusTypes, ReviewRegTypes } from '@/composables/analystQueue/enums'
+import { enumToLabel } from '@/utils'
 
-const transformEnumToLabel = (enumValue: string): string => {
-  return enumValue
-    .split('_')
-    .map(word => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase())
-    .join(' ')
+export const transformEnumToLabel = (enumValue?: string | null): string => {
+  return enumToLabel(enumValue)
 }
 
 export const queueTableColumns = [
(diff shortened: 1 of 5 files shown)


ppr-ui/src/composables/analystQueue/resources.ts
```diff
@@ -1,11 +1,9 @@
...
+import { enumToLabel } from '@/utils'
 
-const transformEnumToLabel = (enumValue: string): string => {
-  return enumValue
-    .split('_')
-    .map(word => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase())
-    .join(' ')
+export const transformEnumToLabel = (enumValue?: string | null): string => {
+  return enumToLabel(enumValue)
 }
```
cameron-eyds: I'm not sure you need this, if your remove/moving the logic? 
Could just use `enumToLabel()` directly no?

cameron-eyds (commented):
Code looks fine to me!

The code/component refactored does have implications in many places, as this is a common header component. I would take some to ensure the removal and refactoring of that status logic doesn't have any implications in other places.

ppr-ui/src/composables/analystQueue/resources.ts
eve-git: you are right, it is better to replace transformEnumToLabel with enumToLabel

eve-git: Thanks for the review! I agree that changes to a common header component need to be handled carefully. I traced the code and tested as thoroughly as possible, and everything looks good. I’ll add your comment to the ticket so QA is aware.

cameron-eyds (approved):
Thanks Eve, looks good!

One thing you may of noticed, in the Unit Tests, some of them are failing now because of auto import errors in the unit testing env.

This is no fault of your own just something to be aware of. 
When we implemented Nuxt, there was a lot of finessing to correct the unit testing environment. 

This is something we could create a tech debt ticket for, as well as fixing the environment errors.

Screenshot 2026-01-16 at 1 38 45 PM

doug-lovett (approved):
Looks great.

PR #2280, opened 2026-01-19
31866-2 MHR-API: Update serial number search algorithm to prioritize exact match (QS)

chenhongjing:
*Issue #:* /bcgov/entity#31866

*Description of changes:*
- Update serial number search query for QS
- Update unit tests
- mhr-api version = 2.1.5 (no change)

### DIFF
--- a/mhr-api/poetry.lock
+++ b/mhr-api/poetry.lock
@@ -835,18 +835,9 @@ files = [
 [package.dependencies]
 google-auth = ">=2.14.1,<3.0.0"
 googleapis-common-protos = ">=1.56.2,<2.0.0"
-grpcio = [
-    {version = ">=1.49.1,<2.0.0", optional = true, markers = "python_version >= \"3.11\" and extra == \"grpc\""},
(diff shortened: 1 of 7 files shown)


doug-lovett (commented):
I should have been more clear, currently serial number wildcard search is registries staff only. For all other users serial number searches are by the original search key matching. In general, the term "qualified suppliers" includes dealers, manufacturers, and lawyers/notaries.

chenhongjing:
Looks like there're some changes for CI script on SRE side. I try to lock python version to 3.12 and solves dependency installation error, otherwise it will install 3.14 automatically. But there's new error after it. 😅

chenhongjing:
> Looks like there're some changes for CI script on SRE side. I try to lock python version to 3.12 and solves dependency installation error, otherwise it will install 3.14 automatically. But there's new error after it. 😅

_Update: Run on previous CI flow (all passed), and Patrick will investigate the issue of the new flow._

doug-lovett (approved):
Looks great.

cameron-eyds (approved):
Looks good!

PR #2328, opened 2026-05-21
PPR API boost CD container timeout for large search reports.

doug-lovett:
*Issue #:* /bcgov/entity###

*Description of changes:*
PPR API large search result reports 2500+ pages, 70+ MB with a table of contents containing page numbers and registration links can take over 15 minutes to generate. Increase the container timeout - impacts callback async requests only.


bolyachevets (approved):
small chance those values may need to be in quotes (but let's try as is)

PR #2329, opened 2026-05-29
PPR replace GCP SA keys with container ADC for GCP API calls

doug-lovett:
*Issue #:* /bcgov/entity#33582

*Description of changes:*
SRE has a policy of rotating service account keys every 3 months. Removing the keys from the container configuration removes the need to update the containers and redeploy when the keys change.
- Verify or enable IAM Service Account Credentials API on the project.
- Assign the container service accounts the correct roles: storage admin, pub/sub publisher, and service account token creator.
- Update all PPR jobs that use a GCP SA key: securd party notification.
- Update the PPR API service
- Remove environment variable from container configuration. 


eve-git (approved):

eve-git:
@doug-lovett I approved the code changes, but I noticed the code-coverage check is failing. Does this need to be addressed in this PR?

doug-lovett:
@eve-git thanks, the CI script code coverage issue has been around for a while, I think it can be left as a separate cleanup task in a different PR.

